Issue #1, opened 2017-01-29
incrementVersion API does not bump version correctly when blank line is missing


szczepiq:
When version declared in the version file does not end with blank line the version bumping fails.

Example usage of "incrementVersion" API in *.gradle file: 
```
project.versionFile.incrementVersion()
```

Example version file:
```
version=1.5.0
```

See the [original ticket](https://github.com/mockito/mockito/issues/915) in Mockito project.

PR #2, opened 2017-01-29
Fixed version bumping when no line break after version was present.

szczepiq:
Fixes issue #1

### DIFF
--- a/src/main/groovy/org/mockito/release/version/DefaultVersionFile.java
+++ b/src/main/groovy/org/mockito/release/version/DefaultVersionFile.java
@@ -41,6 +41,11 @@ class DefaultVersionFile implements VersionFile {
         VersionBumper bumper = new VersionBumper();
         version = bumper.incrementVersion(this.version);
         String content = IOUtil.readFully(versionFile);
+        if (!content.endsWith("\n")) {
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/mockito/release/version/DefaultVersionFile.java
```diff
@@ -41,6 +41,11 @@ public String incrementVersion() {
         VersionBumper bumper = new VersionBumper();
         version = bumper.incrementVersion(this.version);
         String content = IOUtil.readFully(versionFile);
+        if (!content.endsWith("\n")) {
+            //This makes the regex simpler. Add arbitrary end of line at the end of file should not bother anyone.
+            //See also unit tests for this class
+            content += "\n";
+        }
         String updated = content.replaceAll("(?m)^version=(.*?)\n", "version=" + version + "\n");
```
TimvdLippe: Why not remove the two \n on this line?

szczepiq: It wouldn't have the desired outcome. Try it and run tests :)

szczepiq:
I will merge this change. Happy to improve the code if there is more feedback.

PR #30, opened 2017-03-10
Added versioning plugin

szczepiq:
- Added and documented versioning plugin (for now javadoc)
- Also fixed minor problem with using the version number lazily by release notes generator


szczepiq:
Pushed to master directly

Issue #16, opened 2017-02-26
Undesired transitive dependency via json simple


szczepiq:
### Problem

Artifact created by mockito-release-tools pulls in "junit" transitive dependency. When release plugin is used here is the build environment (Gradle buildscript classpath):

```
...
+--- gradle.plugin.org.mockito:mockito-release-tools:0.0.3
|    \--- com.googlecode.json-simple:json-simple:1.1.1
|         \--- junit:junit:4.10 (*)
...
```

We don't want junit to be on client classpath when release plugin is used because it is a test dependency and should not be littering the build classpath.

### Root cause

json-simple library pulls is junit as compile dependency, see issue https://github.com/fangyidong/json-simple/issues/91. To solve the problem we should use a different json library. Let's choose a smallest, simplest json library that has no other dependencies. I suggest we use to fork for json-simple that seems to be well maintained: https://cliftonlabs.github.io/json-simple. We can use 1.x version or 2.x (has some deprecations, some updates in our code would be needed).

szczepiq:
Thank you for the fix!

PR #42, opened 2017-03-22
Undesired transitive dependency via json simple #16

mstachniuk:
Migrate from com.googlecode.json-simple:json-simple:1.1.1 to com.github.cliftonlabs:json-simple:2.1.2 to avoid JUnit transitive dependency

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -27,8 +27,7 @@ dependencies {
     //TODO SF use jcabi to edit issues after the release so that they have the milestone attached
     //compile "com.jcabi:jcabi-github:0.17"
 
-    //Using @jar to avoid transitive dependencies (json-simple declares compile dependency on junit!)
-    compile "com.googlecode.json-simple:json-simple:1.1.1@jar"
+    compile "com.github.cliftonlabs:json-simple:2.1.2"
 
     testCompile("org.spockframework:spock-core:0.7-groovy-2.0") {
         exclude module: "groovy-all"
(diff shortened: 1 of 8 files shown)


src/main/groovy/org/mockito/release/notes/improvements/GitHubTicketFetcher.java
```diff
@@ -32,24 +34,26 @@
...
+                List<JsonObject> page = issues.nextPage();
 
                 out.addAll(extractImprovements(
                         dropTicketsAboveMaxInPage(tickets, page),
                         page, onlyPullRequests));
             }
         } catch (Exception e) {
-            throw new RuntimeException("Problems fetching " + ticketIds.size() + " from GitHub", e);
+            String message = "Problems fetching " + ticketIds.size() + " tickets from GitHub";
+            LOG.info(message, e);
```
szczepiq: Given that we throw an exception, let's avoid also logging the message. Otherwise we might have confusing and untidy dual stack trace when it fails. "Log XOR throw" :)

build.gradle
```diff
@@ -27,8 +27,7 @@ dependencies {
     //TODO SF use jcabi to edit issues after the release so that they have the milestone attached
     //compile "com.jcabi:jcabi-github:0.17"
 
-    //Using @jar to avoid transitive dependencies (json-simple declares compile dependency on junit!)
-    compile "com.googlecode.json-simple:json-simple:1.1.1@jar"
+    compile "com.github.cliftonlabs:json-simple:2.1.2"
```
szczepiq: Great! I checked and this new library has the junit dependency declared at correct scope.

Thanks for updating all the codes!!!

src/main/groovy/org/mockito/release/notes/improvements/GitHubTicketFetcher.java
mstachniuk: It's odd. Throwing only `RuntimeException(message, e)` doesn't print a stacktrace :-( But as you wish I will update this code.

szczepiq:
Thanks a lot!!!

PR #56, opened 2017-04-02
Fixed generation of notable release notes

szczepiq:
- Notable release notes generation now correctly generates notes from head when notable version is released
- Tested with Mockito project
- Bumped version


szczepiq:
Given that the other contributors may not be up to speed to review my changes, and I need to make progress, I'll merge this PR.

Issue #54, opened 2017-03-31
Make updateNotableReleaseNotes task incremental


szczepiq:
In order to make the local builds faster (and testing faster!) let's make updateNotableReleaseNotes task incremental (NotableReleaseNotesGeneratorTask).

### Suggested design

NotableReleaseNotesGeneratorTask class currently fetches release notes data from git / GitHub every time it runs. Instead, let's split the logic into discrete steps to and leverage incremental build. The first step of NotableReleaseNotesGeneratorTask currently is to fetch Collection<ReleaseNotesData> from generator object. Let's push that logic to a separate Gradle task, "fetchNotableNotes" (type: NotableNotesFetcherTask). The new new task will:

- fetch the data and construct Collection<ReleaseNotesData>
- save it to the file (we can use Java serialization for now but we really want something nicer, like JSON)
- declare correct inputs and outputs so that it is incremental, inputs are various properties currently available on NotesGeneration type. Output is a file.

The existing NotableReleaseNotesGeneratorTask needs to:
 - load release notes data (Collection<ReleaseNotesData>) from the file
 - depend on "fetchNotableNotes" task

### Testing and prototyping

- see [Mockito Release Tools readme](https://github.com/mockito/mockito-release-tools#testing) for info on testing
- run "./gradlew updateNotableReleaseNotes" in example project

wwilk:
The main part of the task (making updateNotableReleaseNotes incremental) is done, so I'm closing the task and moving improvements in serialization (JSON instead of Java serialization) to a new issue.

PR #63, opened 2017-04-03
new incremental task fetchNotableNotes #54

wwilk:
As specified in #54 NotableReleaseNotesGeneratorTask was divided and new incremental task NotableNotesFetcherTask was created.

I wasn't sure if I should divide NotesGeneration into two separate configuration classes, so for now I moved it outside of NotableReleaseNotesGeneratorTask to avoid code duplication. 

As we discussed with @szczepiq notable release notes are currently serialized with java serialization. I had to make some classes Serializable because of that. I chose the projectDir for location of temporary serialization file, to keep it simple at first, but probably it's not the best smart default.

Unit test for serializer only tests serialization/deserialization of basic ReleaseNotesData. To test a realistic structure we would need implementation of equals method in most of the POJOs or to use some kind of recursive reflection equality check.


szczepiq:
Woa! First pull request! I will review as soon as I can :)

src/main/groovy/org/mockito/release/internal/gradle/NotesGeneration.java
```diff
@@ -0,0 +1,127 @@
+package org.mockito.release.internal.gradle;
+
+import java.io.File;
+import java.util.Collection;
+
+/**
+ * Created by wilk.wojtek@gmail.com.
+ */ //TODO expose as public API
```
mstachniuk: Please remove TODO and auto generated javadoc

src/main/groovy/org/mockito/release/internal/gradle/DefaultContinuousDeliveryPlugin.java
```diff
@@ -315,11 +315,17 @@ public void execute(Task task) {
     private static void configureNotableReleaseNotes(Project project) {
         VersionFile versionFile = project.getExtensions().getByType(VersionFile.class);
         NotableReleaseNotesGeneratorTask task = (NotableReleaseNotesGeneratorTask) project.getTasks().getByName("updateNotableReleaseNotes");
+        NotableNotesFetcherTask fetcherTask = (NotableNotesFetcherTask) project.getTasks().getByName("fetchNotableNotes");
+
```
szczepiq: Let's keep naming consistency, for clarity. For example:
"updateNotableReleaseNotes + fetchNotableReleaseNotes", "updateNotableNotes + fetchNotableNotes".

src/test/groovy/org/mockito/release/internal/gradle/util/DefaultReleaseNotesSerializerTest.groovy
```diff
@@ -0,0 +1,39 @@
+package org.mockito.release.internal.gradle.util
+
+import org.apache.commons.lang.builder.EqualsBuilder
```
szczepiq: If you want to use commons lang, please add explicit testCompile dependency. You're getting commons lang transitively via Gradle API :) Explicit dependencies improve transparency and are safer in case Gradle API stops shipping / isolates its dependencies.

I'm ok to add more dependencies for testing. However, we should avoid adding compile dependencies because they make the library bigger and increase liability.

src/main/groovy/org/mockito/release/internal/gradle/util/ReleaseNotesSerializer.java
```diff
@@ -0,0 +1,10 @@
+package org.mockito.release.internal.gradle.util;
+
+import org.mockito.release.notes.model.ReleaseNotesData;
+
+import java.util.Collection;
+
+public interface ReleaseNotesSerializer {
```
szczepiq: I suspect that we don't need the interface. I use interfaces for public APIs of packages, while keeping the implementation class package-protected.

In your case DefaultReleaseNotesSerializer is public internal class used directly by the clients. The interface is not really needed and removing it will make the code slightly cleaner.

src/main/groovy/org/mockito/release/internal/gradle/NotesGeneration.java
```diff
@@ -0,0 +1,127 @@
...
+public class NotesGeneration {
```
szczepiq: Good idea to extract this to a separate class and reuse / share in tasks!

src/main/groovy/org/mockito/release/internal/gradle/NotesGeneration.java
szczepiq: Nice catch. The TODO is actually mine and we can leave it for now :)

src/main/groovy/org/mockito/release/internal/gradle/DefaultReleaseNotesPlugin.java
```diff
@@ -95,4 +119,9 @@ private static void configureNotes(DefaultReleaseNotesExtension notes, Project p
 
         //TODO make use of: ext.gh_writeAuthTokenEnvName = "GH_WRITE_TOKEN"
     }
+
+    private static File getTemporaryReleaseNotesFile(Project project){
```
szczepiq: Why not use the "build" dir (e.g. project.getBuildDir())? If you don't want to use the build dir so that the data survives "clean", use ".gradle/mockito-release-tools/". 

".gradle" folder is guaranteed to exist and it's a nice place, already included in .gitignore for all Gradle-like projects.

I vote for using "build" dir for now (because it guarantees garbage collection) and perhaps use something else later on.

src/main/groovy/org/mockito/release/internal/gradle/NotableNotesFetcherTask.java
```diff
@@ -0,0 +1,34 @@
...
+import org.mockito.release.notes.model.ReleaseNotesData;
+
+import java.util.Collection;
+
+public class NotableNotesFetcherTask extends DefaultTask {
+
+    //TODO documentation
+    private final NotesGeneration notesGeneration = new NotesGeneration();
+
+    public NotesGeneration getNotesGeneration() {
```
szczepiq: You probably want to use @Nested property here (check out Gradle docs on this :).

src/main/groovy/org/mockito/release/internal/gradle/DefaultReleaseNotesPlugin.java
```diff
@@ -56,15 +57,29 @@ public void execute(Task task) {
             }
         });
 
+        project.getTasks().create("fetchNotableNotes", NotableNotesFetcherTask.class, new Action<NotableNotesFetcherTask>() {
+            public void execute(NotableNotesFetcherTask task) {
+                final NotesGeneration gen = task.getNotesGeneration();
+                preconfigureNotableNotes(project, gen);
+
+                task.getOutputs().file(getTemporaryReleaseNotesFile(project));
```
szczepiq: I suggest to use Gradle annotations to drive the inputs/output, @OutputFile, @Input, etc. Check out Gradle docs on this :)

src/main/groovy/org/mockito/release/internal/gradle/DefaultContinuousDeliveryPlugin.java
```diff
@@ -315,11 +315,17 @@ public void execute(Task task) {
...
         task.getNotesGeneration().setTargetVersions(versionFile.getNotableVersions());
+
+        fetcherTask.getNotesGeneration().setTargetVersions(versionFile.getNotableVersions());
+        fetcherTask.getInputs().property("notableVersions", versionFile.getNotableVersions());
```
szczepiq: You can use @Input annotation instead, check out Gradle docs :)

Issue #76, opened 2017-04-17
New extension object for configuring the plugins


szczepiq:
Let's start using an extension object to configure the plugins. This is more idiomatic approach used by Gradle plugins and we should adhere to standards as much as possible. Here's how the configuration would look in Mockito. Please give feedback:

```gradle
//minimal
releasing {
    github {
        repository = "mockito/mockito"
        writeAuthUser = "szczepiq"
        readOnlyAuthToken = "a0a4c0f41c200f7c653323014d6a72a127764e17"
    }
}

//fully configured
releasing {
    github {
        repository = "mockito/mockito"
        writeAuthUser = "szczepiq"
        writeAuthToken = System.getenv("GH_WRITE_TOKEN")
        readOnlyAuthToken = "a0a4c0f41c200f7c653323014d6a72a127764e17"
    }
    releaseNotes {
        file = "doc/release-notes/official.md"
        notableFile = "doc/release-notes/notable.md"
        labelMapping = [
            '1.* incompatible': 'Incompatible changes with previous major version (v1.x)',
            'java-9': "Java 9 support",
            'java-8': "Java 8 support",
            'new feature': "New features",
            'BDD': 'Behavior-Driven Development support',
            'bug': "Bugfixes",
            'enhancement': "Enhancements",
            'android': "Android support",
            'docs': 'Documentation'
        ]
    }
    git {
        genericUser = "Mockito Release Tools"
        genericEmail = "<mockito.release.tools@gmail.com>"
        releasableBranchRegex = "master|release/.+"  // matches 'master', 'release/2.x', 'release/3.x', etc.
        branch = System.getenv("TRAVIS_BRANCH")
    }
    team {
        developers = ['szczepiq:Szczepan Faber', 'bric3:Brice Dutheil', 'raphw:Rafael Winterhalter', 'TimvdLippe:Tim van der Lippe']
        contributors = []
    }
}
```

mstachniuk:
It looks better then current configuration. 
For me `git.user` and `git.email` might be confused for what it is.
Maybe instead of `library` it should be `general` or `info` or `extras`?
What about bintray/maven upload repo settings?

szpak:
I also found it handy to use one extension (usually with multiple other classes used under the hood to map subclosures (`github`, `releaseNotes`) to keep code in a good shape), however there could be two topics to address:
[...]

PR #77, opened 2017-04-20
Safe, easy to use and well documented release configuration

szczepiq:
See rationale and example at #76 

Review suggestions:
 - focus on the design and public API suggestion at #76 
 - take a look at a couple of commits to see how the change is implemented. It's basically replacing 'ext' informal extension with formal extension object that has decent api.
 - it probably don't make sense to review such huge change line by line.

Thank you!


szczepiq:
Update the description, added review suggestions, please review :)

szczepiq:
Merging... the design was reviewed already at #76

Issue #36, opened 2017-03-18
Automatically include contributors in pom.xml file


szczepiq:
## Use case

In order to reward contributors and further motivate for more high quality contributions it would be great to list contributors in the [relevant section of pom.xml](http://maven.apache.org/pom.html#Contributors) file. As of 3/18/17 generation of pom.xml logic is a part of Mockito project automation and it is not generic enough to be part of mockito release tools.

## Design

This idea needs to be fleshed out further. An example technical design that we could try implementing:

- currently every contribution is explicitly mentioned in the release notes but it should also automatically include the contributor in the pom.xml of every future version of mockito.
- see [TODO with brainstorming ideas](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/internal/gradle/pom/DefaultPomPlugin.java#L17) in the code
- need a way to generate pom.xml template with contributor information for past releases so that we include contributors of the past (we can simplify and ignore contributors of 1.x version).
- Maven pom reference for "contributors" section: http://maven.apache.org/pom.html#Contributors

## Action plan

This is a great ticket for external contribution! Please consider helping out! 

- The next action is to think through the design, perhaps do some prototyping and comment back on this ticket with the technical design we actually want to execute. The previous design note was my quicky brainstorming.
- Submit PR! (@mstachniuk, wdyt?)

mstachniuk:
I think it can be done without any pom.xml template - it will be hard to keep template and pom.xml in sync. I will look more deeply on it in April.

szczepiq:
I've created initial pom plugin. Added a TODO to the code with [some examples](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/internal/gradle/pom/DefaultPomPlugin.java#L17) how to approach the problem :)

mstachniuk:
Design question
I fought about how to solve this problem. `GitHubContributorsProvider.mapContributorsToGitHubUser(...)` always ask only for contributors between two commits (usually last released version and HEAD). My first idea was to call this method inside `PomCustomizer.customizePom(...)`, but then only last contributors will be added to the pom.xml (and hardcoded in `gradle/release.gradle`).

[...]

PR #78, opened 2017-04-21
#36 read all project contributors and add them to pom.xml

mstachniuk:
TODO: 
- [x]  define dependencies between tasks better,
- [ ]  use releasing.team internal object for keeping data about contributors
- [x]  configuration option to disable adding contributors to pom.xml from github
- [x]  do not fetching data from GH if file with all project contributors exist (make it UP-TO-DATE)
- [ ]  remove old TODOs

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
@@ -33,7 +33,7 @@ import org.mockito.release.internal.gradle.util.PomCustomizer;
  * <ul>
  *     <li>Automatically includes "LICENSE" file in all jars.</li>
  *     <li>Adds build.dependsOn "publishToMavenLocal" to flesh out publication issues during the build</li>
- *     <li>Adds publishToMavenLocal.dependsOn "fetchAllContributorsFromGitHub" to load all contributors from GitHub</li>
+ *     <li>Adds publishToMavenLocal.dependsOn "fetchAllProjectContributorsFromGitHub" to load all contributors from GitHub</li>
  * </ul>
  */
 public class BaseJavaLibraryPlugin implements Plugin<Project> {
@@ -94,9 +94,9 @@ public class BaseJavaLibraryPlugin implements Plugin<Project> {
 
         //so that we flesh out problems with maven publication during the build process
         project.getTasks().getByName("build").dependsOn("publishToMavenLocal");
-        // TODO 1 is this dependency correct? Result of fetchAllContributorsFromGitHub is needed in
+        // TODO 1 is this dependency correct? Result of fetchAllProjectContributorsFromGitHub is needed in
         // TODO PomCustomizer.customizePom(project, publication);
         // TODO 2 Where define dependency to ContributorsPlugin?
-        project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllContributorsFromGitHub");
+        project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllProjectContributorsFromGitHub");
     }
 }
--- a/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
@@ -33,6 +33,7 @@ import org.mockito.release.internal.gradle.util.PomCustomizer;
  * <ul>
  *     <li>Automatically includes "LICENSE" file in all jars.</li>
  *     <li>Adds build.dependsOn "publishToMavenLocal" to flesh out publication issues during the build</li>
+ *     <li>Adds publishToMavenLocal.dependsOn "fetchAllContributorsFromGitHub" to load all contributors from GitHub</li>
  * </ul>
  */
 public class BaseJavaLibraryPlugin implements Plugin<Project> {
@@ -93,5 +94,9 @@ public class BaseJavaLibraryPlugin implements Plugin<Project> {
 
         //so that we flesh out problems with maven publication during the build process
         project.getTasks().getByName("build").dependsOn("publishToMavenLocal");
+        // TODO 1 is this dependency correct? Result of fetchAllContributorsFromGitHub is needed in
+        // TODO PomCustomizer.customizePom(project, publication);
+        // TODO 2 Where define dependency to ContributorsPlugin?
+        project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllContributorsFromGitHub");
     }
 }
--- a/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
@@ -41,6 +41,7 @@ public class BaseJavaLibraryPlugin implements Plugin<Project> {
     private final static Logger LOG = Logging.getLogger(BaseJavaLibraryPlugin.class);
 
     final static String PUBLICATION_NAME = "javaLibrary";
+    final static String POM_TASK = "generatePomFileFor" + PUBLICATION_NAME + "Publication";
 
     public void apply(final Project project) {
         final ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
--- a/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
@@ -40,7 +40,7 @@ public class BaseJavaLibraryPlugin implements Plugin<Project> {
 
     private final static Logger LOG = Logging.getLogger(BaseJavaLibraryPlugin.class);
 
-    final static String PUBLICATION_NAME = "javaLibrary";
+    final static String PUBLICATION_NAME = "JavaLibrary";
     final static String POM_TASK = "generatePomFileFor" + PUBLICATION_NAME + "Publication";
 
     public void apply(final Project project) {
--- a/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
@@ -97,6 +97,6 @@ public class BaseJavaLibraryPlugin implements Plugin<Project> {
         // TODO 1 is this dependency correct? Result of fetchAllProjectContributorsFromGitHub is needed in
         // TODO PomCustomizer.customizePom(project, publication);
         // TODO 2 Where define dependency to ContributorsPlugin?
-        project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllProjectContributorsFromGitHub");
+//         project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllProjectContributorsFromGitHub");
     }
 }
--- a/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
@@ -46,6 +46,19 @@ public class ContinuousDeliveryPlugin implements Plugin<Project> {
         project.getPlugins().apply(VersioningPlugin.class);
         project.getPlugins().apply(GitPlugin.class);
 
+        project.allprojects(new Action<Project>() {
+            @Override
+            public void execute(final Project project) {
+                project.getPlugins().withType(JavaLibraryPlugin.class, new Action<JavaLibraryPlugin>() {
+                    @Override
+                    public void execute(JavaLibraryPlugin plugin) {
+                        project.getTasks().getByName("publishToMavenLocal")
+                                .dependsOn(":fetchAllProjectContributorsFromGitHub");
+                    }
+                });
+            }
+        });
+
         final boolean notableRelease = project.getExtensions().getByType(VersionInfo.class).isNotableRelease();
 
         //TODO use constants for all task names
--- a/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
@@ -6,7 +6,6 @@ import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
-import org.gradle.api.publish.maven.tasks.GenerateMavenPom;
 import org.gradle.api.specs.Spec;
 import org.gradle.api.tasks.Exec;
 import org.gradle.process.ExecResult;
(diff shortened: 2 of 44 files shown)


src/main/groovy/org/mockito/release/notes/contributors/AllContributorsSerializer.java
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    public void serialize(ProjectContributorsSet contributorsSet) {
+        Collection<ProjectContributor> allContributors = contributorsSet.getAllContributors();
+        String json = Jsoner.serialize(allContributors);
+        LOG.info("Serialize contributors to: {}", json);
+        IOUtil.writeFile(file, json);
+    }
+
+    public ProjectContributorsSet desrialize() {
```
wwilk: typo

src/main/groovy/org/mockito/release/notes/contributors/GitHubAllContributorsFetcher.java
```diff
@@ -0,0 +1,89 @@
...
+        private final String authToken;
+
+        public GitHubProjectContributorsBuilder(String repository, String authToken) {
+            this.repository = repository;
+            this.authToken = authToken;
+        }
+
+        GitHubProjectContributors build() {
+            // see API doc: https://developer.github.com/v3/repos/#list-contributors
+            String nextPageUrl = String.format("%s%s",
```
wwilk: why not just simple concatenation instead of format in this case?

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
```diff
@@ -90,5 +91,9 @@ public void execute(MavenPublication publication) {
 
         //so that we flesh out problems with maven publication during the build process
         project.getTasks().getByName("build").dependsOn("publishToMavenLocal");
+        // TODO 1 is this dependency correct? Result of fetchAllContributorsFromGitHub is needed in
+        // TODO PomCustomizer.customizePom(project, publication);
+        // TODO 2 Where define dependency to ContributorsPlugin?
```
szczepiq: I suggest we do it this way:

1. Contributors are meaningful in the scope of the entire project. So it makes sense to just have one fetcher task for the entire project.
2. ContinuousDeliveryPlugin is intended to be applied in root project. So perhaps it can apply the contributors plugin? Then it could do something like:

```java
            public void execute(final Project project) {
                //...
                project.getPlugins().withType(BaseJavaLibraryPlugin.class, new Action<JavaLibraryPlugin>() {
                    public void execute(JavaLibraryPlugin plugin) {
                        project.getTasks().getByName("dudeThatGeneratesPom").dependsOn(fetcherTask);
                    }
                });
            }
        });
```

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
```diff
@@ -82,13 +85,14 @@ class PomCustomizer {
...
-            def contributors = root.appendNode('contributors')
-            rootProject.ext.pom_contributors.each {
-                def split = it.split(':')
-                assert split.length == 2
-                def c = contributors.appendNode('contributor')
-                c.appendNode('name', split[1])
-                c.appendNode('url', "https://github.com/${split[0]}")
+            def contributorsNode = root.appendNode('contributors')
+            def fileName = Contributors.getAllProjectContributorsFileName(project.getBuildDir())
+            AllProjectContributorsReader reader = AllProjectsContributorsProvider.allProjectContributorsReader()
```
szczepiq: Instead of reading the file on the fly here, every time we generate the pom, I suggest an alternative approach:

1. We add the fetcher task that fetches from GitHub and write the file. This task is incremental and does not fetch from github if file was already fetched. The task should be on root project only (continuous delivery plugin?) because we only want to fetch once for the entire project, regardless of how many jars we produce.
2. We add 'teamConfigurer' task (depends on fetcher) that loads the file and configures the releasing.team object (needs my PR to be merged first https://github.com/mockito/mockito-release-tools/pull/77). This task is not incremental because it configures internal state (releasing.team object). For the same reasons as previous task, this task should be on the root project, single task for the entire build.
3. We make the every 'generatePom' task depend on teamConfigurer. We can do it continuous delivery plugin, using 'allprojects' and 'withType' trick

The added benefit of doing it this way is that users can easily disable the tasks and explicitly configure developers and contributors on the 'releasing.team' extension.

What do you think?

src/test/groovy/org/mockito/release/notes/contributors/AllContributorsSerializerTest.groovy
```diff
@@ -0,0 +1,118 @@
+package org.mockito.release.notes.contributors
+
+import org.mockito.release.notes.model.ProjectContributor
+import spock.lang.Specification
+import spock.lang.Subject
+
+class AllContributorsSerializerTest extends Specification {
```
szczepiq: Nice test. Thanks!

src/test/groovy/org/mockito/release/notes/contributors/AllContributorsSerializerTest.groovy
```diff
@@ -0,0 +1,118 @@
...
+        when:
+        serializer.serialize(contributors)
+        def actual = serializer.desrialize()
+
+        then:
+        actual.getAllContributors().containsAll(contributors.getAllContributors())
+        contributors.getAllContributors().containsAll(actual.getAllContributors())
+    }
+
+    def "serialization and deserialization of 5 contributors"() {
```
szczepiq: Does testing against 5 sample contributors adds value to the coverage? I suspect that 2 sample contributors are enough to cover all complexity / code paths. I suggest to keep tests simple and test with as little sample data as possible to prove that the feature is working!

src/test/groovy/org/mockito/release/notes/contributors/GitHubAllContributorsJsonTest.groovy
```diff
@@ -0,0 +1,45 @@
+package org.mockito.release.notes.contributors
+
+import org.json.simple.JsonObject
+import spock.lang.Specification
+
+class GitHubAllContributorsJsonTest extends Specification {
```
szczepiq: Really nice coverage, thank you!

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
```diff
@@ -90,5 +91,9 @@ public void execute(MavenPublication publication) {
 
         //so that we flesh out problems with maven publication during the build process
         project.getTasks().getByName("build").dependsOn("publishToMavenLocal");
```
szpak: Is it possible to use the "classic" `uploadArchives` from the `maven` plugin?

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
szpak: Is it possible to stay with the old approach and have only defined developers in `pom.xml` (there is a chance that I missed a switch in some other part of code)?

szczepiq: Good point, we need to make the implementation more flexible. It should be easy to just configure the users explicitly if the users want such behavior. I made a comment about it in the code.

Thanks for review!

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
szczepiq: I think this question is unrelated @mstachniuk PR. The plugins are opinionated and they use new publishing plugin. If someone wants to use classic upload archives then he would have to avoid using our plugins and wire our tasks himself. That's not very convenient. If we have use cases / users asking for supporting classic publication, we can add it. Perhaps you can contribute? :)

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
mstachniuk: @szpak In pom.xml are 2 sections: developers and contributors. Developers come from static configuration (see line above: `def developers = root.appendNode('developers')`). Contributors are fetched from GitHub. I will add some configuration property to disable this feature and combine contributors with old approach list.
I need also remove developers from contributors to don't duplicate them.

szpak: @mstachniuk 👍

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
szpak: @szczepiq It is unrelated. I was just curios seeing that code. Once, I will come back to CD in FOSS there is a chance for contribution ;).

src/main/groovy/org/mockito/release/notes/contributors/GitHubAllContributorsFetcher.java
mstachniuk: to see difference between base url and parameters

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -46,6 +46,19 @@ public void apply(final Project project) {
         project.getPlugins().apply(VersioningPlugin.class);
         project.getPlugins().apply(GitPlugin.class);
 
+        project.allprojects(new Action<Project>() {
+            @Override
+            public void execute(final Project project) {
+                project.getPlugins().withType(JavaLibraryPlugin.class, new Action<JavaLibraryPlugin>() {
+                    @Override
+                    public void execute(JavaLibraryPlugin plugin) {
+                        project.getTasks().getByName("publishToMavenLocal")
```
szczepiq: It is safer to find the exact task that generates pom file and leverages customizePom method. Otherwise fetchAllProjectContributorsFromGitHub might get executed too late, after the poms are generated but before publishToMavenLocal runs.

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -46,6 +46,19 @@ public void apply(final Project project) {
...
+                                .dependsOn(":fetchAllProjectContributorsFromGitHub");
```
szczepiq: This is making an assumption that the root project will have "fetchAllProjectContributorsFromGitHub" task. I suggest to make it a little bit more flexible:
1. ContinuousDeliveryPlugin can apply the ContributorsPlugin that adds fetchAllProjectContributorsFromGitHub task
2. Then, here you can just configure

//subproject - project in the 
subproject.getTasks().getByName("publishToMavenLocal").dependsOn(project.getTasks().getByName("fetcher")

It's a lot of writing... let me just push a commit on your branch :)

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
```diff
@@ -84,13 +83,7 @@ class PomCustomizer {
...
             def contributorsNode = root.appendNode('contributors')
             def fileName = Contributors.getAllProjectContributorsFileName(project.getBuildDir())
-            AllProjectContributorsReader reader = AllProjectsContributorsProvider.allProjectContributorsReader()
-            def contributors = reader.loadAllContributors(fileName)
-            contributors.getAllContributors().each {
-                def c = contributorsNode.appendNode('contributor')
-                c.appendNode('name', it.name != "" ? it.name : it.login)
-                c.appendNode('url', it.profileUrl)
-            }
+            ContributorsToPom.include(contributorsNode, fileName, conf.team.contributors, conf.team.developers)
```
szczepiq: I _really_ like the idea of pushing out this complexity to the separate Java class. It not only streamlines testing but also we have a chance to move some code from Groovy to Java!!!

src/test/groovy/org/mockito/release/notes/contributors/AllContributorsSerializerTest.groovy
```diff
@@ -45,29 +44,7 @@ class AllContributorsSerializerTest extends Specification {
...
-        contributors.addContributor(contributor1)
-        contributors.addContributor(contributor2)
-        contributors.addContributor(contributor3)
-        contributors.addContributor(contributor4)
-        contributors.addContributor(contributor5)
-
-        when:
-        serializer.serialize(contributors)
-        def actual = serializer.desrialize()
+        def actual = serializer.deserialize()
```
szczepiq: Thank you for making the tests simpler!!!

src/test/groovy/org/mockito/release/notes/contributors/ContributorsToPomTest.groovy
```diff
@@ -0,0 +1,98 @@
...
+        sut.include(node, contributorsSet, contributorsStringList, developers)
+
+        then:
+        node.children().get(0).get("name").text() == "name 1"
+        node.children().get(0).get("url").text() == "https://github.com/login1"
+        node.children().get(1).get("name").text() == "name 2"
+        node.children().get(1).get("url").text() == "https://github.com/login2"
+    }
+
+    def "should ignore contributor from set when it is a developer but check only login"() {
```
szczepiq: Nice! Thank you for covering all the use cases!

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
mstachniuk: Sounds good. I will continue soon.

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
```diff
@@ -97,6 +97,6 @@ public void execute(MavenPublication publication) {
         // TODO 1 is this dependency correct? Result of fetchAllProjectContributorsFromGitHub is needed in
         // TODO PomCustomizer.customizePom(project, publication);
         // TODO 2 Where define dependency to ContributorsPlugin?
-        project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllProjectContributorsFromGitHub");
+//         project.getTasks().getByName("publishToMavenLocal").dependsOn("fetchAllProjectContributorsFromGitHub");
```
mstachniuk: I will remove in next commits

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
mstachniuk: I tried to find exactly task for that (by throwing exception from PomCustomizer), and it was probably: `generatePomFileForJavaLibraryPublication` (in Mockito project). I didn't find this task in mockito-release-tools. Too much magic happens there. Could you help me with this?

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
mstachniuk: Thank you.

src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
mstachniuk: Your welcome!

mstachniuk:
I need to fix a few things, a whole list of TODOs is on the top of this  PR.

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
szczepiq: I think "generatePomFileForJavaLibraryPublication" is the correct one. It is automagically generated from "generatePomFileFor" + publication name (specified in the BaseJavaLibraryPlugin) + "Publication". Fancy :)

I suggest to make a constant for this String inside BaseJavaLibraryPlugin because it will make the couplings between the plugins more clear.

Nice, you're learning how to hack & understand Gradle :)

szczepiq:
I started reviewing and pushed some changes to your branch. It could be useful for you so that you get up to speed with Gradle / tasks, etc.

I'll complete review by the end of today (and potentially push more to your branch). I don't test eagerly the stuff I push to your branch (blushing... just a little bit). My main goal is to get you up-to-speed with coding & design principles when working with Gradle plugins. I hope it's ok, if I mess something up with your code, let me know :)

szczepiq:
Made some tweaks :) It's ready!

mstachniuk:
Because of a many changes it's hard to follow all comments. I closed this Pull Request and will create a new one.

Issue #71, opened 2017-04-12
fetchNotableReleaseNotes fails ungracefully when incorrect tag is supplied


szczepiq:
fetchNotableReleaseNotes task fails with below stack trace when one of the input tags is missing. Let's have more graceful handling of this scenario, the exception message should indicate a potential problem and suggest a resolution.

Stack trace:

```
:fetchNotableReleaseNotes FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':fetchNotableReleaseNotes'.
> Problems parsing date: [fatal: ambiguous argument 'v2.7.0': unknown revision or path not in the working tree.
  Use '--' to separate paths from revisions, like this:
  'git <command> [<revision>...] -- [<file>...]']. Required format is: [yyyy-MM-dd HH:mm:ss Z].

* Try:
Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.

BUILD FAILED
```

szczepiq:
@wwilk, something you can take look?

mstachniuk:
This tag is missing in Mockito project, but it's in version.properties

12 kwi 2017 7:17 AM "Szczepan Faber" <notifications@github.com> napisał(a):

szczepiq:
@wwilk, is this completed? Can you close if you believe it's don? Thank you! :)

wwilk:
Yes, it's completed. Closing it!

PR #80, opened 2017-04-21
RevisionDateProvider fails gracefully when revision number is incorrect #71

wwilk:
@szczepiq

### DIFF
--- a/src/main/groovy/org/mockito/release/notes/vcs/RevisionDateProvider.java
+++ b/src/main/groovy/org/mockito/release/notes/vcs/RevisionDateProvider.java
@@ -3,6 +3,7 @@ package org.mockito.release.notes.vcs;
 import org.mockito.release.exec.ProcessRunner;
 
 import java.util.Date;
+import java.util.regex.Pattern;
 
 import static org.mockito.release.notes.internal.DateFormat.parseDate;
 
@@ -12,6 +13,12 @@ import static org.mockito.release.notes.internal.DateFormat.parseDate;
 class RevisionDateProvider {
 
     private final ProcessRunner runner;
+    /**
+     * pattern for validating revision date in ISO format
+     * Example of valid date: 2017-01-29 08:14:09 -0800
+     */
+    private final Pattern REVISION_DATE_PATTERN = Pattern.compile(
+            "\\s?\\d{4}-\\d{2}-\\d{2} \\d{2}:\\d{2}:\\d{2}\\s[-+]\\d{4}\\s?");
 
     RevisionDateProvider(ProcessRunner runner) {
         this.runner = runner;
@@ -19,9 +26,10 @@ class RevisionDateProvider {
 
     public Date getDate(String rev) {
         String gitOutput = runner.run("git", "log", "--pretty=%ad", "--date=iso", rev, "-n", "1");
-        //TODO process runner needs to fail on error here.
-        // Otherwise the 'date' variable holds some error message and not really any date.
-        //Example output returned by running git command: 2017-01-29 08:14:09 -0800
+        if(!REVISION_DATE_PATTERN.matcher(gitOutput).matches()){
+           throw new IllegalArgumentException("Can't get a proper date for revision number " + rev + ". Are you sure this revision exists?");
+        }
+
         return parseDate(gitOutput.trim());
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/mockito/release/notes/vcs/RevisionDateProvider.java
```diff
@@ -12,16 +13,23 @@
...
         this.runner = runner;
     }
 
     public Date getDate(String rev) {
         String gitOutput = runner.run("git", "log", "--pretty=%ad", "--date=iso", rev, "-n", "1");
-        //TODO process runner needs to fail on error here.
-        // Otherwise the 'date' variable holds some error message and not really any date.
-        //Example output returned by running git command: 2017-01-29 08:14:09 -0800
+        if(!REVISION_DATE_PATTERN.matcher(gitOutput).matches()){
+           throw new IllegalArgumentException("Can't get a proper date for revision number " + rev + ". Are you sure this revision exists?");
```
szczepiq: "this revision exists?" -> "this revision or tag exists?"

szczepiq: Minor, it might be good to call out 'tag' explicitly here to make it easy to grasp by users: "Are you sure this revision exists?" -> "Are you sure this revision or tag exists?"

src/main/groovy/org/mockito/release/notes/vcs/RevisionDateProvider.java
```diff
@@ -12,16 +13,23 @@
...
+        }
+
         return parseDate(gitOutput.trim());
```
szczepiq: For consideration: perhaps it's simpler if we just catch exception from parseDate instead of doing regexp validation. If we catch any exception it means the date is coming in invalid format or that output from git is unexpected.

src/main/groovy/org/mockito/release/notes/vcs/RevisionDateProvider.java
szczepiq: Might be worth adding the message we got from git to the exception message. Something like that:

- i cannot extract the date for revision 'xxx' from git log. Are you sure the revision/tag exists? Following output was returned by git:\n xxxxx

wwilk:
Done!

wwilk:
Of course, I tested it with nonexistent tag. Error looks like this: 

> fatal: ambiguous argument 'v2.7.0': unknown revision or path not in the working tree.
>   Use '--' to separate paths from revisions, like this:
>   'git <command> [<revision>...] -- [<file>...]'

from parseDate you get: 

> Problems parsing date: [fatal: ambiguous argument 'v2.7.0': unknown revision or path not in the working tree.
>   Use '--' to separate paths from revisions, like this:
>   'git <command> [<revision>...] -- [<file>...]']. Required format is: [yyyy-MM-dd HH:mm:ss Z].

Having exact git message seems really useful! Maybe we can keep regex check and just append gitOutput to error message? It seems easier (and clearer) than extracting git message from parseDate exception.

szczepiq:
>Having exact git message seems really useful! Maybe we can keep regex check and just append gitOutput to error message? It seems easier (and clearer) than extracting git message from parseDate exception.

Sure!

wwilk:
Done :)

src/test/groovy/org/mockito/release/notes/vcs/RevisionDateProviderTest.groovy
```diff
@@ -12,9 +12,21 @@ class RevisionDateProviderTest extends Specification {
...
+    }
+
+    def "fails if revision number is incorrect what causes git call to fail"() {
+        runner.run("git", "log", "--pretty=%ad", "--date=iso", "v1.0.0", "-n", "1") >> "fatal: ambiguous argument 'v1.0.0'"
+
+        when:
+        provider.getDate("v1.0.0")
+
+        then:
+        thrown IllegalArgumentException
```
szczepiq: For extra safety, let's add a simple assertion message on the exception content. We can assert on full exception message or part of it, whatever is simpler.

Otherwise, we might not be testing the right thing. IllegalArgumentException is a generic exception that can be thrown for other reasons than the one that we expect in this test.

Issue #22, opened 2017-03-05
Enable consistent code style validation


szczepiq:
To keep the code clean and also to maximize the value of the code review, let's make the tools validate consistent code style for us.

- [x] enable Gradle Checkstyle plugin and make the build fail on coding style issues
- [x] enable Gradle Codenarc plugin, fail on issues
- [x] clean up the codebase and submit PR
- [ ] ensure it is enabled for test code (see https://github.com/mockito/shipkit/issues/22#issuecomment-333379214)

szczepiq:
@epeee, can we close this ticket after your PR was merged (see the checklist :)?

Thank you for contribution!!!

epeee:
@szczepiq yes, I think so.

You're welcome!

szczepiq:
Closing with pleasure :) BTW. do you have a twitter handle?

epeee:
@szczepiq no twitter handle (yet).

szczepiq:
Hmmm, can we check if this is enabled for test code? I just stumbled upon PluginSpecification class that had incorrect whitespace and Checkstyle did not detect it.

Let's keep the house clean! :)

epeee:
I will have look at this one.

szczepiq:
Thank you!

PR #83, opened 2017-04-23
Enable consistent code style validation

epeee:
#22 added checkstyle and codenarc plugin + added some modules/rulesets.


szczepiq:
Very nice contribution, thank you!

I will merge this with pleasure. Consistent code style ensures that code review focuses on essential complexity rather than on whitespace nitpicking.

Can you make it so that the build fails when the code style rule is violated? (please ignore if it already does it :).

Issue #70, opened 2017-04-12
Applying the plugin on project without 'version.properties' file yields stack trace


szczepiq:
Applying one of our plugins, "org.mockito.mockito-release-tools.versioning" or "org.mockito.mockito-release-tools.continuous-delivery" plugin in a project that does not have 'version.properties' file yields nasty stack trace. We need to handle this use case gracefully. Example solutions:

- fail with decent exception telling the user what to do
- generate the file for the user and log information that the file should be checked-in
- create some sort of bootstrap task that will setup all necessary files for getting continuos delivery set up for a project

wwilk:
Done

PR #89, opened 2017-04-23
version.properties file created automatically if doesnt exist #70

wwilk:
Version.properties is created automatically and set to project.version. 

I also added a unit test for checking if extensions are correctly set (that's why I removed TODO line).

@szczepiq @mstachniuk 
What do you think about the case when user doesn't have project.version set at the moment when VersioningPlugin is being configured? Project.version returns "unspecified" then and we have a failure a little later (in VersionBumper#incrementVersion) because "unspecified" doesn't comply with our regex pattern for version.

Would setting the version to '0.1.0' be too much of intrusion into the project?

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
@@ -39,7 +39,9 @@ public class VersioningPlugin implements Plugin<Project> {
 
     private static Logger LOG = Logging.getLogger(VersioningPlugin.class);
 
-    public final static String VERSION_FILE_NAME = "version.properties";
+    public static final String VERSION_FILE_NAME = "version.properties";
+    public static final String GRADLE_DEFAULT_VERSION = "unspecified";
+    public static final String FALLBACK_INITIAL_VERSION = "0.0.1";
 
     public void apply(Project project) {
         final File versionFile = project.file(VERSION_FILE_NAME);
@@ -71,11 +73,24 @@ public class VersioningPlugin implements Plugin<Project> {
 
     private void createVersionPropertiesFile(Project project, File versionFile) {
         LOG.lifecycle("  Required file version.properties doesn't exist. Creating it automatically. Remember about checking it into VCS!");
-        LOG.lifecycle("  Initial project version in version.properties set to {}", project.getVersion());
+        LOG.lifecycle("  You shouldn't configure project.version in build.gradle anymore. Version from version.properties will be used instead.");
+
+        String version = determineVersion(project);
+
         String versionFileContent = "#Version of the produced binaries. This file is intended to be checked-in.\n"
                 + "#It will be automatically bumped by release automation.\n"
-                + "version=" + project.getVersion() + "\n";
+                + "version=" + version + "\n";
 
         IOUtil.writeFile(versionFile, versionFileContent);
+    }
+
+    private String determineVersion(Project project){
+        if(GRADLE_DEFAULT_VERSION.equals(project.getVersion()) ){
+            LOG.lifecycle("  BEWARE! Project.version is unspecified. Version will be set to {}. You can change it manually in version.properties.", FALLBACK_INITIAL_VERSION);
+            return FALLBACK_INITIAL_VERSION;
+        } else{
+            LOG.lifecycle("  Initial project version in version.properties set to {} (taken from project.version property).", project.getVersion());
+            return project.getVersion().toString();
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
```diff
@@ -62,4 +68,14 @@ public void execute(BumpVersionFileTask t) {
             }
         });
     }
+
+    private void createVersionPropertiesFile(Project project, File versionFile) {
+        LOG.lifecycle("  Required file version.properties doesn't exist. Creating it automatically. Remember about checking it into VCS!");
```
szczepiq: Let's also write that the user _should not_ configure the version in the 'build.gradle' file. Only version.properties should contain the version.

src/test/groovy/org/mockito/release/internal/gradle/VersioningPluginTest.groovy
```diff
@@ -0,0 +1,61 @@
...
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import org.mockito.release.version.VersionInfo
+import spock.lang.Specification
+
+class VersioningPluginTest extends Specification {
+
+    @Rule TemporaryFolder tmp = new TemporaryFolder()
+    def project
+    def initialVersionFileContent = "version=1.0.0\nnotableVersions=0.1.0\n"
```
szczepiq: I suggest to inline this variable. It's nice to avoid duplication. Problem is that tests are easier to read and maintain if they have a little bit of conscious duplication. DAMP VS DRY principle for unit tests :)

src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
szczepiq: - [ ] Let's also write that the user _should not_ configure the version in the 'build.gradle' file. Only version.properties should contain the version.

src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
```diff
@@ -62,4 +68,14 @@ public void execute(BumpVersionFileTask t) {
...
+        LOG.lifecycle("  Initial project version in version.properties set to {}", project.getVersion());
+        String versionFileContent = "#Version of the produced binaries. This file is intended to be checked-in.\n"
+                + "#It will be automatically bumped by release automation.\n"
+                + "version=" + project.getVersion() + "\n";
```
szczepiq: Great catch on the problem of 'unspecified' version! Let's handle this situation cleanly. I'll suggest and approach in the comment to the PR.

src/test/groovy/org/mockito/release/internal/gradle/VersioningPluginTest.groovy
```diff
@@ -0,0 +1,61 @@
+package org.mockito.release.internal.gradle
+
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import org.mockito.release.version.VersionInfo
+import spock.lang.Specification
+
+class VersioningPluginTest extends Specification {
```
szczepiq: Nice test coverage. Thank you for keeping the bar high!

src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
```diff
@@ -30,6 +31,9 @@
  * </ul>
  *
  * Also, the plugin configures all projects' version property to the value specified in "version.properties"
+ *
+ * BEWARE! If version.properties doesn't exists, this plugin will create it automatically and set
```
szczepiq: Nice javadoc! The Javadoc descriptions of our plugins will be a part of our official documentation. It's great to describe exactly what happens in the plugin.

szczepiq:
>Would setting the version to '0.1.0' be too much of intrusion into the project?

Great catch! I suggest '0.0.1' because it's the lowest possible.

Other alternatives to handling this scenario will make everything complicated / inconvenient to use:
 - use deferred configuration / after evaluate to wait for user version setting
 - fail early instead of creating the file

wwilk:
OK, I set it to 0.0.1 if no other version specified. Other suggestions taken too :)

mstachniuk:
What do you think to use current version defined in build.gradle if
version.properties doesn't exists?


23 kwi 2017 11:29 PM "Wojtek Wilk" <notifications@github.com> napisał(a):

wwilk:
@mstachniuk 
That's exactly what I'm doing here, if there is project.version specified. My question regards the opposite case - when there is no project.version specified. I like this solution with setting it to 0.0.1. What do you think?

mstachniuk:
Ahhh it's fine to set it to 0.0.1 when project version doesn't exist.

24 kwi 2017 12:13 AM "Wojtek Wilk" <notifications@github.com> napisał(a):

PR #90, opened 2017-04-23
pre-commit hook + tasks for installing this hook #22

wwilk:
@szczepiq @mstachniuk 

Since we already have codenarc and checkstyle validation, I added proposition of pre-commit hook and tasks to install it in .git/hooks. No tasks depend on it currently, maybe we should make it dependent of 'install' or 'clean'? 

What do you think?

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -16,6 +16,8 @@ apply plugin: 'maven'
 apply plugin: 'checkstyle'
 apply plugin: 'codenarc'
 
+apply from: 'gradle/precommit.gradle'
+
 sourceCompatibility = 1.6
 targetCompatibility = 1.6
 
@@ -86,16 +88,6 @@ task travisRelease {
                     "-Pgradle.publish.secret=${System.getenv('GRADLE_PUBLISH_SECRET')}"
         }
     }
-}
-
-task installGitPreCommitHook(type: Copy) {
-    from 'pre-commit'
-    into '.git/hooks'
-}
-
-task makeGitHooksExecutable { Task t ->
-    t.dependsOn(installGitPreCommitHook)
-    Runtime.getRuntime().exec("chmod -R +x .git/hooks/");
 }
 
 task fastInstall { Task t ->
--- a/build.gradle
+++ b/build.gradle
@@ -88,6 +88,16 @@ task travisRelease {
     }
 }
 
+task installGitPreCommitHook(type: Copy) {
+    from 'pre-commit'
+    into '.git/hooks'
+}
+
+task makeGitHooksExecutable { Task t ->
+    t.dependsOn(installGitPreCommitHook)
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -88,6 +88,16 @@ task travisRelease {
...
 }
 
+task installGitPreCommitHook(type: Copy) {
+    from 'pre-commit'
+    into '.git/hooks'
+}
+
+task makeGitHooksExecutable { Task t ->
+    t.dependsOn(installGitPreCommitHook)
+    Runtime.getRuntime().exec("chmod -R +x .git/hooks/");
```
szczepiq: You're missing 'doLast' action here! Without it, the 'exec' is triggered during Gradle's 'configuration time', that is every time Gradle prepares for running any task :)

- [ ] - I suggest to use ExecTask for "makeGitHooksExecutable". See Gradle docs on ExecTask.

build.gradle
```diff
@@ -88,6 +88,16 @@ task travisRelease {
     }
 }
 
+task installGitPreCommitHook(type: Copy) {
```
szczepiq: - [ ] - Can you move those 2 tasks into "gradle/precommit.gradle" script and use "apply from" to load the functionality? This way, we encapsulate complexity in separate files and keep the code clean and easy to understand.

build.gradle
```diff
@@ -88,6 +88,16 @@ task travisRelease {
...
+    from 'pre-commit'
+    into '.git/hooks'
+}
+
+task makeGitHooksExecutable { Task t ->
```
szczepiq: - [ ] - I suggest a simple name like "installPrecommit". Easier to use :)

- [ ] - Please add descriptions to the tasks and group "precommit", they are visible when one runs "gradle tasks" and make the project easier to work with.

build.gradle
```diff
@@ -88,6 +88,16 @@ task travisRelease {
...
+    t.dependsOn(installGitPreCommitHook)
+    Runtime.getRuntime().exec("chmod -R +x .git/hooks/");
+}
+
```
szczepiq: - [ ] - Can you also add "uninstallPrecommit" task? This way we can uninstall easily in case the feature is problematic.

mstachniuk:
General question. With this hook a gradle command will be start twice on CI
server:
1. defined in .travis.yml
2. during commit changes
This will slow down a whole build, isn't it? Or the second call will be
fast?
Isn't better to configure a build to fail-fast? WDYT?

Anyway there should be an option to disable this feature, to be more
flexible.

24 kwi 2017 6:55 AM "Szczepan Faber" <notifications@github.com> napisał(a):

szczepiq:
Thank you for addressing all my feedback! Absolutely merging :)

Can you add a note in README.md about the experiment with git hooks and how folks can use it (if they want :).

Issue #92, opened 2017-04-24
Configurable commit message used by release tools automation


szczepiq:
### Problem

The message used when release automation commits code [is hardcoded](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java#L147) in the plugin implementation. In Mockito project, we have a use case where we want to run the build even for the automated version bumps. We want to do it to work around the issue with codecov (https://github.com/mockito/mockito/issues/984).

### Suggested implementation

Make the [automated commit message](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java#L147)  configurable and settable via new "releasing.git.commitMessagePostfix" configuration setting. Full-blown example:

```Gradle
releasing {
  def buildNo = System.getenv("TRAVIS_BUILD_NUMBER")
  git.commitMessagePostfix = buildNo? " by CI build $buildNo [ci skip-release]" : " [ci skip-release]"
}
```

### Test coverage 

- let's use the new configuration setting in the release-tools-example project
- unit tests in the code

szczepiq:
Thank you very much @epeee for implementation and pull request!

PR #93, opened 2017-04-24
Configurable commit message used by release tools automation

epeee:
solves #92

already did some local testing using mockito-release-tools-example project.

### DIFF
--- a/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
@@ -28,6 +28,7 @@ public class ReleaseConfiguration {
         //Configure default values
         this.git.setTagPrefix("v"); //so that tags are "v1.0", "v2.3.4"
         this.git.setReleasableBranchRegex("master|release/.+");  // matches 'master', 'release/2.x', 'release/3.x', etc.
+        this.git.setCommitMessagePostfix(" [ci skip]");
     }
 
     private boolean dryRun = true;
@@ -255,6 +256,21 @@ public class ReleaseConfiguration {
          */
         public void setTagPrefix(String tagPrefix) {
             configuration.put("git.tagPrefix", tagPrefix);
+        }
+
+
+        /**
+         * Postfix which will be added to the commit message.
+         */
+        public String getCommitMessagePostfix() {
+            return getString("git.commitMessagePostfix");
+        }
+
+        /**
+         * See {@link #getCommitMessagePostfix()}
+         */
+        public void setCommitMessagePostfix(String commitMessagePostfix) {
+            configuration.put("git.commitMessagePostfix", commitMessagePostfix);
         }
     }
 
--- a/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
@@ -260,7 +260,8 @@ public class ReleaseConfiguration {
 
 
         /**
-         * Postfix which will be added to the commit message.
+         * Text which will be included in the commit message for all commits automatically created by the release
+         * automation.
          */
         public String getCommitMessagePostfix() {
             return getString("git.commitMessagePostfix");
--- a/src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java
@@ -39,22 +39,29 @@ public class GitPlugin implements Plugin<Project> {
         TaskMaker.execTask(project, COMMIT_TASK, new Action<Exec>() {
             public void execute(final Exec t) {
                 t.setDescription("Commits staged changes using generic --author");
-                t.doFirst(new Action<Task>() {
-                    public void execute(Task task) {
-                        //doFirst (execution time) to pick up user-configured setting
+                lazyConfiguration(t, new Runnable() {
+                    @Override
+                    public void run() {
                         t.commandLine("git", "commit", "--author",
-                                GitUtil.getGitGenericUserNotation(conf), "-m", commitMessage("Bumped version and updated release notes"));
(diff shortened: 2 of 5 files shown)


src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
```diff
@@ -256,6 +257,21 @@ public String getTagPrefix() {
...
+
+        /**
+         * See {@link #getCommitMessagePostfix()}
+         */
+        public void setCommitMessagePostfix(String commitMessagePostfix) {
+            configuration.put("git.commitMessagePostfix", commitMessagePostfix);
+        }
     }
 
     /**
```
szczepiq: Very nice & clean changes to this public API class. Thank you!

src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java
```diff
@@ -39,22 +39,29 @@ public void apply(final Project project) {
         TaskMaker.execTask(project, COMMIT_TASK, new Action<Exec>() {
             public void execute(final Exec t) {
                 t.setDescription("Commits staged changes using generic --author");
-                t.doFirst(new Action<Task>() {
-                    public void execute(Task task) {
-                        //doFirst (execution time) to pick up user-configured setting
+                lazyConfiguration(t, new Runnable() {
```
szczepiq: Can you use deferredConfiguration here instead of lazyConfiguration? Please take a look at the javadocs for deferredConfiguration & lazyConfiguration. I will add a TODO for myself to explain when we should use lazyConfiguration and when we should use deferredConfiguration.

src/main/groovy/org/mockito/release/internal/gradle/util/GitUtil.java
```diff
@@ -46,4 +46,11 @@ public static Object getGitGenericUserNotation(ReleaseConfiguration conf) {
     public static String getTag(ReleaseConfiguration conf, Project project) {
         return conf.getGit().getTagPrefix() + project.getVersion();
     }
+
+    /**
+     * Returns Git commit message based on release configuration and the given message
+     */
+    public static String getCommitMessage(ReleaseConfiguration conf, String message) {
```
szczepiq: Good idea to move this method here and add unit test. This makes the code easier to test and the solution is cleaner, too.

src/test/groovy/org/mockito/release/internal/gradle/GitPluginTest.groovy
```diff
@@ -0,0 +1,28 @@
+package org.mockito.release.internal.gradle
+
+import org.gradle.testfixtures.ProjectBuilder
+import spock.lang.Specification
+
+
+class GitPluginTest extends Specification {
```
szczepiq: I love the smell of clean tests in the morning :)

src/test/groovy/org/mockito/release/internal/gradle/GitPluginTest.groovy
```diff
@@ -0,0 +1,28 @@
...
+        expect:
+        project.releasing.git.commitMessagePostfix == " [ci skip]"
+    }
+
+    def "custom commitMessagePostfix"() {
+        def buildNo = 1234
+        project.releasing.git.commitMessagePostfix = buildNo? " by CI build $buildNo [ci skip-release]" : " [ci skip-release]"
+
+        expect:
+        project.releasing.dryRun == true
```
szczepiq: not needed here? less lines of code in tests == tests easier to grasp!

src/test/groovy/org/mockito/release/internal/gradle/GitPluginTest.groovy
```diff
@@ -0,0 +1,28 @@
...
+    }
+
+    def "default commitMessagePostfix"() {
+        expect:
+        project.releasing.git.commitMessagePostfix == " [ci skip]"
+    }
+
+    def "custom commitMessagePostfix"() {
+        def buildNo = 1234
+        project.releasing.git.commitMessagePostfix = buildNo? " by CI build $buildNo [ci skip-release]" : " [ci skip-release]"
```
szczepiq: to keep the tests as simple as possible, lets avoid conditional logic from tests. I suggest to just configure the postfix to fixed value and assert on it.

src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
```diff
@@ -256,6 +257,21 @@ public String getTagPrefix() {
         public void setTagPrefix(String tagPrefix) {
             configuration.put("git.tagPrefix", tagPrefix);
         }
+
+
+        /**
+         * Postfix which will be added to the commit message.
```
szczepiq: "Postfix which will be added to the commit message." -> "Text which will be included in the commit message for all commits automatically created by the release automation"

This will make it a bit more clearer for the users.

src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java
```diff
@@ -39,22 +39,29 @@ public void apply(final Project project) {
...
 
         TaskMaker.execTask(project, TAG_TASK, new Action<Exec>() {
-            public void execute(Exec t) {
+            public void execute(final Exec t) {
                 t.mustRunAfter(COMMIT_TASK);
-                String tag = "v" + project.getVersion();
+                final String tag = "v" + project.getVersion();
                 t.setDescription("Creates new version tag '" + tag + "'");
-                t.commandLine("git", "tag", "-a", tag, "-m", commitMessage("Created new tag " + tag));
+                lazyConfiguration(t, new Runnable() {
```
szczepiq: Let's use deferredConfiguration for the same reason I mentioned earlier.

epeee:
@szczepiq thx a lot for your feedback!
I already integrated your suggestions.

szczepiq:
I'll make a couple of minor tweaks and I'm merging. Nice work!

PR #96, opened 2017-04-26
#36 Automatically include contributors in pom.xml file

mstachniuk:
New Pull Request after rebase on master

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/AllContributorsFetcherTask.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/AllContributorsFetcherTask.java
@@ -18,7 +18,7 @@ public class AllContributorsFetcherTask extends DefaultTask {
     private static final Logger LOG = Logging.getLogger(AllContributorsFetcherTask.class);
 
     @Input private String repository;
-    @Input private String authToken;
+    @Input private String readOnlyAuthToken;
 
     @OutputFile private File contributorsFile;
 
@@ -26,7 +26,7 @@ public class AllContributorsFetcherTask extends DefaultTask {
     public void fetchAllProjectContributorsFromGitHub() {
         LOG.lifecycle("  Fetching all contributors for project");
 
-        GitHubContributorsProvider contributorsProvider = Contributors.getGitHubContibutorsProvider(repository, authToken);
+        GitHubContributorsProvider contributorsProvider = Contributors.getGitHubContibutorsProvider(repository, readOnlyAuthToken);
         ProjectContributorsSet allContributorsForProject = contributorsProvider.getAllContributorsForProject();
 
         AllContributorsSerializer serializer = Contributors.getAllContributorsSerializer(contributorsFile);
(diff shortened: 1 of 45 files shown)


src/main/groovy/org/mockito/release/internal/gradle/AllContributorsFetcherTask.java
```diff
@@ -0,0 +1,64 @@
...
+    @TaskAction
+    public void fetchAllProjectContributorsFromGitHub() {
+        if(skipTaskExecution) {
+            LOG.lifecycle("  Fetching all contributors for project SKIPPED (disabled in configuration)");
+            return;
+        }
+
+        // Fetching info about all contributors is expensive.
+        // Task execution is skipped when output file exist for speed up local builds.
+        if(contributorsFile.exists()) {
```
szczepiq: I'm not sure we need this. Gradle's inputs/outputs mechanism should take care of it. You have declared the contributorsFile as @OutputFile which should be enough :)

src/main/groovy/org/mockito/release/internal/gradle/AllContributorsFetcherTask.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Fetch data about all project contributors and store it in file. It is used later in generation pom.xml.
+ */
+public class AllContributorsFetcherTask extends DefaultTask {
+
+    private static final Logger LOG = Logging.getLogger(AllContributorsFetcherTask.class);
+
+    @Input private String repository;
+    @Input private String readOnlyAuthToken;
+    @Input private boolean skipTaskExecution;
```
szczepiq: I suspect we don't really need this field. You can use existing Gradle Task API to do this:

```java
//during Gradle's configuration phase:
someTask.setEnabled(false);

//or, alternatively hook up the task dependencies conditionally during configuration phase:
if (...) {
  someTask.dependsOn(otherTask);
}

//during Gradle's task execution phase
someTask.onlyIf(new Spec() { ... })
```

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
```diff
@@ -33,13 +35,15 @@
...
  *     <li>Adds build.dependsOn "publishToMavenLocal" to flesh out publication issues during the build</li>
+ *     <li>Makes pom-generating task depend on "fetchAllProjectContributorsFromGitHub" to load all contributors from GitHub</li>
  * </ul>
  */
 public class BaseJavaLibraryPlugin implements Plugin<Project> {
 
     private final static Logger LOG = Logging.getLogger(BaseJavaLibraryPlugin.class);
 
-    final static String PUBLICATION_NAME = "javaLibrary";
+    final static String PUBLICATION_NAME = "JavaLibrary";
```
szczepiq: We already capitalize the name so we can keep publication name as "javaLibrary". It is consistent with Gardle domain object convention names.

"JavaLibrary" -> "javaLibrary"

I'll make a quick fix and push it to your branch :)

Issue #84, opened 2017-04-23
Avoid publishing new version if nothing has changed


szczepiq:
## Problem

Publishing new version when it is binary equal to the previous version is confusing for the clients. At the moment we don't check if new version is different than the current release. If someone makes a change in .gitignore it can trigger build and potentially a publication.

## Suggested implementation

Only a suggested plan, please execute as you see fit!

### Robustness of the algorithm

The comparison algorithm is _not_ on the critical path and we are allowed to be pragmatic in coming up the implementation and we can iterate on the solution to make it better. Here are the effect of false positive / false negative signal from the algorithm:
 - false negative, publications incorrectly considered non-equal: extra version is released that does not have changes on top of previous version. It is minor problem for projects that leverage the "notable releases" feature. For other projects it should not be a huge deal because versions are shipped frequently in continuous delivery.
 - false positive, publications incorrectly considered equal: release is not triggered and needs to be triggered semi-manually. Not a critical problem.

### Gradle plugins / tasks

1. We have a "releaseNeeded" task that determines whether the release should be triggered. This task needs to depend on some other task(s) that pull down and compare publications.
2. We could create "PublicationsComparatorPlugin" that applies "BaseJavaLibraryPlugin" and "VersioningPlugin". New plugin would add task that that compares publications. New plugin would be applied by "JavaLibraryPlugin".
3. releaseNeeded would depend on all tasks added at step 2) from projects that use PublicationsComparatorPlugin.

### Implementation

1. It is problematic to use Gradle's configurations to download previous versions of jars / poms for comparisons because Gradle automatically replaces dependencies with local project dependency with the version that is currently being built. We could reach out to Gradle folks on the forum and ask how to do it. If we cannot get it to work, we need to hit the url directly and download file without using Gradle's dependency resolution.
2. The task should be incremental, e.g. when nothing changes and build dir is intact, we don't download jars again.
3. Current publication model is that we we publish all jars from the project during the release. We want to stick with this model of releasing because it is simple and powerful - the project is a unit of deployment/release and all jars go out with the same version. If only one jar has changed, we still publish all jars with new version.
4. Stuff to consider when comparing publications:
 - pom.xml content
 - jar content / binaries (for that, perhaps it's simpler to just compare the sources jar, it will cover compiled classes and javadocs at the same time)
 - MANIFEST.MF from jars
5. There is already some code to be reused, legacy of Mockito automation :) See "org.mockito.release.internal.comparison" package.

[...]

PR #97, opened 2017-04-26
avoid publishing if nothing changes - initial version, with comparing only poms  #84

wwilk:
@szczepiq 

Like I said on Slack this is a draft, just to show you the dependencies that I made between plugins. 
Right now it runs like that: 
:bumpVersionFile
:fetchNotableReleaseNotes
...
:gitCommit
:gitPush
...
:{subprojects}:jar
...
:{subprojects}:publishToMavenLocal
:{subprojects}:comparePublications
:assertReleaseNeeded
:{subprojects}:bintrayUpload
:bintrayUploadAll
:performRelease

If all of the publications are equal it fails on :assertReleaseNeeded

If I didn't miss anything there is no configuration or runtime dependencies between two hierarchies of plugins. BaseJavaLibraryPlugin checks if bumpVersionFile task is available before using it, so if there is no VersioningPlugin, release will be always needed because there is no way to compare publications without versions. 

ContinuousDeliveryPlugin only makes assertReleaseNeeded dependent on comparePublications if there are projects who use BaseJavaLibraryPlugin.


src/test/groovy/org/mockito/release/version/DefaultVersionInfoTest.groovy
```diff
@@ -112,7 +115,45 @@ notableVersions=1.0.0
         then:
         f.text == """version=1.0.1
 notableVersions=1.0.0
+previousVersion=1.0.0
 """
         v.notableVersions.toString() == ["1.0.0"].toString()
     }
+
+    def "bumps previous version"() {
```
szczepiq: Great unit test. Makes it really clear to understand the new behavior.

szczepiq:
Nice work! Really looking forward to enabling this logic in Mockito project.

I also very much like the previous version setting in the version file and the version info object. We will leverage it for release notes!

1. What do you think about a separate plugin “PublicationsComparatorPlugin”, that would apply “BaseJavaLibraryPlugin” and “VersioningPlugin". “JavaLibraryPlugin” would apply “PublicationsComparatorPlugin” and "“BaseJavaLibraryPlugin”. This way, the “BaseJavaLibraryPlugin” remains simple, decoupled from the publication comparison logic and reusable outside of releasing. We should start documenting the diagrams of plugins :)

2. I’m not sure why "compare publications task” depends on bumpVersion file. It would be kind of awkward because each time you compare publications, the version will be incremented. Here’s the workflow I would suggest:

a) assertReleaseNeeded depends on new task comparePublicationsAll
b) comparePublicationsAll depends on comparePublications in all submodules (similar to how we do it with bintrayUploadAll). comparePublicationsAll also keeps all tasks it depends on in a list / some field on the task type. When executed, comparePublicationsAll will iterate over the comparePublications tasks and see if all are equal
b) assertReleaseNeeded is intended to be called as a separate Gradle invocation. In travis.yml of the example project we do:

```
after_success:  - ./gradlew assertReleaseNeeded && ./gradlew travisReleasePrepare && ./gradlew performRelease
```

This is the simplest way to implement ‘release needed functionality’. If the assert task fails, other commands (travisReleasePrepare, performRelease) will not be executed at all. Note that we don’t need to run ‘bumpVersion’ task at all in above workflow. We just compare current version (you get it from version.properties) with previous version.

I'll make some tweaks to your code and push to your branch :)

src/main/groovy/org/mockito/release/gradle/ReleaseNeededTask.java
```diff
@@ -29,6 +29,7 @@
 
     private String branch;
     private String releasableBranchRegex;
+    private boolean allPublicationsEqual;
```
szczepiq: Nice new property. Modeling the API like that makes the code and API clean.

src/main/groovy/org/mockito/release/gradle/ReleaseNeededTask.java
```diff
@@ -78,7 +87,8 @@ public void setReleasableBranchRegex(String releasableBranchRegex) {
                 "\n    - skip by env variable: " + skipEnvVariable +
                 "\n    - skip by commit message: " + skippedByCommitMessage +
                 "\n    - is pull request build:  " + pullRequest +
-                "\n    - is releasable branch:  " + releasableBranch;
+                "\n    - is releasable branch:  " + releasableBranch +
+                "\n    - anything changed in publications since the last release:  " + allPublicationsEqual;
```
szczepiq: This is also nice :) Will make it really helpful when debugging stuff 🐛 :)

src/main/groovy/org/mockito/release/internal/comparison/PublicationsComparatorTask.java
```diff
@@ -26,17 +34,142 @@ public boolean isPublicationsEqual() {
...
+        FileOutputStream output = null;
+        try {
+            String url = getRemoteUrl(extension);
+            input = new BufferedInputStream(new URL(url).openStream());
+            String localUrl = getLocalArtifactUrl(getPreviousVersion(), extension);
+
+            IOUtil.createParentDirectory(new File(localUrl));
+
+
+            FileOutputStream fos = new FileOutputStream(localUrl);
```
mstachniuk: Maybe you can move lines 89-109 to `IOUtil`?

src/main/groovy/org/mockito/release/internal/comparison/PublicationsComparatorTask.java
```diff
@@ -26,17 +34,142 @@ public boolean isPublicationsEqual() {
...
+
+
+    /**
+     *
+     * @param extension, suffix of artifact eg ".pom" or "-sources.jar"
+     * @return eg
+     * https://bintray.com/shipkit/examples/download_file?file_path=/org/mockito/release-tools-example/api/0.15.1/api-0.15.1.pom";
+     */
+    private String getRemoteUrl(String extension){
+        return "https://bintray.com/shipkit/examples/download_file?file_path="
```
mstachniuk: Please don't forget that not everyone use bintray ;-)

wwilk:
@mstachniuk 
Thanks for the review! As I said above, it's a draft, that's why Bintray is hardcoded, I wanted to connect everything together first and then focus on details. I wrote about that in corresponding issue - the idea is to allow user to provide us with a closure that will return download URL based on some parameters (group, name, version etc). This way he can use whatever remote repository he wants. 

Good idea with moving download code to IOUtil, I'll do that!

szczepiq:
Pushed some more Gradle API guidance. Hope that helps!

wwilk:
@szczepiq Thanks! I committed a few things, but still not done, will be tomorrow probably.

src/main/groovy/org/mockito/release/internal/comparison/BaseProjectProperties.java
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    public String getName() {
+        return name;
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o) return true;
+        if (o == null || getClass() != o.getClass()) return false;
```
mstachniuk: Don't forget to run checkstyle before merge

src/main/groovy/org/mockito/release/internal/comparison/ZipComparator.java
```diff
@@ -1,52 +1,133 @@
...
 
-    Result compareFiles() {
-        final File file1 = this.file1.call();
-        final File file2 = this.file2.call();
-        notNull(file1, "zip/jar file to compare", file2, "zip/jar file to compare");
+    boolean compareZips(String filePath1, String filePath2) {
+        ZipFile file1 = openZipFile(filePath1);
+        ZipFile file2 = openZipFile(filePath2);
+
+        Set<String> set1 = new LinkedHashSet<String>();
```
mstachniuk: Extract method

src/main/groovy/org/mockito/release/internal/comparison/ZipComparator.java
```diff
@@ -1,52 +1,133 @@
...
+        }
+        LOG.info(filecount + " entries matched");
+        if (errcount > 0) {
+            LOG.info(errcount + " entries did not match");
+            return false;
+        }
+        return true;
+    }
+
+    private ZipFile openZipFile(String filePath) {
```
mstachniuk: Consider create a class ZipUtils and move this method there.

src/main/groovy/org/mockito/release/internal/gradle/BaseJavaLibraryPlugin.java
```diff
@@ -77,6 +86,61 @@ public void execute(Jar jar) {
...
+        TaskMaker.task(project, COMPARE_PUBLICATIONS_TASK, PublicationsComparatorTask.class, new Action<PublicationsComparatorTask>() {
+            public void execute(final PublicationsComparatorTask t) {
+                t.setDescription("Compares artifacts and poms between last version and the currently built one to see if there are any differences");
+
+                project.getRootProject().getPlugins().withType(
+                        VersioningPlugin.class,
+                        new Action<VersioningPlugin>() {
+                            @Override
+                            public void execute(VersioningPlugin versioningPlugin) {
+                                System.out.println("here");
```
mstachniuk: System.out.println()

src/test/groovy/org/mockito/release/internal/comparison/PomSiblingDependencyRemoverTest.groovy
```diff
@@ -0,0 +1,112 @@
...
+                "</project>"
+
+        when:
+        def result = new PomSiblingDependencyRemover().removeSiblingDependencies(emptyPom,
+                [new BaseProjectProperties("org.mockito", "independent")] as Set)
+
+        then:
+        result == "<project>\n  <dependencies/>\n</project>\n"
+    }
+}
```
mstachniuk: Consider add <classifier> to tests

wwilk: What do you mean? That I should cover more cases?

mstachniuk: Sorry I used `<>` in reply and GitHub cutted a part of message:-)
Some dependencies can use `<classifier>` tag. It can be important for comparator.

wwilk: Yeah good edge case! I'll add it to a unit test, new version of implementation will handle this situation.

src/main/groovy/org/mockito/release/internal/comparison/PomFilter.groovy
```diff
@@ -0,0 +1,47 @@
+package org.mockito.release.internal.comparison
+
+class PomFilter {
```
szczepiq: This class is fantastic! I really like the API and the model, and neat unit test. Thanks!

src/test/groovy/org/mockito/release/gradle/ReleaseNeededTaskTest.groovy
```diff
@@ -0,0 +1,75 @@
+package org.mockito.release.gradle
+
+import org.gradle.api.GradleException
+import org.gradle.testfixtures.ProjectBuilder
+import org.mockito.release.internal.comparison.PublicationsComparator
+import spock.lang.Specification
+
+class ReleaseNeededTaskTest extends Specification {
```
szczepiq: This is a great test. Thank you very much for adding it!!!!!!!

We were long overdue for the unit test for this task.

szczepiq:
Merging, lets work on outstanding TODOs in separate RBs. For now, I have unhooked the new tasks from the main workflow. So now the code is testable and integrated, but not integrated in workflow.

Looking forward to using this feature in Mockito! ;)

Issue #101, opened 2017-04-29
Simplify configuration - get rid of "releasing.build.branch"


szczepiq:
### Problem

We should be able to to simplify release configuration and get rid of "releasing.build.branch" completely:

```Gradle
releasing {
  //we should be able to get rid of the below and configure automatically:
  build.branch = System.getenv("TRAVIS_BRANCH")
}
```

It is super annoying that I have to configure TRAVIS_BRANCH env variable even for local testing or releasing from local (I'm not Travis!). Fixing it will make the tools easier to use today, and easier to integrate with other tools in the future.

### Workflow

1. travisReleasePrepare task ensures that TRAVIS_BRANCH is checked out.
2. in other workflows (non-travis), for example local testing, deployment from local, we don’t read nor require the TRAVIS_BRANCH property at all.

### Suggested implementation

See the suggested starting point design at: https://github.com/mockito/mockito-release-tools/pull/102

szczepiq:
@epeee, this is related to your previous neat work, making the release configuration even simpler. Do you want to take a stab at it? :)

epeee:
@szczepiq yes, I will do so tomorrow.

PR #106, opened 2017-05-02
#101 Simplify configuration

epeee:
a very first version, I will try to add some tests soon.

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
@@ -206,7 +206,7 @@ public class ContinuousDeliveryPlugin implements Plugin<Project> {
 
                 lazyConfiguration(t, new Runnable() {
                     public void run() {
-                        String branch = project.getExtensions().getByType(GitPlugin.GitStatus.class).getBranch();
+                        String branch = project.getExtensions().getByType(GitStatusPlugin.GitStatus.class).getBranch();
                         t.setBranch(branch);
                     }
                 });
--- a/src/main/groovy/org/mockito/release/internal/gradle/GitStatusPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/GitStatusPlugin.java
@@ -1,0 +1,53 @@
+package org.mockito.release.internal.gradle;
+
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.mockito.release.exec.Exec;
+import org.mockito.release.exec.ProcessRunner;
+
+public class GitStatusPlugin implements Plugin<Project> {
+
+    private GitStatus gitStatus;
+
+    @Override
+    public void apply(Project project) {
+        if (project.getParent() == null) {
+            gitStatus = project.getExtensions().create(GitPlugin.class.getName(), GitStatus.class, project);
+        } else {
+            gitStatus = project.getRootProject().getExtensions().create(GitPlugin.class.getName(), GitStatus.class, project);
+        }
+    }
+
+    public GitStatus getGitStatus() {
+        return gitStatus;
+    }
+
+    public static class GitStatus {
+
+        private final ProcessRunner processRunner;
+        private volatile String branchName;
+        private static final Object SYNC = new Object();
+
+        public GitStatus(Project project) {
+             processRunner = Exec.getProcessRunner(project.getRootDir());
+        }
+
+        public void setBranchName(String branchName) {
+            synchronized (SYNC) {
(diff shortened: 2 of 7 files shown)


src/main/groovy/org/mockito/release/internal/gradle/GitStatusPlugin.java
```diff
@@ -0,0 +1,53 @@
...
+
+        private final ProcessRunner processRunner;
+        private volatile String branchName;
+        private static final Object SYNC = new Object();
+
+        public GitStatus(Project project) {
+             processRunner = Exec.getProcessRunner(project.getRootDir());
+        }
+
+        public void setBranchName(String branchName) {
```
szczepiq: Let's make it even simpler :)

1. I suggest we get rid of the 'setBranchName' method from here
2. We can create a new task type 'GitCheckOutTask' task that has 'setRev' method which is the target revision we want to check out.
3. New task type 'GitCheckOutTask' will be used for creating 'checkOutBranch' task  in GitPlugin
4. TravisPlugin already applies GitPlugin so travis plugin can configure checkOutBranch.setRev with the value acquired from env variable

This way, we don't offer way to set branch and things are simpler.

src/main/groovy/org/mockito/release/internal/gradle/TravisPlugin.java
```diff
@@ -35,7 +37,7 @@ public void apply(Project project) {
         boolean isPullRequest = pr != null && !pr.trim().isEmpty() && !pr.equals("false");
 
         conf.getBuild().setCommitMessage(System.getenv("TRAVIS_COMMIT_MESSAGE"));
-        conf.getBuild().setBranch(System.getenv("TRAVIS_BRANCH"));
+        gitStatus.setBranchName(System.getenv("TRAVIS_BRANCH"));
```
szczepiq: This is a little bit risky. If some other plugin requests the branch during Gradle's configuration time, and before travis plugin is applied then some objects might get wrong value of the branch. I suggest to apply my earlier suggestion and get rid of setBranchName method.

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -206,7 +206,8 @@ public void execute(final ReleaseNeededTask t) {
 
                 lazyConfiguration(t, new Runnable() {
                     public void run() {
-                        t.setBranch(conf.getBuild().getBranch());
+                        String branch = project.getExtensions().getByType(GitStatusPlugin.GitStatus.class).getBranch();
```
szczepiq: Nice!

src/main/groovy/org/mockito/release/internal/gradle/GitStatusPlugin.java
```diff
@@ -0,0 +1,53 @@
+package org.mockito.release.internal.gradle;
+
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.mockito.release.exec.Exec;
+import org.mockito.release.exec.ProcessRunner;
+
+public class GitStatusPlugin implements Plugin<Project> {
```
szczepiq: Great new plugin! Javadocs please :)

src/main/groovy/org/mockito/release/internal/gradle/TravisPlugin.java
epeee: k, thx for pointing that out. I will apply your suggestion later on today.

epeee:
@szczepiq ad 3, joining the team, yes, would be great and would make collaboration easier, thx.

src/main/groovy/org/mockito/release/internal/gradle/TravisPlugin.java
szczepiq: Awesome!

szczepiq:
>@szczepiq ad 3, joining the team, yes, would be great and would make collaboration easier, thx.

Done. Welcome!

szczepiq:
@epeee, please create a dev branch in release-tools-project that you will work on. It will make it easier for us to review and make changes directly on your branch.

src/main/groovy/org/mockito/release/internal/gradle/GitStatusPlugin.java
epeee: Thx, your suggestion makes it way cleaner. :+1:

epeee:
@szczepiq I will create a new PR using my dev branch from release-tools-project.

mstachniuk:
Closed because of new PR: https://github.com/mockito/mockito-release-tools/pull/109

PR #109, opened 2017-05-04
 Simplify configuration - get rid of "releasing.build.branch" (#101) (#106)

### LINKED ISSUE #101: Simplify configuration - get rid of "releasing.build.branch"

epeee:
- removes "releasing.build.branch" property from configuration object
- changes "travisReleasePrepare" to "ciReleasePrepare"

Fixes #101


epeee:
@szczepiq thx!

ad pushing directly to my branch: sure, that's ok.

Issue #87, opened 2017-04-23
Release tools project dogfoods versioning plugin


szczepiq:
### Problem

Release tools project currently defines its version in root "build.gradle" file and it is problematic because:
 - it is not bumped automatically with the release
 - it triggers merge conflicts in "build.gradle" file instead of merge conflicts in a simpler 'version.properties' file.

### Suggested implementation

1. Have the release tools project consume the 'org.mockito.release-tools.versioning' plugin from the previously released version of tools.
2. Update the travisRelease task so that when we release we bump version automatically.
3. Bonus - we should configure versioning plugin in 'pluginBundle' section of 'build.gradle' file. See the documentation of Gradle plugins portal: https://plugins.gradle.org/docs/publish-plugin.

### Future

We should make the release tools plugin actually consume the entire continuous delivery plugin. We can get there step by step.

szczepiq:
@wwilk, do you want to take a stab at it? It's related to the version fix you've done earlier and might be good to get you immersed in the project more!

wwilk:
Sure! I will work on it soon.

szczepiq:
Fiendly ping :) It bothers me that we always have to remember to bump the version of tools when publishing :/

szczepiq:
I'm thinking that we should make the plunge and start dogfooding release notes generation in this project, too. Let's rock! What do you think? Or perhaps new plugin like "GradlePluginReleasingPlugin"? :D

wwilk:
Sure, we can do that!

PR #110, opened 2017-05-04
AutoVersioningPlugin introduced #87

wwilk:
@szczepiq 

What do you think about this? The idea was to keep VersioningPlugin as simple as possible, since it's needed for releasing.previousReleaseVersion, so I added AutoVersioningPlugin which handles commiting and pushing to GitHub. 

I created a new project to test only auto-versioning. 
https://github.com/wwilk/shipkit-versioning-example
You can see exemplary bumping version commits there.

It's not fully configured yet, since I want to first release auto-versioning plugin and then use it here. 
In travis-release task I would just call ./gradlew publishPlugins bumpVersionAndPush.


src/test/groovy/org/mockito/release/internal/gradle/AutoVersioningPluginTest.groovy
```diff
@@ -0,0 +1,15 @@
+package org.mockito.release.internal.gradle
+
+import org.gradle.testfixtures.ProjectBuilder
+import spock.lang.Specification
+
+class AutoVersioningPluginTest extends Specification {
```
szczepiq: Thank you for adding a lifecycle test. I often forget to do so for the plugins I create!

src/test/groovy/org/mockito/release/internal/gradle/util/GitUtilTest.groovy
```diff
@@ -18,4 +19,57 @@ class GitUtilTest extends Specification {
         ""    | "[ci skip]" | " [ci skip]"    //info will never be empty, only documenting behavior
         "foo" | ""          | "foo"
     }
+
+    def "git push" () {
```
szczepiq: GREAT!!! Thank you very much for adding the missing unit tests!

version.properties
```diff
@@ -0,0 +1,6 @@
+#Version of the produced binaries. This file is intended to be checked-in.
+#It will be automatically bumped by release automation.
+version=0.8.6
```
szczepiq: Ha! Finally :)

szczepiq:
@wwilk, FYI I made few changes on your branch. The changes are small enough that I will allow myself to merge the PR. Thanks again for the codes! :D

PR #111, opened 2017-05-04
PublicationsComparator interface removed #84

wwilk:
@szczepiq 
As you wrote in the javadoc this interface wasn't needed, so I removed it and fixed tests accordingly. I would prefer to use mockingg instead of a setter, but it seems that it's not possible with Spock Mocking API,  all methods call default constructor which is not allowed for gradle tasks.


szczepiq:
Merge at will!

PR #118, opened 2017-05-06
Rework and simplification around fetching contributors GitHub

szczepiq:
It's best to review commit-by-commit, some TODOs are still pending!!!

### DIFF
--- a/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
@@ -32,7 +32,6 @@ public class ReleaseConfiguration {
         //Configure default values
         git.setTagPrefix("v"); //so that tags are "v1.0", "v2.3.4"
         git.setReleasableBranchRegex("master|release/.+");  // matches 'master', 'release/2.x', 'release/3.x', etc.
-        team.setAddContributorsToPomFromGitHub(true);
     }
 
     //TODO currently it's not clear when to use class fields and when to use the 'configuration' map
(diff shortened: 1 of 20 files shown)


src/main/groovy/org/mockito/release/internal/gradle/util/PomCustomizer.groovy
```diff
@@ -80,14 +59,20 @@ class PomCustomizer {
...
+            if (!conf.team.contributors.isEmpty()) {
+                def devs = new HashSet(conf.team.developers)
+                def node = root.appendNode('contributors')
+                conf.team.contributors.each {
+                    if (!devs.contains(it)) {
+                        def split = it.split(':')
+                        assert split.length == 2
+                        def d = node.appendNode('contributor')
+                        d.appendNode('name', split[1])
+                        d.appendNode('roles').appendNode('role', 'Core developer')
```
szczepiq: Ha! Great catch, thank you.

src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
```diff
@@ -387,23 +386,6 @@ public void setDevelopers(Collection<String> developers) {
...
             configuration.put("team.contributors", contributors);
         }
-
-        /**
-         * A boolean flag for fetch all contributors from GitHub to include them in generated pom file.
-         * This is optional value, by default set to true.
-         * <p>
-         * See POM reference for <a href="https://maven.apache.org/pom.html#Contributors">Contributors</a>.
-         */
-        public boolean isAddContributorsToPomFromGitHub() {
```
szczepiq: Yes, the use case is important and we satisfy it: when releasing.team.contributors has any elements, we don't fetch contributors from github and don't populate releasing.team.contributors. Thoughts?

mstachniuk: Ah, ok. We need to document this behaviour.

PR #119, opened 2017-05-07
Refactorings, tidy-up and finally getting contributors links in new release notes!!!

szczepiq:
Best to review commit-by-commit


src/main/groovy/org/mockito/release/internal/gradle/RecentContributorsFetcherTask.java
```diff
@@ -53,73 +54,105 @@ public void setReleaseNotesData(File releaseNotesData) {
...
+     * See {@link #getOutputFile()}
+     */
+    public void setOutputFile(File outputFile) {
+        this.outputFile = outputFile;
     }
 
-    public String getToRevision() {
-        return toRevision;
+    @TaskAction
+    public void fetch() {
```
mstachniuk: Consider move this method up, before getters and setters

src/main/groovy/org/mockito/release/notes/format/DetailedFormatter.java
```diff
@@ -141,7 +141,7 @@ private static String itemizedAuthors(ContributionSet contributions) {
         StringBuilder sb = new StringBuilder();
         boolean showIndividualCommits = contributions.getAuthorCount() > 1;
         for (Contribution c : contributions.getContributions()) {
-            sb.append(c.getAuthorName());
+            sb.append(authorLink(c));
```
mstachniuk: 👍 :clap:

src/main/groovy/org/mockito/release/internal/gradle/RecentContributorsFetcherTask.java
szczepiq: I moved the setters up for consistency with other code. I think you are right that it is more important than the setters so it should probably by up :) For now I'll leave things consistent. Your feedback was noted!

szczepiq:
Merging with pleasure ;) Thanks @mstachniuk on all the GitHub API nitty gritty that enabled this feature to work!!!

PR #121, opened 2017-05-07
Introducing PublicationsComparatorPlugin

wwilk:
Added documentation to PublicationsComparatorTask. Extracted this task into PublicationsComparatorPlugin and configured smart defaults for artifact URLs. For now only for BintrayPlugin.


szczepiq:
Merging this feature! This is fantastic, thank you! Comparing publications and avoiding unnecessary releases is one of the core features of our library. It's hard to implement manually by users and it drives full automation of continuous delivery.

I pushed directly to the branch some TODOs around:
 - making the tasks incremental
 - simplifying the ComparePublications task API

Before we hook this comparison by default, I think we should improve the output from comparePublications / releaseNeeded tasks. Currently, it's not clear for me why the release is needed / not needed. Also, we should identical poms by default which really increases the noise.

Since the feature is not hooked up automatically to our workflow, I'm very pleased to merge it right now and give us some more time to work on the details :)

Issue #125, opened 2017-05-07
ReleaseConfiguration automatically loaded from conventional .gradle file


wwilk:
Currently there are some moments at the configuration part of the build, when ReleaseConfiguration properties values are default or null, even though user overrides them in one of the .gradle files. It depends on when exactly this file is applied. It shouldn't be like that - we should be always sure that at each moment only the value provided by the user is used. 

As discussed on Slack with @szczepiq and @mstachniuk it would be best to have a conventional file of the specified name and location (for example /gradle/shipkit.gradle) which would be loaded by mockito-release-tools automatically when ReleaseConfigurationPlugin is applied (but after application of default values to allow user to override them).

Additionally when user doesn't have the conventional file in their project, we should create this file, similarly to what we do with version.properties file if it doesn't exist.

wwilk:
Done already :)

PR #136, opened 2017-05-09
ReleaseNeededPlugin introduced #84

wwilk:
Extracted releaseNeeded tasks into a separate new plugin - ReleaseNeededPlugin.

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/ReleaseNeededPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/ReleaseNeededPlugin.java
@@ -1,0 +1,83 @@
+package org.mockito.release.internal.gradle;
+
+import org.gradle.api.Action;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.mockito.release.gradle.ReleaseConfiguration;
+import org.mockito.release.gradle.ReleaseNeededTask;
+import org.mockito.release.internal.comparison.PublicationsComparatorTask;
+import org.mockito.release.internal.gradle.util.TaskMaker;
+
+import static org.mockito.release.internal.gradle.configuration.LazyConfiguration.lazyConfiguration;
+
+/**
+ * Adds tasks for checking if release is needed
+ *
+ * Applies following plugins and preconfigures tasks provided by those plugins:
+ *
+ * <ul>
+ *     <li>{@link ReleaseConfigurationPlugin}</li>
+ *     <li>{@link GitStatusPlugin}</li>
+ * </ul>
+ *
+ * Adds following tasks:
+ *
+ * <ul>
+ *     <li>assertReleaseNeeded</li>
+ *     <li>releaseNeeded</li>
+ * </ul>
+ */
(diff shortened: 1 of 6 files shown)


src/main/groovy/org/mockito/release/internal/gradle/ReleaseNeededPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+        releaseNeededTask(project, "releaseNeeded", conf)
+                .setExplosive(false)
+                .setDescription("Checks and prints to the console if criteria for the release are met.");
+    }
+
+    private static ReleaseNeededTask releaseNeededTask(final Project project, String taskName, final ReleaseConfiguration conf) {
+        return TaskMaker.task(project, taskName, ReleaseNeededTask.class, new Action<ReleaseNeededTask>() {
+            public void execute(final ReleaseNeededTask t) {
+                t.setDescription("Asserts that criteria for the release are met and throws exception if release not needed.");
+                t.setExplosive(true);
```
mstachniuk: This two properties are configured above in lines 42-42 and 48-49 so you can remove them.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseNeededPlugin.java
```diff
@@ -0,0 +1,83 @@
...
+                                // make this task depend on all comparePublications tasks
+                                Task task = subproject.getTasks().getByName(PublicationsComparatorPlugin.COMPARE_PUBLICATIONS_TASK);
+                                t.addPublicationsComparator((PublicationsComparatorTask) task);
+                            }
+                        });
+                    }
+                });
+
+                final GitStatusPlugin.GitStatus gitStatus = project.getPlugins().apply(GitStatusPlugin.class).getGitStatus();
+                lazyConfiguration(t, new Runnable() {
```
mstachniuk: `lazyConfiguration` should be used only for setting sensitive information like password tokens etc. Can you use `DeferredConfiguration.deferredConfiguration` here instead?

wwilk: When we implement issue #125 we won't need deferredConfiguration any more and we can move this configuration out of lazyConfig block.

szczepiq: Nice catch guys! In this case, we want to use lazyConfiguration because getting branch is expensive - we need to execute external process.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseNeededPlugin.java
szczepiq: Good catch! I think that @wwilk attempts to work around the validation imposed by TaskMaker.task - it validates if description is provided. @wwilk, you can pass description as parameter to releaseNeededTask and configure 'explosive' only outside of 'releaseNeededTask' method.

Hope that helps ;)

wwilk: @szczepiq I only moved this code :D But I will do these changes anyway ;)

PR #137, opened 2017-05-09
ReleaseNeededPlugin applied #84

wwilk:


build.gradle
```diff
@@ -79,29 +80,13 @@ pluginBundle {
...
-                    "  Release to Gradle Plugin Portal is not needed.", path)
-        }
-    }
-}
-
-//TODO ww add -Preleasing.dryRun=false to commandLine when releaseNeeded takes into account [ci skipRelease] in commit msg
 task travisRelease {
     dependsOn releaseNeeded
-    onlyIf { releaseNeeded.needed }
+    onlyIf { !releaseNeeded.releaseNotNeeded }
```
szczepiq: I suggest we use 'assertReleaseNeeded' in Travis CI instead of onlyIf (similar to how it is set up for mockito and example project). It makes things a bit simpler complexity wise, it makes things slightly easier to deploy/test locally, if you need to test/publish from local.

build.gradle
```diff
@@ -79,29 +80,13 @@ pluginBundle {
...
     doLast {
         logger.lifecycle("{} - Publishing to Gradle Plugin Portal...", path)
         exec {
-            commandLine "./gradlew", "publishPlugins", "bumpVersionAndPush",
+            commandLine "./gradlew", "publishPlugins", "bumpVersionAndPush", "-Preleasing.dryRun=false",
```
szczepiq: Excellent!

Now let's try to figure out how to:
1. Hide all that in gradle-plugin-releasing (and also wrap release-needed into that plugin)
2. Avoid forking off Gradle build from inside Gradle build - this makes the build output really hard to grok.

PR #140, opened 2017-05-12
AllContributorsSerializer changed to get and return JSON String

### LINKED ISSUE #91: Use JSON serialization/deserialization in ReleaseNotesSerializer
We want to use JSON format instead of Java serialization one for storing release notes. It would be probably best to use json-simple for this, since it's already a dependency.

NagRock:
Please give me feedback if this is correct way. I can continue with this to close #91 but as I wrote in this issue, I think that serializers should only do one job, serialize to JSON without writing data to file. This can give us more flexibility to reuse serializers in other serializers.
Cheers!


szczepiq:
Looks really good! I don't have any meaningful feedback regarding the implementation other than it's good stuff :) Thank you!

PR #141, opened 2017-05-13
Simplified how release notes consume contributors

### LINKED ISSUE #120: Simplify getting contributors information
### Problem

Now we have 2 ways of getting contributors and it adds some complexity to the code. During the build we use both ways of getting contributors information, wrapped in separate Gradle tasks. Let's have one robust service for getting contributors and one Gradle task

### Workflow

1. First, we hit repo/contributors endpoint and get all of the guys (https://developer.github.com/v3/repos/#list-contributors).
 - Let’s use max page size :D
2. Since we may not get the most recent contributors due to GitHub performance limitations, let’s do another hit for commits (https://developer.github.com/v3/repos/commits/#list-commits-on-a-repository).
 - Let’s limit the commits to last 24hrs, earlier commits should be covered by repo/contributors request
3. We combine the results

### Implementation

1. Update [GitHubAllContributorsFetcher](https://github.com/mockito/mockito-release-tools/blob/929e3e73b2192fbac87a1a2bf27570b023eaa4c3/src/main/groovy/org/mockito/release/notes/contributors/GitHubAllContributorsFetcher.java#L20). @mstachniuk, can you take a stab at it at some point? :)
2. Rework Gradle tasks, eliminate dead code - I'll do it with pleasure! (@szczepiq)

szczepiq:
The formatting of release notes is getting out of hand.
I have tried to make it a bit simpler but it's not super clean.

- removed bunch of classes that we don't need
- avoided weird logic that mutated release notes data model (contributors)

Fixes #120


mstachniuk:
👍

Issue #143, opened 2017-05-14
Release needed is always false because we cannot correctly identify branch


szczepiq:
Mockito build cannot publish at the moment because release needed is always false. From the [build log](https://travis-ci.org/mockito/mockito-release-tools-example/builds/229781873#L333):

```
  Current branch 'HEAD' matches 'master|release/.+': false
:assertReleaseNeeded FAILED
```

Note that the task has identified the current branch as 'HEAD'. It should identify it as "master" or something like that.

I'm working on a resolution right now.

PR #144, opened 2017-05-14
Fixed the problem with incorrect handling of branch

szczepiq:
Useful to review commit-by-commit

- Fixed the symptom of the problem is described in #143
- Refactorings and tidy-ups

Solution to #143:
- Made the TravisPlugin configure ReleaseNeededTask explicitly.
 We no longer need to use the release configuration object to pass state between tasks.
 That was really awkward.
- Simplified the configuration. The "build" part of the configuration did not make much sense.
It is not a setting the user would configure, it was for internal use of our tasks


szczepiq:
Thank you for review!

Issue #147, opened 2017-05-15
Some contributor links are missing in Mockito release notes


szczepiq:
Some contributor links are missing in recent [Mockito release notes](https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md). Copy paste from notes, Christian Schwarz and Serge Bishyr link is missing:

**2.8.29** - [7 commits](https://github.com/mockito/mockito/compare/v2.8.28...v2.8.29) by [Szczepan Faber](http://github.com/szczepiq) (5), Christian Schwarz (1), Serge Bishyr (1) - *2017-05-15* - published to [maven/mockito-development](https://bintray.com/mockito/maven/mockito-development)

Most likely, some of my recent refactorings broke it :)

szczepiq:
I was thinking about options to solve the contributors problem. I think it's very useful to get it right because it will pave the way how we will solve similar changes when we grow the plugins. Here are our options (there could be even more, but I think the list is pretty much complete). The options are written using Gradle jargon and with my own mental shortcuts - pardon me if this is hard to grok.

### Options

1. **File based**: Task that needs contributors has an optional File input property, and depends on fetch all contributors task. During execution they would deserialize all contributors unless the file is not provided. This is the approach we did before we introduced "configureContributors" task.

* simple, intuitive
* duplicated deserialization efforts if there are many tasks that need contributors, pushes us to make contributors serialization/deserialization utilities public otherwise the tasks are not easily reusable outside of our plugins (or we would have to make the JSON format public.

2. **Task based**: Task that needs contributors depends on “configureContributors” task and gets configured from 'team.contributors'. The approach is relatively simple, single deserialization per build. This is more less the approach we have right now (with bugs :).

* 2a) use “doFirst" to configure contributors from releasing.team.contributors (doFirst to make sure we configure the contributors after ‘configureContributors’ have executed). Using doFirst to configure self makes the task hard to reconfigure by the user, it is also counter intuitive and easy to forget.
* 2b) “configureContributors” task can explicitly configure contributors on the specific tasks, instead of updating releasing.team.contributors. Makes configureContributors coupled with every task that needs contributors or we would have to expose a new interface called RequiresContributors (things are getting complicated)

3. **Shared state**: Task that needs contributors shares state with other tasks that need or provide contributors. This is the approach suggested by @mstachniuk in the pull request. Very simple approach but tasks are harder to reuse, cannot tweak one task without changing other task.

[...]

PR #150, opened 2017-05-15
#147 Some contributor links are missing in Mockito release notes

mstachniuk:
ReleaseConfiguration().getTeam().getContributors() is used in different
places in code to define task's configuration. It needs to be always
the same reference during execution.

### DIFF
--- a/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
@@ -2,10 +2,7 @@ package org.mockito.release.gradle;
 
 import org.gradle.api.GradleException;
 
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import java.util.*;
 
 import static org.mockito.release.internal.gradle.util.team.TeamParser.validateTeamMembers;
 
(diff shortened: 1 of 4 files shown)


szczepiq:
Thank you for looking into this issue and identifying the root cause!

Let me think about it a bit more before we merge it. I hope that there is a nicer approach than depending on reusing the same instance of a collection in different tasks. Although your idea does solve the problem :)

mstachniuk:
The problem was that `ReleaseNotesPlugin` set always empty contributors list to `IncrementalReleaseNotes`:
https://github.com/mockito/mockito-release-tools/blob/2c567535a8a123cd9b2b86465e6b7bd47a9fe62c/src/main/groovy/org/mockito/release/internal/gradle/ReleaseNotesPlugin.java#L112

szczepiq:
I needed to clear out my thinking about this configuration problem and wrote an essay on the topic here: https://github.com/mockito/mockito-release-tools/issues/147#issuecomment-301791747 Hopefully it is useful :)

I'd say we avoid sharing state between tasks as suggested in this PR. It may be surprising to the user, it makes it a bit harder to reconfigure tasks (you tweak one task but in fact, all other tasks that share the state are tweaked).

mstachniuk:
Close and NOT merge this PR because of new solution in https://github.com/mockito/mockito-release-tools/pull/167 suggested by @szczepiq in https://github.com/mockito/mockito-release-tools/issues/147

PR #151, opened 2017-05-16
Splitting comparePublications into two tasks 

wwilk:
Tasks - downloadPreviousReleaseArtifacts and comparePublications. Additionally downloadPreviousReleaseArtifacts made incremental. Tested on example project by running "./gradlew releaseNeeded" and checking if it downloads artifacts only once.

In the next days I will work on improving logging and splitting comparePublications further, as @szczepiq suggested in TODO comment, but this PR doesn't break any functionality so it can be merged now if everything is OK.


src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
```diff
@@ -75,9 +79,57 @@ class PublicationsComparatorPluginTest extends Specification {
...
+        given:
+        def parent = new ProjectBuilder().withProjectDir(tmp.root).withName("parent").build()
+        def child = new ProjectBuilder().withName("child").withParent(parent).build()
+
+        when:
+        child.plugins.apply("org.mockito.mockito-release-tools.publications-comparator")
+        child.evaluate()
+
+        then:
+        def task = (DownloadPreviousReleaseArtifactsTask) child.getTasks()
```
mstachniuk: Do you need casting here?

szczepiq: You are right that using cast here is technically not necessary (Groovy is dynamic!).

I think @wwilk used it to get better IDEA support, easier to work with the code. Otherwise you don't get IDE suggestions and searches are less smart. 

I'd say it is a good idea to use casting in those situations.

src/main/groovy/org/mockito/release/internal/comparison/DownloadPreviousReleaseArtifactsTask.java
```diff
@@ -0,0 +1,82 @@
...
+    private String previousVersionSourcesJarUrl;
+
+    @OutputFile
+    private File previousVersionPomLocalFile;
+    @OutputFile
+    private File previousVersionSourcesJarLocalFile;
+
+
+    @TaskAction public void downloadPreviousReleases(){
+        if(previousVersionPomUrl == null || previousVersionSourcesJarUrl == null){
```
szczepiq: This is automatically checked by Gradle because you are using @Input annotation. You can just remove the entire 'if' block.

src/main/groovy/org/mockito/release/internal/comparison/DownloadPreviousReleaseArtifactsTask.java
```diff
@@ -0,0 +1,82 @@
...
+                    + "If you use one of the supported publishing plugins default url will be configured for you.\n"
+                    + "Currently supported plugins: Bintray"
+            );
+        }
+        downloadRemoteFile(".pom", previousVersionPomUrl, previousVersionPomLocalFile);
+        downloadRemoteFile("-sources.jar", previousVersionSourcesJarUrl, previousVersionSourcesJarLocalFile);
+    }
+
+    private void downloadRemoteFile(String extension, String remoteUrl, File localFile) {
+        LOG.lifecycle("Downloading remote artifact\n" +
```
szczepiq: I like this message. When we download stuff or start potentially long operations it is always useful to give a heads-up to the user.

src/main/groovy/org/mockito/release/internal/comparison/DownloadPreviousReleaseArtifactsTask.java
```diff
@@ -0,0 +1,82 @@
...
+
+
+        IOUtil.downloadToFile(remoteUrl, localFile);
+    }
+
+    public String getPreviousVersionPomUrl() {
+        return previousVersionPomUrl;
+    }
+
+    public void setPreviousVersionPomUrl(String previousVersionPomUrl) {
```
szczepiq: +1 :)

src/main/groovy/org/mockito/release/internal/comparison/PublicationsComparatorTask.java
```diff
@@ -157,32 +104,24 @@ public void setPreviousVersion(String previousVersion) {
...
 
-    public void setDefaultArtifactUrlResolver(DefaultArtifactUrlResolver defaultArtifactUrlResolver) {
-        this.defaultArtifactUrlResolver = defaultArtifactUrlResolver;
+    public void setPreviousVersionSourcesJarFile(File previousVersionSourcesJarFile) {
+        this.previousVersionSourcesJarFile = previousVersionSourcesJarFile;
     }
 
     /**
-     * only use for test purposes!
+     * only use for testing purposes!
```
szczepiq: Curious, do you still need this?

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
```diff
@@ -75,23 +118,26 @@ public void execute(final PublicationsComparatorTask t) {
...
+
+    private String getBasePreviousVersionArtifactPath(Project project, ReleaseConfiguration conf, Jar sourcesJar) {
+        return project.getBuildDir().getAbsolutePath() + PREVIOUS_RELEASE_ARTIFACTS_DIR
+                + File.separator + sourcesJar.getBaseName() + "-" + conf.getPreviousReleaseVersion();
+    }
+
+    private String getDefaultIfNull(String url, String variableName, String extension, DefaultArtifactUrlResolver defaultArtifactUrlResolver) {
+        if(url == null){
+            if(defaultArtifactUrlResolver == null){
+                return null;
```
szczepiq: Curious, in what circumstances the resolver is null?

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
```diff
@@ -75,23 +118,26 @@ public void execute(final PublicationsComparatorTask t) {
...
+            }
+            String defaultUrl = defaultArtifactUrlResolver.getDefaultUrl(extension);
+            LOG.lifecycle("Variable {} not set. Setting it to default value - {}", variableName, defaultUrl);
```
szczepiq: Is this 'lifecycle' message useful? From looking at it, it seems like a 'info' level message. Also users will not know what the 'variable' is (they would wonder if this is an env variable? is it a property on a task?).

src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
```diff
@@ -20,9 +21,11 @@ class PublicationsComparatorPluginTest extends Specification {
 
     def "configures comparePublications task correctly"() {
         given:
-        def parent = new ProjectBuilder().withName("parent").build()
+        def parent = new ProjectBuilder().withProjectDir(tmp.root).withName("parent").build()
         def child = new ProjectBuilder().withName("child").withParent(parent).build()
 
+        new File(tmp.root.absolutePath + "/version.properties") << "version=1.0.1\npreviousVersion=1.0.0"
+
         parent.allprojects.each { p ->
```
szczepiq: You can also write it in a more compact/simpler way: ```parent.allprojects { group = "org.group" }```

src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
wwilk: Exactly that was the reason :)

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
wwilk: Currently resolver is null when you don't have a BintrayPlugin applied.

szczepiq:
@wwilk, this PR is approved so consider code review feedback (it's your discretion what suggestions you apply and what you skip) and merge at will!

src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
mstachniuk: Then maybe it's better to do that way: `DownloadPreviousReleaseArtifactsTask  task = child....`?

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
mstachniuk: Then wire in comment.

src/main/groovy/org/mockito/release/internal/comparison/DownloadPreviousReleaseArtifactsTask.java
```diff
@@ -0,0 +1,82 @@
...
+        downloadRemoteFile(".pom", previousVersionPomUrl, previousVersionPomLocalFile);
+        downloadRemoteFile("-sources.jar", previousVersionSourcesJarUrl, previousVersionSourcesJarLocalFile);
+    }
+
+    private void downloadRemoteFile(String extension, String remoteUrl, File localFile) {
+        LOG.lifecycle("Downloading remote artifact\n" +
+                "  - from {}\n" +
+                "  - and saving it to {}", remoteUrl, localFile);
+
+
```
wwilk: 👌

src/main/groovy/org/mockito/release/internal/comparison/PublicationsComparatorTask.java
wwilk: Yes, it is used in ReleaseNeededTest

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
wwilk: Yeah good idea:)

src/main/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPlugin.java
wwilk: Yes, it's a task property, I will improve the message

src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
wwilk: Yeah, it may be more readable in this case ;)

src/test/groovy/org/mockito/release/internal/gradle/PublicationsComparatorPluginTest.groovy
wwilk: Thanks!

wwilk:
Closed because of the incorrect user name in commit. I will create another one and merge it since most of the suggestions were applied

PR #152, opened 2017-05-16
Use JSON serialization/deserialization in ReleaseNotesSerializer #91

NagRock:
Please take a special look at replacement of `Predicate` into `IgnoredCommit`.


src/main/groovy/org/mockito/release/notes/vcs/IgnoredCommit.java
```diff
@@ -0,0 +1,11 @@
+package org.mockito.release.notes.vcs;
+
+import org.json.simple.Jsonable;
+import org.mockito.release.notes.model.Commit;
+
+import java.io.Serializable;
+
+public interface IgnoredCommit extends Jsonable, Serializable {
```
szczepiq: Wow, that's pretty cool, I didn't know it's possible to cleanly serialize ignored commit predicate.

I suggest to make the code super simple and remove the need to serialize the predicate. You can [test for predicate here](https://github.com/mockito/mockito-release-tools/blob/b045b8fad5e1e9deceddc8de31c54d334a34ae65/src/main/groovy/org/mockito/release/notes/vcs/GitContributionsProvider.java#L44) and avoid keeping it as state in DefaultContributionSet.

src/main/groovy/org/mockito/release/notes/model/Commit.java
```diff
@@ -1,12 +1,14 @@
...
+import org.json.simple.Jsonable;
+
 import java.io.Serializable;
 import java.util.Collection;
 
 /**
  * VCS change, a commit
  */
-public interface Commit extends Serializable{
+public interface Commit extends Jsonable, Serializable {
```
NagRock: Yes, I tried to remove extending Jsonable from interface, but our implementation classes uses interfaces instead of other classes implementation.

src/test/groovy/org/mockito/release/notes/vcs/GitCommitSerializerTest.groovy
```diff
@@ -0,0 +1,31 @@
+package org.mockito.release.notes.vcs
+
+import org.apache.commons.lang.builder.EqualsBuilder
```
NagRock: I get it from other tests, and sorry but don't know how to check it. We got dependency `testCompile "commons-lang:commons-lang:2.4"` from which we got `EqualsBuilder`. Did you ask about that?

src/main/groovy/org/mockito/release/notes/model/Commit.java
szczepiq: I see. Thank you for clarifying this one!

src/test/groovy/org/mockito/release/notes/vcs/GitCommitSerializerTest.groovy
szczepiq: That works, thank you for checking!

src/main/groovy/org/mockito/release/notes/vcs/IgnoredCommitProvider.java
```diff
@@ -1,24 +0,0 @@
-package org.mockito.release.notes.vcs;
-
-import java.util.HashMap;
-
-public class IgnoredCommitProvider {
```
szczepiq: Nice! Thank you for making the contribution simpler and easier to maintain!

PR #159, opened 2017-05-19
Improved test coverage of ReleaseNeededTask

wwilk:
Since ReleaseNeededTask logic is a little complex already and it may grow, I thought it would be good to make it more reliable by adding unit tests for the value of releaseNotNeeded variable.

@szczepiq 
I haven't included skipEnvVariable in these tests yet, because it's an environment variable. Is there any reason why it can't be a system property? It would be easier for testing. If it has to stay env variable, there are some hacks we can use to set its value in tests.

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/DefaultEnvPropertyAccessor.java
+++ b/src/main/groovy/org/mockito/release/internal/DefaultEnvPropertyAccessor.java
@@ -1,0 +1,11 @@
+package org.mockito.release.internal;
+
+import org.mockito.release.internal.util.EnvPropertyAccessor;
+
+public class DefaultEnvPropertyAccessor implements EnvPropertyAccessor{
+
+    @Override
+    public String getenv(String name) {
+        return System.getenv(name);
+    }
+}
--- a/src/test/groovy/org/mockito/release/gradle/ReleaseNeededTaskTest.groovy
+++ b/src/test/groovy/org/mockito/release/gradle/ReleaseNeededTaskTest.groovy
@@ -4,11 +4,46 @@ import org.gradle.api.GradleException
 import org.gradle.testfixtures.ProjectBuilder
 import org.mockito.release.internal.comparison.PublicationsComparatorTask
 import spock.lang.Specification
+import spock.lang.Unroll
 
 class ReleaseNeededTaskTest extends Specification {
 
     def tasks = new ProjectBuilder().build().getTasks();
     ReleaseNeededTask underTest = tasks.create("releaseNeeded", ReleaseNeededTask)
+
+    @Unroll
+    def "release is needed" (commitMessage, branch, pullRequest, publicationsComparatorsResults, releaseNotNeeded){
+        given:
+        underTest.setCommitMessage(commitMessage)
+        underTest.setBranch(branch)
+        underTest.setPullRequest(pullRequest)
+        int i = 0;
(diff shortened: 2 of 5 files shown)


src/test/groovy/org/mockito/release/gradle/ReleaseNeededTaskTest.groovy
```diff
@@ -4,12 +4,47 @@ import org.gradle.api.GradleException
...
+            def task = tasks.create("" + i++, PublicationsComparatorTask)
+            task.setPublicationsEqual(it)
+            underTest.addPublicationsComparator(task)
+        }
+        underTest.setReleasableBranchRegex("master")
+
+        expect:
+        underTest.releaseNeeded()
+        
+        underTest.isReleaseNotNeeded() == releaseNotNeeded
```
szczepiq: Nitpick: can we get rid of "isReleaseNotNeeded()" method and instead just make "releaseNeeded()" return boolean? This will simplify the code a little bit.

src/test/groovy/org/mockito/release/gradle/ReleaseNeededTaskTest.groovy
```diff
@@ -4,12 +4,47 @@ import org.gradle.api.GradleException
...
 import spock.lang.Specification
+import spock.lang.Unroll
 
 class ReleaseNeededTaskTest extends Specification {
 
     def tasks = new ProjectBuilder().build().getTasks();
     ReleaseNeededTask underTest = tasks.create("releaseNeeded", ReleaseNeededTask)
 
+    @Unroll
+    def "release is needed" (commitMessage, branch, pullRequest, publicationsComparatorsResults, releaseNotNeeded){
```
szczepiq: This is a very good test. Normally I'm against using imperative loops inside test method but in this case it makes sense to get good coverage. It covers the conditional logic that is likely to break when we change stuff. Thank you!

szczepiq:
>I haven't included skipEnvVariable in these tests yet

The reason we use env variable is because it is easier to use when setting up matrix of builds in Travis configuration file (see [example from Mockito project](https://github.com/mockito/mockito/blob/release/2.x/.travis.yml#L34)).

Instead of accessing env variables directly, we can wrap them in some type, use constructor injection, mock it in test and prosper :)

Absolutely approved merge at will!

wwilk:
Thanks for the suggestions! I applied them all and extracted System.getenv into a separate object. Just to be sure I tested it on example project - it still recognizes env property.

src/main/groovy/org/mockito/release/internal/DefaultEnvPropertyAccessor.java
```diff
@@ -0,0 +1,11 @@
+package org.mockito.release.internal;
+
+import org.mockito.release.internal.util.EnvPropertyAccessor;
+
+public class DefaultEnvPropertyAccessor implements EnvPropertyAccessor{
```
szczepiq: Nitpicky:
- I would probably not create an interface because it's an internal env variable wrapper so we can keep things simple
- Use "env variable" instead of "env property" in the name. Example class name: "EnvVariables".
- Add tiny javadoc code comment on the class explaining why we're wrapping access to env variables

szczepiq:
I made a nitpicky comment to help driving even cleaner code, based on my subjective assessment :) It's not blocking the merge in any way, so I'm merging.

PR #161, opened 2017-05-20
Configuration loaded automatically from shipkit.gradle

wwilk:
The config file is created automatically (if doesn't exist) and filled with example values of all needed properties. I wonder what I should do after creating the file - idea is to throw exception then, telling user that he/she should configure this file and rerun the task (see ReleaseConfigurationPlugin#61). It seems easier to use, since when we configure ReleaseConfiguration properties to example values,  it can fail anywhere, and in this case error message may be misleading. The only problem is that in this case we would have to create this file first in unit tests for each Plugin depending on ReleaseConfigurationPlugin, because otherwise this test would throw exception.

@szczepiq What do you think?


src/test/groovy/org/mockito/release/internal/gradle/ConfigurationFileBuilderTest.groovy
```diff
@@ -0,0 +1,29 @@
...
+                .withProperty("gitHub.repository", "mockito/mockito")
+                .withProperty("git.user", "Mockito Release Tools")
+                .withProperty("team.developers", new ArrayList())
+                .withProperty("team.contributors", Arrays.asList("wwilk:Wojtek Wilk", "mstachniuk:Marcin Stachniuk"))
+                .withProperty("releaseNotes.labelMapping", ["noteworthy":"Noteworthy","bugfix":"Bugfixes"])
+                .withProperty("releaseNotes.labelMapping2", [:])
+                .withProperty("gitHub.writeAuthToken", new ConfigurationFileBuilder.Expression("System.getenv(\"TOKEN\")"))
+                .build()
+
+        content == "releasing {\n" +
```
szczepiq: Using Groovy's multiline String will probably give us cleaner test. Try it!

wwilk:
@szczepiq 
I applied the suggestions. I created a base class for plugin tests, it's extendable and avoids duplication, but maybe makes tests a little less readable. What do you think? 

I configured some defaults, as we discussed on Slack.

Right now I started with throwing exception if file doesn't exist. Maybe we can discuss it later, along with other team members what would be the best solution in this case? Right now the important part is to have configuration from shipkit.gradle applied at the beginning, so that configuration is easier.

src/test/groovy/testutil/PluginSpecification.groovy
```diff
@@ -0,0 +1,28 @@
+package testutil
+
+import org.gradle.testfixtures.ProjectBuilder
+import org.junit.Rule
+import org.junit.rules.TemporaryFolder
+import spock.lang.Specification
+
+class PluginSpecification extends Specification{
```
szczepiq: I like the new class a lot! It also hides other complexities like 'tmp' dir, creation of 'project'.

Let's add some Javadocs on why and how

src/test/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPluginTest.groovy
```diff
@@ -55,4 +61,37 @@ class ReleaseConfigurationPluginTest extends Specification {
         expect:
         root.plugins.apply(ReleaseConfigurationPlugin).configuration.notableRelease
     }
+
+    def "creates shitpkit.gradle file if it doesn't exist"() {
```
szczepiq: I like this test a lot! It offers complete coverage for generated template.

src/test/groovy/org/mockito/release/internal/gradle/TemplateResolverTest.groovy
```diff
@@ -0,0 +1,81 @@
...
+
+    def "should fail on null key" (){
+        given:
+        def resolver = new TemplateResolver("test")
+
+        when:
+        resolver.withProperty(null, "value")
+
+        then:
+        def ex = thrown(IllegalArgumentException)
```
szczepiq: Nice! Thank you for covering edge cases.

src/test/groovy/org/mockito/release/internal/gradle/TemplateResolverTest.groovy
```diff
@@ -0,0 +1,81 @@
...
+                    repo = 'examples'
+                    licenses = ['MIT']
+                    labels = ['continuous delivery', 'release automation', 'shipkit']
+                }
+            }
+        }
+    }
+"""
+
+    def "should resolve template" (){
```
szczepiq: I suspect that you can trim this test - already cover generation of the template. Here we might just test a simple template with few tokens to replace.

src/test/groovy/org/mockito/release/internal/gradle/AutoVersioningPluginTest.groovy
```diff
@@ -1,11 +1,8 @@
 package org.mockito.release.internal.gradle
 
-import org.gradle.testfixtures.ProjectBuilder
-import spock.lang.Specification
+import testutil.PluginSpecification
 
-class AutoVersioningPluginTest extends Specification {
-
-    def project = new ProjectBuilder().build()
```
szczepiq: nice :) I like removed code in the morning.

src/main/groovy/org/mockito/release/internal/gradle/TemplateResolver.java
```diff
@@ -0,0 +1,36 @@
+package org.mockito.release.internal.gradle;
+
+import org.mockito.release.internal.util.ArgumentValidation;
+
+import java.util.HashMap;
+import java.util.Map;
+
+public class TemplateResolver {
```
szczepiq: I like this new template resolver! Let's add some Javadocs on why & how.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
```diff
@@ -43,10 +57,66 @@ public void apply(Project project) {
         }
     }
 
+    private void loadShipKitConfigFile(Project rootProject) {
+        final File configFile = rootProject.file(CONFIG_FILE_RELATIVE_PATH);
+        if (!configFile.exists()) {
+            createShipKitConfigFile(configFile);
+            throw new GradleException("Config file created at " + configFile.getAbsolutePath() + ". Please configure it and rerun the task.");
+        } else {
+            rootProject.apply(new Action<ObjectConfigurationAction>() {
```
szczepiq: Thank you for updating the code so that it uses Java types instead of the Map!

Can you update the Javadoc of this plugin class to explain how we handle the configuration? This way we will have comprehensive documentation.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
```diff
@@ -19,6 +29,8 @@
 
     private ReleaseConfiguration configuration;
 
+    private static final String CONFIG_FILE_RELATIVE_PATH = "gradle/shipkit.gradle";
```
szczepiq: Consider making it public and reusing in the places where refer to this file using String literal. It would be slightly more cleaner I think.

src/test/groovy/testutil/PluginSpecification.groovy
wwilk: Done :)

src/test/groovy/org/mockito/release/internal/gradle/TemplateResolverTest.groovy
wwilk: Yeah, you are right!

wwilk:
Applied suggestions and merging.

PR #162, opened 2017-05-20
EnvPropertyAccessor improved

wwilk:
@szczepiq 
Suggestions taken :)


szczepiq:
Nice, no more suggestions ;)

Issue #157, opened 2017-05-19
Discontinue notable release notes file


szczepiq:
### Problem

Maintaining notable release notes file is problematic (the logic becomes complicated). From the user standpoint, browsing 2 separate files is also not ideal.

### Plan

- [x] - Remove notable release notes generation logic for now
- [x] - Manually update Mockito's release notes so that [release notes](https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md) are clear.
- [x] - Update Mockito to use latest tools
- [x] - Update Mockito release pipeline description (it mentions notable versions file and the concept). See https://github.com/mockito/mockito/pull/1120

### Future

Let's see how it goes with a single release notes file without the concept of notable releases. Let's keep things simple. We can improve as we gather more feedback and perhaps resurrect the idea of notable release notes. Mockito will still be configured to release to Maven Central periodically.

szczepiq:
Completed!

PR #163, opened 2017-05-21
Removed the concept of notable release for now

szczepiq:
For motivation, see #157. Let's cut down features before 1.0 to keep the library simple. Best to review commit by commit.

 - removed 'notable releases' concept for now. I left some internal logic for now in case we need to salvage it.
 - refactorings, tidy-ups, javadoc

Tested with:
 - mockito
 - example

### DIFF
--- a/src/main/groovy/org/mockito/release/version/DefaultVersionInfo.java
+++ b/src/main/groovy/org/mockito/release/version/DefaultVersionInfo.java
@@ -69,6 +69,10 @@ class DefaultVersionInfo implements VersionInfo {
         return previousVersion;
     }
 
+    public DefaultVersionInfo bumpVersion() {
+        return bumpVersion(false);
+    }
+
     public DefaultVersionInfo bumpVersion(boolean updateNotable) {
         String content = IOUtil.readFully(versionFile);
         if (updateNotable) {
@@ -103,6 +107,8 @@ class DefaultVersionInfo implements VersionInfo {
     }
 
     public Collection<String> getNotableVersions() {
+        //TODO if by the end of Q3 (9/30/17) we don't find the concept of 'notable versions' useful
+        //we should delete 'notableVersions' code from this class
         return notableVersions;
     }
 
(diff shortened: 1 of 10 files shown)


szczepiq:
Friendly ping :)

src/main/groovy/org/mockito/release/version/DefaultVersionInfo.java
```diff
@@ -103,11 +107,12 @@ public DefaultVersionInfo bumpVersion(boolean updateNotable) {
     }
 
     public Collection<String> getNotableVersions() {
+        //TODO if by the end of Q3 (9/30/17) we don't find the concept of 'notable versions' useful
+        //we should delete 'notableVersions' code from this class
         return notableVersions;
```
mstachniuk: Maybe is it better to delete it now and add it later (if needed), because of compatibility after 1.0?

src/main/groovy/org/mockito/release/version/DefaultVersionInfo.java
```diff
@@ -103,11 +107,12 @@ public DefaultVersionInfo bumpVersion(boolean updateNotable) {
...
     }
 
     public boolean isNotableRelease() {
```
mstachniuk: It can be also deleted

src/main/groovy/org/mockito/release/version/VersionInfo.java
```diff
@@ -20,15 +18,8 @@
...
-    VersionInfo bumpVersion(boolean updateNotable);
-
-    /**
-     * Returns notable versions
      */
-    Collection<String> getNotableVersions();
+    VersionInfo bumpVersion();
 
     /**
      * Informs if the current version is a notable release
```
mstachniuk: You can delete `isNotableRelease()`

szczepiq: Great catch. I missed that. It's a part of public API and it needs to be cleaned.

szczepiq:
Great feedback, thank you!

I hear you that it would be good to remove all notable releases logic. Perhaps you're right ;)

For now, my intention is to remove that from the public API but leave the implementation because I think it will be useful. We did that before with jar/pom comparison logic that was unhooked from the public API until @wwilk plugged that back in. That worked ok.

I'll merge that PR if that's ok. We can have a separate discussion / create GitHub ticket to remove the implementation of notable notes completely. I'm not in favor of this but if the team wishes so, we can have a ticket for someone to do the work. At the moment, it's important to thin down the public API for 1.0 (pragmatic :D)

>The same in README.md

I will update README.md in the next PR. Thanks for the reminder.

Issue #155, opened 2017-05-18
Commit skipping logic is hardcoded and yields subtle bugs in release notes content


szczepiq:
## Problem

For generating release notes with number of commits and contributors, we skip "automated" commits. It is an important feature because the number of commits is off if we consider commits made by release automation. The automated skip message is configurable by the user and Mockito project [uses it](https://github.com/mockito/mockito/blob/release/2.x/gradle/root/release.gradle#L30). However, the implementation of shipkit is hardcoded to ignore commits with [ci skip] - [see the code](https://github.com/mockito/mockito-release-tools/blob/822c8f8a06c5c82e924dc6ba95998af86450d2f6/src/main/groovy/org/mockito/release/notes/vcs/IgnoreCiSkip.java#L17).

Even though the skip message is hardcoded, the recent Mockito release notes seems to calculate commits correctly. Why? :)

## Suggested change

For the time being, let's ignore all commits with "[ci skip]" and also all commits that contain releasing.git.commitMessagePostfix. Code hints:

- Code that contains hardcoded ["[ci skip]"](https://github.com/mockito/mockito-release-tools/blob/822c8f8a06c5c82e924dc6ba95998af86450d2f6/src/main/groovy/org/mockito/release/notes/vcs/IgnoreCiSkip.java#L17)
- Code that contains [commitMessagePostfix configuration](https://github.com/mockito/mockito-release-tools/blob/3abab2fd4de535f4111a838ee35e04ee95c37a0e/gradle/release.gradle#L12)

szczepiq:
@NagRock, do you want to take a stab at it? :)

NagRock:
Yea! You can assign me

PR #164, opened 2017-05-21
#155 Commits skipping logic now uses configuration file

NagRock:
#155 
`NotableReleaseNotesFetcherTask` and `ReleaseNotesFetcherTask` uses different way of getting configuration data, should I fix it? Or maybe one of this task is outdated and should be removed?


src/main/groovy/org/mockito/release/notes/vcs/IgnoreCommitByConfigMessagePostfix.java
```diff
@@ -0,0 +1,18 @@
...
+
+    public IgnoreCommitByConfigMessagePostfix(String postfix) {
+        this.postfix = postfix;
+    }
+
+    @Override
+    public boolean isTrue(Commit commit) {
+        final int expectedPostfixIndex = commit.getMessage().length() - postfix.length();
+        return commit.getMessage().lastIndexOf(postfix) == expectedPostfixIndex;
+    }
```
wwilk: You can use String#endsWith method instead.

src/main/groovy/org/mockito/release/notes/vcs/GitContributionsProvider.java
```diff
@@ -27,7 +27,7 @@ public ContributionSet getContributionsBetween(String fromRev, String toRev) {
 
         DefaultContributionSet contributions = new DefaultContributionSet();
         for (Commit commit : commits) {
-            if (!ignoredCommit.isTrue(commit)) {
+            if (ignoredCommit.isTrue(commit)) {
```
wwilk: You forgot about renaming the field - name "ignoredCommit" suggests that we add ignored commits to contributions

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
```diff
@@ -0,0 +1,52 @@
...
+import spock.lang.Specification
+
+class DefaultCommitApproverTest extends Specification {
+
+    def "should approve commit with sample message"() {
+        def approver = new DefaultCommitApprover()
+        def commit = new GitCommit("sampleId", "sample@email.com", "sampleAuthor", "sample commit message")
+
+        when:
+        def result = approver.isTrue(commit)
```
wwilk: You can use

expect:
approver.isTrue(commit) 

instead

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
```diff
@@ -0,0 +1,52 @@
...
+
+        then:
+        result == false
+    }
+
+    def "should decline commit with defined postfix in message"() {
+        def samplePostfix = "samplePostfix"
```
wwilk: I would name it more like "should decline commit with postfix defined at the end of the message"

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
```diff
@@ -0,0 +1,52 @@
+package org.mockito.release.notes.vcs
+
+import spock.lang.Specification
+
+class DefaultCommitApproverTest extends Specification {
```
wwilk: This test covers implementation of IgnoreCiSkip and IgnoreCommitByConfigMessagePostfix. I would cover it in dedicated tests for these classes and here I would mock all predicates and check if method isTrue() works correctly for empty list of predicates, for predicates that are all false, all true, some of them false some true.

src/main/groovy/org/mockito/release/notes/vcs/GitContributionsProvider.java
NagRock: Ahh, thanks I've changed it before but later reverted all changes.

src/main/groovy/org/mockito/release/notes/vcs/DefaultCommitApprover.java
```diff
@@ -0,0 +1,31 @@
+package org.mockito.release.notes.vcs;
+
+import org.mockito.release.notes.model.Commit;
+import org.mockito.release.notes.util.Predicate;
+
+import java.util.Collection;
+import java.util.LinkedList;
+
+public class DefaultCommitApprover implements Predicate<Commit> {
+
```
wwilk: Maybe CommitIgnored would be a better name since we use Ignore* predicates inside, and reversing logic makes it more complicated than it needs to be. Then it would be used like commitIgnored.isTrue(commit). Seems more readable.

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
NagRock: Its all about what we call unit, but in this case I agree that we can later have more approvers and its better to cover every class in different tests. I will fix it.

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
NagRock: I think that postfix means itself that it is at the and, same as prefix is always at the beginning.

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
NagRock: Then I should remove `when` and `then` right? I always used to have `given`, `when`, `then` blocks but understand thats its a convention and I should replace `result`, `when:`, and `then:` with `expect:` right?

wwilk: Yes exactly, btw "given:" can be done with spock too :)

src/main/groovy/org/mockito/release/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -80,6 +80,7 @@ public void run() {
                         t.setGitHubReadOnlyAuthToken(conf.getGitHub().getReadOnlyAuthToken());
                         t.setGitHubRepository(conf.getGitHub().getRepository());
                         t.setPreviousVersion(conf.getPreviousReleaseVersion());
+                        t.setSkipCommitMessagePostfix(conf.getGit().getCommitMessagePostfix());
```
szczepiq: Hmmm, I think this will not work. The design I suggested earlier has a flaw. The commitMessagePostfix is different with each release build because it includes Travis build number.

I pushed a meaty TODO to this branch with suggested update to the design. Hopefully, it will work better!

Thank you @NagRock for working on this! Removing hardcoded values and making the library more flexible will work in our favor and help getting wider adoption!

src/main/groovy/org/mockito/release/notes/vcs/DefaultCommitApprover.java
```diff
@@ -0,0 +1,31 @@
+package org.mockito.release.notes.vcs;
+
+import org.mockito.release.notes.model.Commit;
+import org.mockito.release.notes.util.Predicate;
+
+import java.util.Collection;
+import java.util.LinkedList;
+
+public class DefaultCommitApprover implements Predicate<Commit> {
```
szczepiq: Design suggestion: it might be simpler to just have a single implementation of the Predicate, and make the impl work with a list of Strings. Something like that: ```new DefaultCommitApprover("[ci skip]", "[ci skip-release]")```.

The simpler code, the easier testing and easier to maintain :)

src/main/groovy/org/mockito/release/gradle/ReleaseNotesFetcherTask.java
```diff
@@ -174,6 +174,24 @@ public String getSkipCommitMessagePostfix() {
...
+        Therefore we cannot really depend on it to exclude commits reliably.
+
+        I suggest that we introduce a new setting to configure how to exclude commits from release generation.
+        This will make it explicit and easy to tweak by the user.
+        Suggested plan
+
+        1. New setting: "releasing.releaseNotes.ignoreCommitsContaining", Collection<String>
+        2. By default, we would configure it to: "ignoreCommitsContaining" = ['[ci skip]']
+        3. Users can tweak it. For example, in Mockito project, in "release.gradle" we would do:
+            releasing.releaseNotes.ignoreCommitsContaining = ['[ci skip]', '[ci skip-release]']
```
NagRock: Looks good but to be sure, even `[ci skip]` will be then configurable right?

src/main/groovy/org/mockito/release/notes/vcs/DefaultCommitApprover.java
```diff
@@ -0,0 +1,31 @@
...
+
+    private Collection<Predicate<Commit>> ignoredCommits = new LinkedList<Predicate<Commit>>();
+
+    public DefaultCommitApprover() {
```
szczepiq: Do we need this? It seems it's used only for testing.

src/main/groovy/org/mockito/release/gradle/ReleaseNotesFetcherTask.java
```diff
@@ -160,10 +162,25 @@ public void setGitHubLabels(Collection<String> gitHubLabels) {
         this.gitHubLabels = gitHubLabels;
     }
 
+    /**
+     * Configurable commit message postfix that will cause with commit skipping in release notes
```
NagRock: Thanks, didn't know about that! I will fix it.

src/main/groovy/org/mockito/release/notes/vcs/DefaultCommitApprover.java
NagRock: Yep, I will remove it.

src/main/groovy/org/mockito/release/notes/vcs/DefaultCommitApprover.java
NagRock: But should then all skip `messages` be postfix? Earlier `[ci skip]` could be in any index of commit message.

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
wwilk: True but in this case postfix is both a parameter and part of the message. It is not clear which of that you mean in the test name. And in the next test it is clear

src/test/groovy/org/mockito/release/notes/vcs/DefaultCommitApproverTest.groovy
wwilk: Thanks!

src/main/groovy/org/mockito/release/gradle/ReleaseNotesFetcherTask.java
szczepiq: Yes. It will be configurable. However, most users will not change the default values. We will preconfigure it to '[ci skip]'.

We don't want such settings to be hardcoded in the tools. There are potential use cases with different CI systems, different requirements where users might want to skip different commits. Also, there might be users who disagree that we should exclude '[ci skip]' because they want to include all commits in the notes.

Your work on this feature is an example of our direction - we build a flexible library :) When new use case / requirement gets in, we try to make the tool more flexible to support it.

gradle/release.gradle
```diff
@@ -11,7 +11,7 @@ releasing {
     def buildNo = System.getenv("TRAVIS_BUILD_NUMBER")
     git.commitMessagePostfix = buildNo? " by CI build $buildNo [ci skip-release]" : " [ci skip-release]"
 
-//    releaseNotes.ignoreCommitsContaining = ['[ci skip]']
+    releaseNotes.ignoreCommitsContaining = ['[ci skip]']
```
mstachniuk: Please remove this line for good!

src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java
```diff
@@ -367,7 +380,7 @@ private String getString(String key, String envVarName) {
     }
 
     private Map getMap(String key) {
-        return (Map) getValue(key, null,"Please configure 'releasing." + key + "' value (Map).");
+        return (Map) getValue(key, null, "Please configure 'releasing." + key + "' value (Map).");
```
mstachniuk: you can define a constant: `private static final String NO_ENV_VARIABLE = null` and pass it instead of `null`. Then it will be easier to read this line

gradle/release.gradle
NagRock: Done

src/test/groovy/org/mockito/release/notes/vcs/IgnoredCommitTest.groovy
```diff
@@ -0,0 +1,24 @@
+package org.mockito.release.notes.vcs
+
+import spock.lang.Specification
+
+class IgnoredCommitTest extends Specification {
+
```
wwilk: I would include the case when list of commitMessageParts is empty.

src/test/groovy/org/mockito/release/notes/vcs/IgnoredCommitTest.groovy
```diff
@@ -0,0 +1,24 @@
...
+        def commitWithCiSkip = new GitCommit("firstCommitId", "sample@email.com", "sampleAuthor", "sample [ci skip] commit message")
+        def commitWithCustomSkip = new GitCommit("thirdCommitId", "sample@email.com", "sampleAuthor", "sample [custom skip] commit message")
+
+        expect:
+        ignoredCommit.isTrue(commitWithCiSkip)
+        ignoredCommit.isTrue(commitWithCustomSkip)
+    }
+
+    def "should not skip commit that does not contain given ignored substrings"() {
+        def ignoredCommit = new IgnoredCommit(["[ci skip]", "[custom skip]"])
```
wwilk: Second method uses the same IgnoredCommit instance so I think only one test method would be better. Especially since this is a quite straightforward logic and only parameter that changes (and matters) is message content.

src/test/groovy/org/mockito/release/notes/vcs/IgnoredCommitTest.groovy
```diff
@@ -0,0 +1,24 @@
...
+        def commit = new GitCommit("firstCommitId", "sample@email.com", "sampleAuthor", "sample commit message")
```
wwilk: You can create a private method like createCommitWithMessage(message) because other properties of the commit don't matter and only make this test less readable

wwilk:
I added a few small comments to IgnoredCommitTest, despite that good job!

src/test/groovy/org/mockito/release/notes/vcs/IgnoredCommitTest.groovy
NagRock: What would you say about `GitCommitBuilder` in `test/java/builders` folder? Then it could be used in other tests.

wwilk: It's a good idea if GitCommit is created also in other places. However in this particular case the simplest solution would be to create a private method, that would instantiate a GitCommit. Since only message is important, you can randomly generate other properties, see RandomStringUtils from commons-lang, it's already on classpath.

PR #165, opened 2017-05-21
GitAdd task removed and GitCommit made generic

wwilk:
GitCommit now gathers all the files that were changed (and appropriate commit messages) and commits it all together at once. Additionally default values are set for release notes output files. 

@szczepiq 
This is not final yet (especially GitCommitTask and GitPushPlugin need better test coverage) but I wanted to get to know your opinion before I do these tests. Additionally we need some mechanism to use ReleaseNotesPlugin and GitPushPlugin together. Please see how it's done with VersioningPlugin and AutoVersioningPlugin. Maybe we can create AutoReleaseNotesPlugin that would add a task "updateReleaseNotesAndPush" which would depend on PERFORM_GIT_PUSH and UPDATE_RELEASE_NOTES. Another option would be to do it in client code, but I would prefer to have it automated.


src/main/groovy/org/mockito/release/internal/gradle/AutoVersioningPlugin.java
```diff
@@ -28,67 +28,23 @@
...
  * <ul>
- *     <li>gitAddBumpVersion</li>
- *     <li>gitCommitBumpVersion</li>
- *     <li>gitPushBumpVersion</li>
  *     <li>bumpVersionAndPush</li>
  * </ul>
  */
 public class AutoVersioningPlugin implements Plugin<Project> {
 
     static final String BUMP_VERSION_AND_PUSH_TASK = "bumpVersionAndPush";
```
szczepiq: Let's start avoiding 'and' in the name of the tasks for clarity and ease of use. I suggest something like that: "pushBumpVersion" or "performBumpVersion". The latter is interesting. We can come up with a naming convention that tasks that start with "perform" are coarse grained, high level tasks.

src/main/groovy/org/mockito/release/internal/gradle/AutoVersioningPlugin.java
```diff
@@ -28,67 +28,23 @@
...
-    static final String ADD_BUMP_VERSION_TASK = "gitAddBumpVersion";
-    static final String PUSH_BUMP_VERSION_TASK = "gitPushBumpVersion";
-    static final String VERSION_BUMP_COMMIT_TASK = "gitCommitBumpVersion";
 
     public void apply(final Project project) {
```
szczepiq: It's great when we have lots of plugins and top-level plugins mostly apply some other fine-grained plugins. It will make the codebase easier to work with and make the tool really flexible for advanced users. This plugin is a good example - mostly it applies other plugins. I would love the "continuous-delivery" plugin to be like that - only applies other plugins, zero other code.

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -146,21 +130,14 @@ public void run() {
...
         TaskMaker.task(project, "performRelease", new Action<Task>() {
             public void execute(final Task t) {
                 t.setDescription("Performs release. " +
                         "Ship with: './gradlew performRelease -Preleasing.dryRun=false'. " +
                         "Test with: './gradlew testRelease'");
 
                 t.dependsOn(VersioningPlugin.BUMP_VERSION_FILE_TASK, "updateReleaseNotes", "updateNotableReleaseNotes");
-                t.dependsOn(AutoVersioningPlugin.ADD_BUMP_VERSION_TASK, "gitAddReleaseNotes", GitPlugin.COMMIT_TASK, GitPlugin.TAG_TASK);
-                t.dependsOn(GitPlugin.PUSH_TASK);
+                t.dependsOn(GitPushPlugin.PERFORM_GIT_PUSH_TASK);
```
szczepiq: +1

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -81,27 +80,12 @@ public void execute(Task t) {
...
-            }
-        });
-
         final Task bintrayUploadAll = TaskMaker.task(project, "bintrayUploadAll", new Action<Task>() {
             public void execute(Task t) {
                 t.setDescription("Depends on all 'bintrayUpload' tasks from all Gradle projects.");
                 //It is safer to run bintray upload after git push (hard to reverse operation)
                 //This way, when git push fails we don't publish jars to bintray
-                t.mustRunAfter(GitPlugin.PUSH_TASK);
+                t.mustRunAfter(GitPushPlugin.GIT_PUSH_TASK);
```
szczepiq: +1

src/main/groovy/org/mockito/release/internal/gradle/AutoVersioningPlugin.java
```diff
@@ -28,67 +28,23 @@
...
-                });
-            }
-        });
 
         TaskMaker.task(project, BUMP_VERSION_AND_PUSH_TASK, new Action<Task>() {
             public void execute(Task t) {
                 t.setDescription("Increments version number, commits and pushes changes to Git repository");
-                t.dependsOn(VersioningPlugin.BUMP_VERSION_FILE_TASK, ADD_BUMP_VERSION_TASK);
-                t.dependsOn(VERSION_BUMP_COMMIT_TASK, PUSH_BUMP_VERSION_TASK);
+                t.dependsOn(VersioningPlugin.BUMP_VERSION_FILE_TASK);
```
szczepiq: +1, +1, task dependencies look good. They are especially interesting when reviewing code - it's critical that we get the task dependencies right (with all the must run after, depends on, etc.)

src/main/groovy/org/mockito/release/internal/gradle/GitCommitTask.java
```diff
@@ -0,0 +1,39 @@
...
+import java.io.File;
+import java.util.ArrayList;
+import java.util.List;
+
+public class GitCommitTask extends Exec{
+
+    private List<File> filesToCommit = new ArrayList<File>();
+    private List<String> messages = new ArrayList<String>();
+
+    public void addChange(List<File> files, String message, Task taskMakingChange){
```
szczepiq: Nitpick: let's call it 'description' as a description of the change. 'message' might be confused with commit message which is something different, something constructed from descriptions.

src/main/groovy/org/mockito/release/internal/gradle/VersioningPlugin.java
```diff
@@ -62,9 +63,16 @@ public void execute(Project project) {
         });
 
         TaskMaker.task(project, BUMP_VERSION_FILE_TASK, BumpVersionFileTask.class, new Action<BumpVersionFileTask>() {
-            public void execute(BumpVersionFileTask t) {
+            public void execute(final BumpVersionFileTask t) {
                 t.setVersionFile(versionFile);
                 t.setDescription("Increments version number in " + versionFile.getName());
+                project.getPlugins().withType(GitPushPlugin.class, new Action<GitPushPlugin>() {
```
szczepiq: To reduce duplication we might consider pushing this to some utility method

src/main/groovy/org/mockito/release/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -158,4 +162,16 @@ private static void configureNotableReleaseNotes(Project project) {
             fetcherTask.getNotesGeneration().setHeadVersion(project.getVersion().toString());
         }
     }
+
+    private static void addToCommittedChanges(final Project project, final Task task, final File changedFile, final String message){
+        project.getPlugins().withType(GitPushPlugin.class, new Action<GitPushPlugin>() {
+            @Override
+            public void execute(GitPushPlugin gitPushPlugin) {
+                final GitCommitTask gitCommitTask = (GitCommitTask) project.getTasks().getByName(GitPushPlugin.GIT_COMMIT_TASK);
+                gitCommitTask.mustRunAfter(task);
```
szczepiq: We already have 'dependsOn' inside 'addChange' method so this line is probably not needed.

I need to think a little bit about when to use 'dependsOn' and when to use 'mustRunAfter' but it should not block you in a any way. 'mustRunAfter' is convenient because allows to us run single task easily, in case we just want to run git commit task. However, it introduces complexity because we need to manage tasks dependsOn somewhere :)

src/main/groovy/org/mockito/release/internal/gradle/AutoVersioningPlugin.java
wwilk: I like that idea with high level "perform" tasks :)

src/main/groovy/org/mockito/release/internal/gradle/AutoVersioningPlugin.java
```diff
@@ -28,67 +20,23 @@
...
  *
  * <ul>
- *     <li>gitAddBumpVersion</li>
- *     <li>gitCommitBumpVersion</li>
- *     <li>gitPushBumpVersion</li>
- *     <li>bumpVersionAndPush</li>
+ *     <li>performVersionBump</li>
  * </ul>
  */
 public class AutoVersioningPlugin implements Plugin<Project> {
```
szczepiq: This plugin is looking great!!! Little complexity, it aggregates other plugins

src/main/groovy/org/mockito/release/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -71,28 +70,12 @@ public void execute(Task t) {
         //TODO we should have tasks from the same plugin to have the same group
         //let's have a task maker instance in a plugin that has sets the group accordingly
 
-        //TODO use constants for all task names
-        TaskMaker.execTask(project, "gitAddReleaseNotes", new Action<Exec>() {
```
szczepiq: Fantastic change!!! I love when we cut down the complexity, especially when we remove complexity from the fattest plugin!

src/main/groovy/org/mockito/release/internal/gradle/GitPlugin.java
```diff
@@ -3,78 +3,104 @@
...
                         t.setSecretValue(conf.getGitHub().getWriteAuthToken());
                     }
                 });
             }
         });
 
+        TaskMaker.task(project, PERFORM_GIT_PUSH_TASK, Task.class, new Action<Task>() {
+            public void execute(final Task t) {
+                t.setDescription("Pushes automatically created commits to remote repo.");
+                t.dependsOn(GIT_COMMIT_TASK, GIT_PUSH_TASK);
```
szczepiq: I think we're missing dependsOn git tag here. This dependency was previously configured in the performRelease task. Without this dependency, git tag may run after git push and the build would break.

src/main/groovy/org/mockito/release/internal/gradle/util/GitPluginUtil.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.io.File;
+import java.util.List;
+
+/**
+ * Utility methods for {@link org.mockito.release.internal.gradle.GitPlugin}
+ */
+public class GitPluginUtil {
+
+    public static void registerChangesIfGitPluginApplied(final Project project, final List<File> changedFiles,
```
szczepiq: Great idea to push that complexity to separate method!

Feedback:
1. Consider moving this method directly to GitPlugin class because it is coupled with that plugin anyway. This would result in simpler code I think.
2. when you pass task instance to the method, you automatically have access to Gradle project object: ```task.getProject()```

src/test/groovy/org/mockito/release/internal/gradle/GitCommitTaskTest.groovy
```diff
@@ -0,0 +1,47 @@
+package org.mockito.release.internal.gradle
+
+import org.apache.commons.lang.RandomStringUtils
+import org.gradle.api.DefaultTask
+import org.gradle.api.Task
+import org.gradle.testfixtures.ProjectBuilder
+import spock.lang.Specification
+
+class GitCommitTaskTest extends Specification {
```
szczepiq: Nice and clean test. Thanks!

PR #167, opened 2017-05-23
#147 Some contributor links are missing in Mockito release notes, solution nr 2

mstachniuk:


src/main/groovy/org/mockito/release/gradle/contributors/ConfigureContributorsTask.java
```diff
@@ -1,71 +0,0 @@
...
-import java.util.List;
-
-import static org.mockito.release.internal.gradle.util.StringUtil.isEmpty;
-
-/**
- * Configures {@link ReleaseConfiguration.Team#getContributors()}
- * based on serialized contributors data.
- * See also {@link org.mockito.release.internal.gradle.ContributorsPlugin}
- */
-public class ConfigureContributorsTask extends DefaultTask {
```
szczepiq: Nice. I like when complexity is reduced :) Less tasks == better!

PR #168, opened 2017-05-24
Fix bug in deserialization in DefaultContributionSetSerializer

mstachniuk:


src/main/groovy/org/mockito/release/notes/vcs/DefaultContributionSetSerializer.java
```diff
@@ -39,7 +35,7 @@ public DefaultContributionSet deserialize(JsonObject jsonObject) {
 
     private DefaultContributionSet addCommits(DefaultContributionSet defaultContributionSet, JsonArray commits) {
         for (int i = 0; i < commits.size(); i++) {
```
wwilk: Is it possible to use for each loop here? JSONArray implements Iterable

src/main/groovy/org/mockito/release/notes/vcs/DefaultContributionSetSerializer.java
```diff
@@ -39,7 +35,7 @@ public DefaultContributionSet deserialize(JsonObject jsonObject) {
...
-            Commit commit = gitCommitSerializer.deserialize((JsonObject) commits.get(0));
+            Commit commit = gitCommitSerializer.deserialize((JsonObject) commits.get(i));
```
NagRock: DefaultContributionSerializer

NagRock: Same mistake is made in `DefaultContributionSerializer` could you fix it too?

src/main/groovy/org/mockito/release/notes/vcs/DefaultContributionSetSerializer.java
mstachniuk: Good catch

src/main/groovy/org/mockito/release/notes/vcs/DefaultContributionSetSerializer.java
mstachniuk: @NagRock Yes. Is class `DefaultContributionSerializer` needed? It looks like only used in test code?

src/test/groovy/org/mockito/release/internal/gradle/util/ReleaseNotesSerializerIntegrationTest.groovy
```diff
@@ -0,0 +1,39 @@
+package org.mockito.release.internal.gradle.util
+
+import org.mockito.release.notes.internal.DefaultImprovement
+import org.mockito.release.notes.internal.DefaultReleaseNotesData
+import org.mockito.release.notes.vcs.DefaultContributionSet
+import org.mockito.release.notes.vcs.GitCommit
+import spock.lang.Specification
+
+class ReleaseNotesSerializerIntegrationTest extends Specification {
+
```
wwilk: Why is this called *IntegrationTest?

src/test/groovy/org/mockito/release/internal/gradle/util/ReleaseNotesSerializerIntegrationTest.groovy
```diff
@@ -0,0 +1,39 @@
...
+                "0.3.5",
+                "0.3.6")
+        def releaseNotes = [releaseNote]
+
+        when:
+        def serializedJson = serializer.serialize(releaseNotes)
+        def result = serializer.deserialize(serializedJson)
+
+        then:
+        result.get(0) == releaseNote
```
wwilk: It would be good to also assert that result only contains one element

src/main/groovy/org/mockito/release/notes/format/DetailedFormatter.java
```diff
@@ -132,7 +132,7 @@ private static String link(String text, String link) {
     }
 
     private static String allAuthors(ContributionSet contributions, Map<String, Contributor> contributors) {
-        if (contributions.getContributions().size() <= MAX_AUTHORS) {
+        if (contributions.getAuthorCount() <= MAX_AUTHORS) {
```
szczepiq: Is this meaningful for fixing the issue or for clarity?

src/main/groovy/org/mockito/release/notes/internal/DefaultImprovement.java
```diff
@@ -87,4 +87,40 @@ public String toJson() {
     public void toJson(Writer writable) throws IOException {
         writable.append(toJson());
     }
+
+    @Override
+    public boolean equals(Object o) {
```
szczepiq: Are the equals/hashcode methods meaningful to fix the issue? Ideally, we don't have to maintain those methods unless we have to. They add complexity.

src/main/groovy/org/mockito/release/notes/format/DetailedFormatter.java
mstachniuk: This is for clarity, see another usage of `MAX_AUTHORS` in this file.

src/main/groovy/org/mockito/release/notes/internal/DefaultImprovement.java
mstachniuk: equals/hashcode methods are relevant for tests. Without it the current assertions in `DefaultContributionSetSerializerTest` will not work. 

The previous assertions:

```
EqualsBuilder.reflectionEquals(defaultContributionSet.getAllCommits(), deserializedData.getAllCommits())
```

also didn't work, I don't know why.

src/test/groovy/org/mockito/release/internal/gradle/util/ReleaseNotesSerializerIntegrationTest.groovy
mstachniuk: Because in this test `ReleaseNotesSerializer`'s dependencies are not mocked. `DefaultContributionSetSerializerTest` is unit test.

src/test/groovy/org/mockito/release/internal/gradle/util/ReleaseNotesSerializerIntegrationTest.groovy
mstachniuk: Done

src/main/groovy/org/mockito/release/notes/internal/DefaultImprovement.java
mstachniuk: And I added equals/hashCode implementation only to the "domain" classes which are mostly used with collections

szczepiq: Got it. Thank you explanation!

+1 to the fix
-1 to extra complexity due to equals+hashcode (extra code to maintain, we will have to always remember to update the methods when we change/refactor how the domain objects keep state. If we forget, we are exposed to subtle bugs).

One cool way of making assertions in tests for complex domain objects is using toString(), for example [here](https://github.com/mockito/mockito-release-tools/blob/3b98d8d413c12ba9ff3ad58a8a5c0b2981088ec9/src/test/groovy/org/mockito/release/notes/vcs/DefaultReleasedVersionsProviderTest.groovy#L42). Since toString() is used mostly for debugging, it's safer than real hashcode+equals. Plus, you can tweak toString() to show meaningful stuff, in compact form. 

No action needed, let's move on ;)

mstachniuk:
Fixed #184

PR #169, opened 2017-05-24
Shipkit.gradle loaded from relative not absolute path

wwilk:
Also redundant updateNotableReleaseNotes task removed and deferredConfiguration is also not needed anymore in ReleaseNotesPlugin


wwilk:
Merging temporarily to see if change in configuration fixes problem with pom generation on travis in mockito release example.

PR #170, opened 2017-05-25
Ensured git push runs as late as possible

### LINKED ISSUE #149: Configure git push task to run as late as possible
## Problem

The release automation pipeline contains multiple tasks, most of them are reversible and will not be problematic if the pipeline fails half way through. However, there are 2 tasks that are hard to reverse and not quite possible to make them atomic:
 - "git push" - pushing release notes and version bump commits to the repo
 - "bintray upload" - uploading binaries to Bintray

## Solution

tbd.

szczepiq:
1. Added initial integration test. It adds reasonable coverage and also helped me a lot implement git push / bintray upload ordering rules. We should grow our integration test suite and also improve the base class so that it is easy to write more integration tests. However, we should never have too many integration tests. Most code will be tested with unit tests (see testing pyramid pattern). There is a big fat TODO in the test. We will make it better soon.
2. git push now runs late in the release process ensuring that when there is a failure we don't push code first.
3. changed a bit how git push is configured so that the branch is configured in the 'doFirst' action. This should be fine for now.

Fixes #149

### DIFF
--- a/src/test/groovy/org/mockito/release/gradle/ContinuousDeliveryPluginIntegTest.groovy
+++ b/src/test/groovy/org/mockito/release/gradle/ContinuousDeliveryPluginIntegTest.groovy
@@ -1,0 +1,87 @@
+package org.mockito.release.gradle
+
+import testutil.GradleSpecification
+
+class ContinuousDeliveryPluginIntegTest extends GradleSpecification {
+
+    def "all tasks in dry run"() {
+
+        /**
+         * TODO this test is just a starting point we will make it better and create more integration tests
+         * Stuff that we should do:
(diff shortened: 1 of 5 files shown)


src/test/groovy/org/mockito/release/gradle/ContinuousDeliveryPluginIntegTest.groovy
```diff
@@ -0,0 +1,87 @@
+package org.mockito.release.gradle
+
+import testutil.GradleSpecification
+
+class ContinuousDeliveryPluginIntegTest extends GradleSpecification {
```
wwilk: 👍 Awesome! Great way to assure the order of tasks :)

PR #173, opened 2017-05-25
Updated dogfooding #87

wwilk:
- assertReleaseNeeded used in travis
- bumped version of MRT
- added shipkit.gradle, removed release.gradle
- fixed a small bug in gitPush args (tag was ommitted)
- description is no longer null in pom if not specified in the project
- increased info logging in PomComparator


wwilk:
Merging because I want to try to release the new version to see if it runs correctly on Travis

PR #175, opened 2017-05-25
AutoReleaseNotesPlugin introduced

wwilk:
AutoReleaseNotesPlugin, plugin similar functionally to AutoVersioningPlugin introduced. It is the previous to last step of release notes dogfooding in MRT. I'm not sure about naming though. @szczepiq what do you think?

I also removed deferredConfiguration for properties based on ReleaseConfiguration since it is now loaded eagerly.

### DIFF
--- a/src/main/groovy/org/mockito/release/internal/gradle/ContributorsPlugin.java
+++ b/src/main/groovy/org/mockito/release/internal/gradle/ContributorsPlugin.java
@@ -6,7 +6,6 @@ import org.gradle.api.Project;
 import org.mockito.release.gradle.ReleaseConfiguration;
 import org.mockito.release.internal.gradle.util.TaskMaker;
 
-import static org.mockito.release.internal.gradle.configuration.DeferredConfiguration.deferredConfiguration;
 import static org.mockito.release.internal.gradle.util.BuildConventions.contributorsFile;
 
 /**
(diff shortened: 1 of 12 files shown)


src/main/groovy/org/mockito/release/internal/gradle/ContributorsPlugin.java
```diff
@@ -32,15 +31,9 @@ public void execute(final AllContributorsFetcherTask task) {
                 task.setGroup(TaskMaker.TASK_GROUP);
                 task.setDescription("Fetch info about all project contributors from GitHub and store it in file");
                 task.setOutputFile(contributorsFile(project));
-
```
szczepiq: Very nice - complexity gets removed from the code!

src/test/groovy/org/mockito/release/internal/gradle/ContributorsPluginTest.groovy
```diff
@@ -5,6 +5,11 @@ import testutil.PluginSpecification
 class ContributorsPluginTest extends PluginSpecification {
 
     def "applies"() {
+        given:
+        def conf = applyReleaseConfiguration()
```
szczepiq: This seems like a logic that we could push out to setup() method of PluginSpecification, no? Repo and token are not relevant for this plugin so removing that setting from test makes the tests easier to read / maintain.

PR #177, opened 2017-05-26
Typo in AutoReleaseNotes plugin config file fixed #87

### LINKED ISSUE #87: Release tools project dogfoods versioning plugin

wwilk:


wwilk:
Merged because it's a bug and I want it fixed immediately

PR #181, opened 2017-05-27
Removed dead method and improved coverage

szczepiq:
Also identified missing unit test coverage


szczepiq:
Thank you for review!

PR #183, opened 2017-05-27
Bootstrap tasks for version.properties and shipkit.gradle

wwilk:
I created two tasks initVersioning and initConfigFile which create configuration files version.properties and shipkit.gradle. There is also a BoostrapPlugin that adds a task initShipkit which groups all these configuration tasks together.
An assumption was made that ./gradlew tasks can be run even when no configuration files exist, that's why all required properties (like github.repository) are set to some defaults. A different problem is bintray.user because it's not set in ReleaseConfiguration, so I moved validation nonNull(bintray.user) to doFirst of BintrayUpload instead of deferredConfiguration. I'm not sure if this is the best solution though, since it will fail later, but on the other hand soon enough so that this task can't be run.


src/main/groovy/org/mockito/release/internal/gradle/BintrayPlugin.java
```diff
@@ -41,20 +41,26 @@ public void apply(final Project project) {
...
+        bintray.setPublish(true);
+
         final BintrayUploadTask bintrayUpload = (BintrayUploadTask) project.getTasks().getByName(BINTRAY_UPLOAD_TASK);
 
         bintrayUpload.doFirst(new Action<Task>() {
             public void execute(Task task) {
                 //TODO unit test
                 BintrayUploadTask t = (BintrayUploadTask) task;
+
+                //workaround for https://github.com/bintray/gradle-bintray-plugin/issues/170
```
szczepiq: Given that your primary use case is making ```./gradlew tasks``` work, I suggest we use "lazyConfiguration" in this case. You can refer to the javadoc for LazyConfiguration class. If it is fuzzy, let me know so we make it clear! I love good docs ;)

src/main/groovy/org/mockito/release/internal/gradle/BootstrapPlugin.java
```diff
@@ -0,0 +1,27 @@
...
+import org.gradle.api.Task;
+import org.mockito.release.internal.gradle.util.TaskMaker;
+
+/**
+ * Creates task initShipkit that all other init tasks should depend on
+ * so that running it would create all configuration needed to start Shipkit
+ */
+public class BootstrapPlugin implements Plugin<Project> {
+
+    public static final String INIT_SHIPKIT_TASK = "initShipkit";
```
szczepiq: Let's use consistent naming. If the plugin is BootstrapPlugin, let's call the task BootstrapShipkit. If the plugin is InitPlugin, let's call the task "initShipkit". I think this will make our plugins easier to use.

src/main/groovy/org/mockito/release/internal/gradle/InitConfigFileTask.java
```diff
@@ -0,0 +1,75 @@
...
+            createShipKitConfigFile();
+            LOG.lifecycle("  Created config file at {}. Please configure it.", configFile.getPath());
+        }
+    }
+
+    private void createShipKitConfigFile() {
+        String content =
+                new TemplateResolver(DEFAULT_SHIPKIT_CONFIG_FILE_CONTENT)
+                        .withProperty("gitHub.repository", "mockito/mockito-release-tools-example")
+                        .withProperty("gitHub.writeAuthUser", "shipkit")
```
szczepiq: I was surprised that this had worked. GitHub auth token seems to ignore the name of the user even though you need to specify one when doing git push. Basically you can do ```git push foobar:<auth token>@somewhere master``` and 'foobar' will be happily accepted so long the auth token is right.

Let's see how it goes and if GitHub does not change it.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
```diff
@@ -32,17 +27,21 @@
     private ReleaseConfiguration configuration;
 
     public static final String CONFIG_FILE_RELATIVE_PATH = "gradle/shipkit.gradle";
+    static final String INIT_CONFIG_FILE_TASK = "initConfigFile";
 
-    public void apply(Project project) {
+    public void apply(final Project project) {
         if (project.getParent() == null) {
             //root project, add the extension
+            project.getPlugins().apply(BootstrapPlugin.class);
```
szczepiq: Let's update the Javadoc that we're applying another plugin here

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
```diff
@@ -52,17 +51,30 @@ public void apply(Project project) {
...
     }
 
-    private void loadShipKitConfigFile(Project rootProject) {
-        final File configFile = rootProject.file(CONFIG_FILE_RELATIVE_PATH);
+    private void loadConfigFromFile(Project rootProject, File configFile) {
         if (!configFile.exists()) {
-            createShipKitConfigFile(configFile);
-            throw new GradleException("Config file created at " + configFile.getAbsolutePath() + ". Please configure it and rerun the task.");
+            // sets some defaults so that they can't be used to run any task (except for bootstrap ones)
+            // but also configuration doesn't fail when running Shipkit for the first time
```
szczepiq: I think this is a good idea!

Let's change the repo and auth token to be something dummy, like "dummy repo - please run the init task" and "dummy auth token - please run..."

Do you think it is worth to emit some kind of warning here as well? Something like shipkit configuration file does not exist, please run init tasks

src/main/groovy/org/mockito/release/gradle/BumpVersionFileTask.java
```diff
@@ -38,7 +38,10 @@ public void setVersionFile(File versionFile) {
     /**
      * See {@link BumpVersionFileTask}
      */
-    @TaskAction public void bumpVersionFile() {
+    @TaskAction public VersionInfo bumpVersionFile() {
+        if(!versionFile.exists()){
```
wwilk: Should I remove it? I'm a little confused because you tweaked the thrown exception's message.

src/main/groovy/org/mockito/release/internal/gradle/BintrayPlugin.java
wwilk: Thanks, I forgot about lazyConfiguration!

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
wwilk: Don't you think it's better to keep the exemplary correct values, especially for repository? It tells you immediately what format is the property value ("mockito/mockito" or "https://github.com/mockito/mockito"?). The token provided here is not correct btw.

szczepiq: >Don't you think it's better to keep the exemplary correct values, especially for repository? 

I'm sold :) Let's do it. I really liked your idea to default it based on git origin. We can do it in a separate ticket, though.>Don't you think it's better to keep the exemplary correct values, especially for repository?

src/main/groovy/org/mockito/release/gradle/BumpVersionFileTask.java
szczepiq: >I'm a little confused because you tweaked the thrown exception's message.

Looks like I've been doing mechanical changes with only half of my brain thinking :) Yes, this can be removed because it will never happen.

src/main/groovy/org/mockito/release/internal/gradle/ReleaseConfigurationPlugin.java
wwilk: OK I will create a ticket for that :)

wwilk:
Applied agreed changes and merging

PR #189, opened 2017-05-30
Renamed extension from releasing to shipkit

NagRock:


szczepiq:
Thank you for following up on this, let's get this finished off!

PR #191, opened 2017-05-30
Clean up public and internal API package distinction: move all other …

### LINKED ISSUE #178: Clean up public and internal API package distinction
### Problem

We need clean separation of public and internal API for Shipkit 1.0. We need smaller public API because it helps us develop new API carefully while keeping compatibility.

### Suggested plan

Current package structure:

```
org.mockito.release.exec - process runner
org.mockito.release.gradle - public Gradle tasks, and public Gradle extensions
org.mockito.release.internal - all internal
org.mockito.release.notes - release notes
org.mockito.release.version - versioning
```

Desired package structure - for 1.0 we will only expose Gradle tasks as public types and Gradle extensions. Everything else will be internal because we want to limit the scope of 1.0 public API. Later on, we can start exposing more public APIs or even a library that can be used without Gradle.

```
org.shipkit.gradle - public Gradle tasks, and public Gradle extensions
org.shipkit.internal - all internal
```

### Suggested implementation

 - [x] - rename "org.mockito.release.gradle" into "org.shipkit.gradle"
 - [ ] - move all other packages and classes under "org.shipkit.internal"

Example:

```
org.mockito.release.exec -> org.shipkit.internal.exec
```

epeee:
…packages and classes under "org.shipkit.internal"

fixes #178 

basically, moved all non `org.shipkit.gradle` packages to `org.shipkit.internal` as described in the ticket.


src/test/groovy/org/shipkit/notes/improvements/CommaSeparatedTest.groovy
```diff
@@ -1,15 +0,0 @@
-package org.shipkit.notes.improvements
-
-import spock.lang.Specification
-
-class CommaSeparatedTest extends Specification {
```
szczepiq: I'm wondering if we even need CommaSeparated class given that we can use org.shipkit.internal.gradle.util.StringUtil#join

src/test/groovy/org/shipkit/internal/version/DefaultVersionInfoTest.groovy
```diff
@@ -10,7 +10,7 @@ class DefaultVersionInfoTest extends Specification {
 
     def "does not support files without 'version' property"() {
         def f = dir.newFile() << "asdf"
-        when: org.shipkit.version.DefaultVersionInfo.fromFile(f)
+        when: org.shipkit.internal.version.DefaultVersionInfo.fromFile(f)
```
szczepiq: We should probably use static import here or at least regular import :)

src/test/groovy/org/shipkit/notes/improvements/CommaSeparatedTest.groovy
epeee: thx, I will do some cleanup in another pr soon.

PR #193, opened 2017-06-01
Clean up shipkit gradle plugin names

### LINKED ISSUE #179: Clean up shipkit Gradle plugin names
Before 1.0 release we need to clean up Gradle plugin names and some behavior, too. See [Shipkit Gradle Plugin Reference](https://github.com/mockito/mockito-release-tools/wiki/Shipkit-Gradle-Plugin-Reference).

### Java plugin clean up

- [ ] - rename "org.mockito.release-tools.continuous-delivery" -> "org.shipkit.java"
- [x] - rename "org.mockito.release-tools.java-library" -> "org.shipkit.java-library"
- [ ] - change "org.shipkit.java" plugin so that it automatically applies "org.shipkit.java-library" plugin for all java subprojects. This will be helpful because our clients will not have to apply "org.shipkit.java-library" explicitly. Suggested implementation, in "org.shipkit.java" plugin (pseudo code :):

```groovy
allprojects { subproject -> 
  subproject.plugins.withId("java") { 
    apply plugin: "org.shipkit.java-library" 
  }
}
```
- [ ] - update mockito-release-tools-example project and remove the use of "org.mockito.release-tools.java-library"

epeee:
fixes #179 

- rename `org.shipkit.continuous-delivery` -> `org.shipkit.java`
- automatically apply `org.shipkit.java-library` to all java projects


CONTRIBUTING.md
```diff
@@ -90,9 +90,9 @@ Gradle will throw an exception during configuration time: "task clean cannot be
     The plugin should be referred via the "plugin id", examples:
 
 ```Groovy
-apply plugin: "org.shipkit.continuous-delivery"
+apply plugin: "org.shipkit.java"
```
NagRock: Shouldn't we also change classes names?

src/main/groovy/org/shipkit/internal/gradle/ContinuousDeliveryPlugin.java
```diff
@@ -52,6 +52,18 @@ public void apply(final Project project) {
 
         project.allprojects(new Action<Project>() {
             @Override
+            public void execute(final Project project) {
+                project.getPlugins().withId("java", new Action<Plugin>() {
```
szczepiq: I suggest we move it to the existing 'allprojects' block. This way we have less code to maintain and the execution is a bit faster (no need to iterate projects multiple times).

CONTRIBUTING.md
szczepiq: Absolutely! Let's keep the String identifiers of plugins in sync with the implementing class name. I suggest to call the new class "ShipkitJavaPlugin".

epeee: @NagRock, @szczepiq thx for your feedback!

Issue #196, opened 2017-06-02
Default value for gitHub.repository property based on origin remote url


wwilk:
Command "git remote get-url origin" can be used for setting default value of gitHub.repository property. In case this git call fails sensible default like "mockito/mockito-release-tools-example" should be set instead.

PR #197, opened 2017-06-02
Default value of github.repository property set to git origin remote url #196

wwilk:


wwilk:
I reverted commit accidentally pushed to the wrong branch :)

src/main/groovy/org/shipkit/internal/config/GitOriginRepoProvider.java
```diff
@@ -0,0 +1,32 @@
...
+     * in case "git remote" call fails it returns {#FALLBACK_REPO}
+     */
+    public String getOriginGitRepo(){
+        try {
+            return processRunner.run("git", "remote", "get-url", "origin") // for GitHub returns git@github.com:user/repo.git
+                    .trim()
+                    .replaceFirst("^.*?:", "") // remove everything before ':'
+                    .replace(".git", ""); // remove .git suffix
+        } catch(Exception e){
+            LOG.error("Failed to get url of git remote origin", e);
```
szczepiq: If this log message is an error, what is the user action to fix the error?

src/main/groovy/org/shipkit/internal/config/GitOriginRepoProvider.java
```diff
@@ -0,0 +1,32 @@
+package org.shipkit.internal.config;
+
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.shipkit.internal.exec.ProcessRunner;
+
+public class GitOriginRepoProvider {
```
szczepiq: +1 to new class - makes the code cleaner. Do you think it belongs somewhere in git / vcs packages?

src/main/groovy/org/shipkit/internal/config/GitOriginRepoProvider.java
```diff
@@ -0,0 +1,32 @@
...
+
+    private static final Logger LOG = Logging.getLogger(GitOriginRepoProvider.class);
+    public static final String FALLBACK_REPO = "mockito/mockito-release-tools-example";
```
szczepiq: Nitpick: let's avoid hardcoding the fallback repo in this class. Let's make this class responsible for identifying the origin repo but not to make a decision what to do when that operation fails.

I suggest to pass fallback repo as String to the 'getOriginGitRepo' method. This way, the client decides what repo to use, in case obtaining the repo fails.

How does it sound?

src/test/groovy/org/shipkit/internal/config/GitOriginRepoProviderTest.groovy
```diff
@@ -0,0 +1,31 @@
+package org.shipkit.internal.config
+
+import org.shipkit.internal.exec.ProcessRunner
+import spock.lang.Specification
+
+class GitOriginRepoProviderTest extends Specification {
```
szczepiq: Thank you for adding comprehensive coverage!

src/main/groovy/org/shipkit/internal/config/GitOriginRepoProvider.java
wwilk: Good idea, this class may be reused for a different use case.

wwilk:
I moved error handling to InitConfigFileTask, because in this case since GitOriginRepoProvider is used for generating default values for config file, user should be informed what default value was used instead and how can they change it. Details about the error will be logged below anyway (ProcessRunner logs it).

src/main/groovy/org/shipkit/internal/notes/vcs/GitOriginRepoProvider.java
```diff
@@ -0,0 +1,23 @@
...
+
+    private final ProcessRunner processRunner;
+
+    public GitOriginRepoProvider(ProcessRunner processRunner){
+        this.processRunner = processRunner;
+    }
+
+    /**
+     * fetches remote url for git origin and returns it in format "user/repo", eg. "mockito/mockito-release-tools"
+     * in case "git remote" call fails it returns {#FALLBACK_REPO}
```
mstachniuk: Comment is out dated - `it returns {#FALLBACK_REPO}` is not valid after move

src/main/groovy/org/shipkit/internal/gradle/InitConfigFileTask.java
```diff
@@ -13,6 +17,13 @@
     private static final Logger LOG = Logging.getLogger(InitConfigFileTask.class);
 
     private File configFile;
+    private GitOriginRepoProvider gitOriginRepoProvider;
+    public static final String FALLBACK_GITHUB_REPO = "mockito/mockito-release-tools-example";
+
+    public InitConfigFileTask(){
+        ProcessRunner runner = new DefaultProcessRunner(getProject().getProjectDir());
```
mstachniuk: You don't need `ProcessRunner` here anymore

wwilk: Good catch, thanks @mstachniuk

src/main/groovy/org/shipkit/internal/notes/vcs/GitOriginRepoProvider.java
wwilk: Good catch, thanks @mstachniuk

src/main/groovy/org/shipkit/internal/gradle/InitConfigFileTask.java
wwilk: I actually do, I can pass only projectDir to GitOriginRepoProvider, but it's easier to test GitOriginRepoProvider if I pass ProcessRunner directly.

mstachniuk: Ahhh ok, I missed it in line 25

PR #199, opened 2017-06-03
Tweaking output of publicationsComparator task

wwilk:
I made a few tweaks. Only differences between poms/zips is displayed now, no contents. I added a small dependency on diffutils. I did a research and this one is the smallest one that does the job, in my opinion. XMLUnit would be also nice for comparing poms but is a few times bigger (136 KB vs 36 KB). Diffutils presents changes in a nice familiar "git diff" way. You can see it below. Maybe we can start with it for version 1.0, and later replace it with our own implementation (or some other one) if we need to.

With these changes output of example looks like that:

```
:impl:comparePublications
:impl:comparePublications - about to compare publications, for versions 0.15.2 and 0.15.11
:impl:comparePublications - pom files equal: false
:impl:comparePublications - source jars equal: false
:assertReleaseNeeded
  Environment variable SKIP_RELEASE present: false
  Commit message to inspect for keyword '[ci skip-release]': <unknown commit message>
  Current branch 'master' matches 'master|release/.+': true

  Results of publications comparison:

  Difference between files:
  --- /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/api/build/previous-release-artifacts/api-0.15.2-sources.jar
  +++ /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/api/build/libs/api-0.15.11-sources.jar

    Modified files:
    +- org/mockito/releasetools/example/SomeApi.java



  Difference between files:
  --- /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/api/build/previous-release-artifacts/api-0.15.2.pom
  +++ /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/api/build/publications/javaLibrary/pom-default.xml

    @@ -5,0 +5,8 @@
    +  <dependencies>
    +    <dependency>
    +      <groupId>com.googlecode.java-diff-utils</groupId>
    +      <artifactId>diffutils</artifactId>
    +      <version>1.3.0</version>
    +      <scope>runtime</scope>
    +    </dependency>
    +  </dependencies>
    @@ -8,1 +16,0 @@
    -  <description/>

  Difference between files:
  --- /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/impl/build/previous-release-artifacts/impl-0.15.2-sources.jar
  +++ /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/impl/build/libs/impl-0.15.11-sources.jar

    Added files:
    ++ Test.java

    Removed files:
    -- Library.java



  Difference between files:
  --- /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/impl/build/previous-release-artifacts/impl-0.15.2.pom
  +++ /Users/wojtekwilk/Projects/mockito-release-tools-example/subprojects/impl/build/publications/javaLibrary/pom-default.xml

    @@ -22,1 +22,0 @@
    -  <description/>

  Release is needed: true
    - skip by env variable: false
    - skip by commit message: false
    - is pull request build:  false
    - is releasable branch:  true
    - anything changed in publications since the previous release:  true
```

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/ReleaseNeededTask.java
+++ b/src/main/groovy/org/shipkit/gradle/ReleaseNeededTask.java
@@ -5,6 +5,7 @@ import org.gradle.api.GradleException;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
 import org.gradle.api.tasks.TaskAction;
+import org.shipkit.internal.comparison.diff.Diff;
 import org.shipkit.internal.util.EnvVariables;
 import org.shipkit.internal.comparison.PublicationsComparatorTask;
 import org.shipkit.internal.util.ExposedForTesting;
@@ -126,6 +127,8 @@ public class ReleaseNeededTask extends DefaultTask {
 
         boolean allPublicationsEqual = areAllPublicationsEqual();
 
+       logDifferencesBetweenPublications();
+
         boolean releaseNotNeeded = allPublicationsEqual || skipEnvVariable || skippedByCommitMessage || pullRequest || !releasableBranch;
 
         String message = "  Release is needed: " + !releaseNotNeeded +
@@ -142,6 +145,26 @@ public class ReleaseNeededTask extends DefaultTask {
         }
 
         return !releaseNotNeeded;
+    }
+
+    private void logDifferencesBetweenPublications() {
+        if(publicationsComparators.isEmpty()){
+           return; // no differences to log
+        }
+        LOG.lifecycle("\n  Results of publications comparison:\n");
+        for(PublicationsComparatorTask task : publicationsComparators){
+            for(Diff diff : task.getDifferences()){
+                if(!diff.areFilesEqual()) {
+                    LOG.lifecycle("  Difference between files:\n"
+                                + "  --- {}\n"
+                                + "  +++ {}\n\n{}\n",
+                            diff.getPreviousFile(),
+                            diff.getCurrentFile(),
+                            diff.getDiffOutput());
+                }
+            }
+        }
+
     }
 
     public void addPublicationsComparator(PublicationsComparatorTask task) {
--- a/src/test/groovy/org/shipkit/internal/comparison/ZipComparatorTest.groovy
+++ b/src/test/groovy/org/shipkit/internal/comparison/ZipComparatorTest.groovy
@@ -2,6 +2,7 @@ package org.shipkit.internal.comparison
 
 import org.junit.Rule
 import org.junit.rules.TemporaryFolder
+import org.shipkit.internal.comparison.diff.DirectoryDiffGenerator
 import spock.lang.Specification
 import testutil.ZipMaker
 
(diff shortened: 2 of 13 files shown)


szczepiq:
I've read the description and it sounds good:
 - In order give feedback on the dev experience when running the task I need to try it out first :)
 - I will review the code shortly (unless some other contributor beats me to it)
 - The direction looks good. We want the cleanest possible output so that the developer can quickly figure out why the release is needed (dev UX!)

Feedback:
 - Thank you for describing the justification for using the external library - this really helps approving the PR given that we want to keep the dependencies small.
 - Can you update the description of the PR and use fenced block for the entire output of the build when the task runs? Currently it seems the build log is included in the description with some fenced blocks in the middle and I'm not sure how to read it.

wwilk:
Description updated!

src/test/groovy/org/shipkit/internal/comparison/diff/FileDiffGeneratorTest.groovy
```diff
@@ -0,0 +1,50 @@
...
+    }
+
+    def "handles same files"() {
+        when:
+        def result = new FileDiffGenerator().generateDiff("a", "b", "aa\nbb", "aa\nbb")
+
+        then:
+        result == ""
+    }
+}
```
szczepiq: Nice readability, good coverage, thank you!

src/test/groovy/org/shipkit/internal/comparison/diff/DirectoryDiffGeneratorTest.groovy
```diff
@@ -0,0 +1,35 @@
...
+    Removed files:
+    -- 1.txt
+    -- 5.txt
+
+    Modified files:
+    +- 2.txt
+
+"""
+    }
+}
```
szczepiq: Nice test, all cases covered with good readability.

src/test/groovy/org/shipkit/internal/comparison/ZipComparatorTest.groovy
```diff
@@ -33,8 +34,32 @@ class ZipComparatorTest extends Specification {
...
+        def directoryDiffGenerator = Mock(DirectoryDiffGenerator)
+
+        def zipComparator = new ZipComparator(directoryDiffGenerator)
+        when:
+        zipComparator.areEqual(zip1, zip2)
+
+        then:
+        1 * directoryDiffGenerator.generateDiffOutput(["6.txt"], ["1.txt", "5.txt"], ["2.txt"])
+
+    }
```
szczepiq: Nice, I can understand the impl by reading the test.

src/main/groovy/org/shipkit/gradle/ReleaseNeededTask.java
```diff
@@ -144,6 +147,26 @@ public ReleaseNeededTask setExplosive(boolean explosive) {
         return !releaseNotNeeded;
     }
 
+    private void logDifferencesBetweenPublications() {
+        if(publicationsComparators.isEmpty()){
+           return; // no differences to log
+        }
+        LOG.lifecycle("\n  Results of publications comparison:\n");
+        for(PublicationsComparatorTask task : publicationsComparators){
+            for(Diff diff : task.getDifferences()){
```
szczepiq: At some point let's think how we can make comparator task incremental. For example, it can drop diff file. If the diff file is empty (or not present, it means there are no differences). Making every task incremental will be very powerful (instant builds) and will increase the credibility of Shipkit. Plus, it will be an exemplary implementation of Gradle plugins :D

szczepiq:
I think that "assertReleaseNeeded" produces too much noise during normal build invocation. If I changed a bunch of files, I'm not sure I need to see all it. We could just have one liner saying: "Binaries are different, found different files (Foo.java, Bar.java, and 13 more)."

No action needed at this time. I might create a ticket for that at some point. Beautiful build output -> happy users :)

PR #202, opened 2017-06-04
Expose "version" and "tagPrefix" as task properties

### LINKED ISSUE #201: Expose “version” and “tagPrefix” as task properties
### Why?
Currently, the task which generates the release notes does not declare `version` and `tagPrefix` as input. Therefore it can be harder to use for advanced use cases.
### How?
 - expose “version” and “tagPrefix” properties as task properties marked with `@input`
 - configure default values via release notes plugin.

epeee:
fixes #201 

 - added task properties (marked with `@Input`)
 - set default values in `ReleaseNotesPlugin`

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/IncrementalReleaseNotes.java
+++ b/src/main/groovy/org/shipkit/gradle/IncrementalReleaseNotes.java
@@ -75,7 +75,7 @@ public abstract class IncrementalReleaseNotes extends DefaultTask {
     }
 
     /**
-     * The tag prefix used for release note generation.
+     * See {@link ReleaseConfiguration.Git#getTagPrefix()}
      */
     @Input
     public String getTagPrefix() {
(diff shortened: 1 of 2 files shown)


epeee:
hmm, I just realized that we do have the same properties in `ReleaseNotesFetcherTask`. Therefore we should unify the code and do the configuration for both tasks in `ReleaseNotesPlugin`.

src/main/groovy/org/shipkit/gradle/IncrementalReleaseNotes.java
```diff
@@ -57,6 +59,36 @@ public void setReleaseNotesFile(File releaseNotesFile) {
...
+     */
+    public void setVersion(String version) {
+        this.version = version;
+    }
+
+    /**
+     * The tag prefix used for release note generation.
+     */
+    @Input
+    public String getTagPrefix() {
```
szczepiq: Let's use Javadoc link to ```{@link ReleaseConfiguration.Git#getTagPrefix()}``` so that we don't have to duplicate the documentation.

src/main/groovy/org/shipkit/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -80,6 +80,9 @@ private static void configureDetailedNotes(final IncrementalReleaseNotes task,
         task.dependsOn(releaseNotesFetcher);
         task.dependsOn(contributorsFetcher);
 
+        task.setVersion(project.getVersion().toString());
+        task.setTagPrefix("v");
```
szczepiq: Let's grab it from ```conf.getGit().getTagPrefix()```

PR #204, opened 2017-06-05
Make applying java-library plugin conditional on a property

### LINKED ISSUE #200: Make applying java-library plugin conditional on a property
Currently we apply java-library on each subproject with java plugin already applied. It should be possible to disable this mechanism. As suggested by @szczepiq we can add a property "shipkit.releaseAllJavaModules" to ReleaseConfiguration. By default this property should be true and in this case java-library would be applied automatically.

epeee:
fixes #200 

introduced `shipkit.releaseAllJavaModules` (which defaults to `true`). If the property mentioned is `true` we'll apply `java-library` to every subproject (which does have the `java` plugin applied).

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
@@ -78,8 +78,8 @@ public class ReleaseConfiguration {
     }
 
     /**
-     * org.shipkit.java-library plugin will be applied to every java subproject if this boolean
-     * is <code>true</code>.
+     * org.shipkit.java-library plugin will be applied to every java subproject (project that applies Gradle's 'java'
+     * plugin) if this boolean is <code>true</code>.
      */
     public boolean isReleaseAllJavaModules() {
         return releaseAllJavaModules;
--- a/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
@@ -52,6 +52,7 @@ public class ReleaseConfiguration {
     //TODO currently it's not clear when to use class fields and when to use the 'configuration' map
     //Let's make it clear in the docs
     private boolean dryRun = true;
+    private boolean releaseAllJavaModules = true;
 
     /**
      * See {@link #isDryRun()}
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -69,6 +70,21 @@ public boolean isDryRun() {
...
 
+    /**
+     * See {@link #isReleaseAllJavaModules()}}
+     */
+    public void setReleaseAllJavaModules(boolean releaseAllJavaModules) {
+        this.releaseAllJavaModules = releaseAllJavaModules;
+    }
+
+    /**
+     * org.shipkit.java-library plugin will be applied to every java subproject if this boolean
```
szczepiq: Nice docs! Let's make it explicit what 'java subproject' means, something like that:

"org.shipkit.java-library plugin will be applied to every java subproject (project that applies Gradle's 'java' plugin) if this ..."

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -52,6 +52,7 @@ public ReleaseConfiguration() {
     //TODO currently it's not clear when to use class fields and when to use the 'configuration' map
     //Let's make it clear in the docs
     private boolean dryRun = true;
+    private boolean releaseAllJavaModules = true;
```
szczepiq: Team, are you ok with this name of the property? It's a public API so let's choose the best name :) Some alternatives:

- shipkit.releaseAllJavaModules
- shipkit.releaseAllJavaSubprojects
 - shipkit.publishAllJava
 - shipkit.publishEveryJavaSubproject
 - shipkit.shipEveryJavaModule
 - shipkit.java.shipAll

src/main/groovy/org/shipkit/internal/gradle/ShipkitJavaPlugin.java
```diff
@@ -53,12 +53,14 @@ public void apply(final Project project) {
         project.allprojects(new Action<Project>() {
             @Override
             public void execute(final Project subproject) {
-                subproject.getPlugins().withId("java", new Action<Plugin>() {
-                    @Override
-                    public void execute(Plugin plugin) {
-                        subproject.getPlugins().apply(JavaLibraryPlugin.class);
-                    }
-                });
+                if (conf.isReleaseAllJavaModules()) {
```
szczepiq: Perfect!

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
wwilk: Fine with me, although shipkit.publishAllJavaSubprojects may be a little more intuitive, since "subproject" is a gradle term and there is a well-known "maven-publish" plugin.

epeee: thx, let's go for `shipkit.publishAllJavaSubprojects`.

Issue #205, opened 2017-06-05
Combine updateReleaseNotes and previewReleaseNotes into one task with parameter


wwilk:
The release generation task is implemented as a pair of tasks: preview task and separate update task. This approach makes the tasks hard to reuse. If someone wants to declare or reconfigure release notes generation at the task level, he needs to do it twice. You can see this problem in #190 - “emphasizeVersion” property needs to configured twice. To solve the problem, we can create single implementation of IncrementalReleaseNotes and use Gradle’s task parameter API to distinguish between preview mode and update mode. Here’s how would this work:

update release notes: ./gradlew updateReleaseNotes
preview release notes: ./gradlew updateReleaseNotes -Ppreview

PR #206, opened 2017-06-05
PreviewReleaseNotes and updateReleaseNotes merged into one task #205

wwilk:
As discussed, both tasks replaced with one, preview can be displayed with parameter -Ppreview. 

I tested some of the functionality in IncrementalReleaseNotes and extracted rest of it into a separate local class, as the first step to improve the coverage. Next step would be to extract this class into a top level one and write tests for it.

@szczepiq 
Important - I removed @InputFile from IncrementalReleaseNotes/UpdateReleaseNotesTask#48. It didn't make sense there for me - releaseNotesFile is already @OutputFile (only NOT in preview mode) so it was always updated when the task was run and this task would never be UP-TO-DATE.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,11 +22,11 @@ It should use the same version that was built in the previous step.
 3. Basic testing (for most contributors):
  - Smoke test (no tasks are run): ```./gradlew testRelease -m```
  - Test most things, without actually making any code pushes/publications: ```./gradlew testRelease -x gitPush -x bintrayUpload```
- - Release notes content: ```./gradlew previewReleaseNotes```
-    To generate sizable release notes content, before running 'previewReleaseNotes' you can downgrade the 'previousVersion' in 'version.properties'.
+ - Release notes content: ```./gradlew updateReleaseNotes -Ppreview```
+    To generate sizable release notes content, before running 'updateReleaseNotes' you can downgrade the 'previousVersion' in 'version.properties'.
     Release notes are generated from 'previousVersion' to current 'version' as declared in 'version.properties' file.
 4. Advanced testing (occasionally, for core developers, edge cases):
- - Release notes in file: ```./gradlew previewReleaseNotes```, then inspect updated file
+ - Release notes in file: ```./gradlew updateReleaseNotes```, then inspect updated file
  - Test release needed task: ```./gradlew assertReleaseNeeded```
  - If you are one of the core developers you can export env variables and even test git push and bintray upload.
  Run ```./gradlew testRelease``` follow the prompts and export necessary env variables.
--- a/src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
+++ b/src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
@@ -44,6 +44,8 @@ public class UpdateReleaseNotesTask extends DefaultTask {
     private Collection<String> contributors;
     private File contributorsDataFile;
     private boolean emphasizeVersion;
+    private boolean previewMode;
+
     private IncrementalNotesGenerator incrementalNotesGenerator = new IncrementalNotesGenerator();
 
     /**
@@ -53,7 +55,7 @@ public class UpdateReleaseNotesTask extends DefaultTask {
     @TaskAction
     public void updateReleaseNotes() {
         String newContent = getNewContent();
-        if (isInPreviewMode()){
+        if (previewMode){
             LOG.lifecycle("  Preview of release notes update:\n" +
                     "  ----------------\n" + newContent + "----------------");
         } else{
@@ -63,11 +65,19 @@ public class UpdateReleaseNotesTask extends DefaultTask {
     }
 
     /**
-     * @return true if task is configured to generate only preview of release notes, and false otherwise
+     * @return true if task is configured to generate only preview of release notes (without appending them to the file), and false otherwise
      */
-    public boolean isInPreviewMode() {
-        return getProject().hasProperty(PREVIEW_PROJECT_PROPERTY);
+    public boolean isPreviewMode() {
+        return previewMode;
     }
+
+    /**
+     * See {@link #isPreviewMode()} ()}
+     */
+    public void setPreviewMode(boolean previewMode) {
+        this.previewMode = previewMode;
+    }
+
 
     /**
      * Release notes file this task operates on.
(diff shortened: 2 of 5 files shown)


szczepiq:
>@szczepiq
Important - I removed @inputfile from IncrementalReleaseNotes/UpdateReleaseNotesTask#48. It didn't make sense there for me - releaseNotesFile is already @outputfile (only NOT in preview mode) so it was always updated when the task was run and this task would never be UP-TO-DATE.

Sounds good! Thank you. Doing review now...

CONTRIBUTING.md
```diff
@@ -22,11 +22,11 @@ It should use the same version that was built in the previous step.
 3. Basic testing (for most contributors):
  - Smoke test (no tasks are run): ```./gradlew testRelease -m```
  - Test most things, without actually making any code pushes/publications: ```./gradlew testRelease -x gitPush -x bintrayUpload```
- - Release notes content: ```./gradlew previewReleaseNotes```
-    To generate sizable release notes content, before running 'previewReleaseNotes' you can downgrade the 'previousVersion' in 'version.properties'.
+ - Release notes content: ```./gradlew updateReleaseNotes -Ppreview```
```
szczepiq: Thank you for not forgetting to update the documentation!

src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
```diff
@@ -19,17 +19,20 @@
...
+import java.io.IOException;
 import java.util.*;
 
 /**
  * Generates incremental, detailed release notes text.
- * that can be appended to the release notes file.
+ * that can be appended to the release notes file or displayed in console as a preview.
  */
-public abstract class IncrementalReleaseNotes extends DefaultTask {
+public class UpdateReleaseNotesTask extends DefaultTask {
```
szczepiq: +1, much better name

src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
```diff
@@ -199,46 +225,40 @@ public boolean isEmphasizeVersion() {
...
+            }
+            if(releaseNotesFile.exists() && !releaseNotesFile.isFile()){
+                throw new GradleException("'" + this.getPath() + ".releaseNotesFile' must be a file.");
+            }
+            if(!releaseNotesFile.exists()){
+                IOUtil.createParentDirectory(releaseNotesFile);
+                try {
+                    releaseNotesFile.createNewFile();
+                } catch (IOException e) {
+                    e.printStackTrace();
```
szczepiq: Let's not forget to handle it with a throw.

src/main/groovy/org/shipkit/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -19,9 +19,7 @@
  *
  * <ul>
  *     <li>fetchReleaseNotes - fetches release notes data, see {@link ReleaseNotesFetcherTask}</li>
- *     <li>updateReleaseNotes - updates release notes file in place, see {@link IncrementalReleaseNotes.UpdateTask}</li>
- *     <li>previewReleaseNotes - prints incremental release notes to the console for preview,
- *          see {@link IncrementalReleaseNotes.PreviewTask}</li>
+ *     <li>updateReleaseNotes - updates release notes file in place, or only displays preview, see {@link UpdateReleaseNotesTask}</li>
```
szczepiq: How about we document the project property here so that we have comprehensive docs.

src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
```diff
@@ -41,11 +44,34 @@
...
+        } else{
+            FileUtil.appendToTop(newContent, getReleaseNotesFile());
+            LOG.lifecycle("  Successfully updated release notes!");
+        }
+    }
+
+    /**
+     * @return true if task is configured to generate only preview of release notes, and false otherwise
+     */
+    public boolean isInPreviewMode() {
```
szczepiq: Instead of inspecting the project property on the fly I suggest to create a boolean property on the task "preview" or "previewMode". This is more natural in Gradle and also makes the task more flexible for folks who want to use the task but not necessarily configure it via project property. The plugin would configure the boolean property based on the project property.

src/main/groovy/org/shipkit/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -55,24 +52,23 @@ public void execute(final ReleaseNotesFetcherTask t) {
...
-                configureDetailedNotes(t, releaseNotesFetcher, project, conf, contributorsFetcher);
-                GitPlugin.registerChangesForCommitIfApplied(
-                        Arrays.asList(project.file(conf.getReleaseNotes().getFile())), "release notes updated", t);
-            }
-        });
+        TaskMaker.task(project, UPDATE_NOTES_TASK, UpdateReleaseNotesTask.class, new Action<UpdateReleaseNotesTask>() {
+            public void execute(final UpdateReleaseNotesTask t) {
+                t.setDescription("Updates release notes file. Run with '-Ppreview' if you only want to see the preview.");
 
-        TaskMaker.task(project, PREVIEW_NOTES_TASK, IncrementalReleaseNotes.PreviewTask.class, new Action<IncrementalReleaseNotes.PreviewTask>() {
```
szczepiq: Nice! Dual tasks were not very clean solution.

src/main/groovy/org/shipkit/internal/gradle/ReleaseNotesPlugin.java
```diff
@@ -55,24 +52,23 @@ public void execute(final ReleaseNotesFetcherTask t) {
...
-            public void execute(final IncrementalReleaseNotes.PreviewTask t) {
-                t.setDescription("Shows new incremental content of release notes. Useful for previewing the release notes.");
                 configureDetailedNotes(t, releaseNotesFetcher, project, conf, contributorsFetcher);
+
+                if(!t.isInPreviewMode()){
+                    File releaseNotesFile = project.file(conf.getReleaseNotes().getFile());
+                    GitPlugin.registerChangesForCommitIfApplied(
+                        Arrays.asList(releaseNotesFile), "release notes updated", t);
+                    t.getOutputs().file(releaseNotesFile);
```
szczepiq: This is super smart!!! I forgot that we can configure the outputs programatically.

I suggest to make the conditional logic simpler. In the plugin we could do something like:

```
if (project.hasProperty("preview")) {
   t.setPreview(true);
   t.getOutputs().file(t.getReleaseNotesFile());
}
```

src/main/groovy/org/shipkit/gradle/UpdateReleaseNotesTask.java
wwilk: Yeah, and easier for testing too, I will do that, thanks!

wwilk:

PR #207, opened 2017-06-05
Maven coordinates renamed to shipkit

wwilk:


wwilk:
Merging to check if everything is published correctly.

PR #208, opened 2017-06-05
Remaining mockito references removed

wwilk:


wwilk:
Merging to see if it is deployed correctly to Bintray

Issue #61, opened 2017-04-03
Support for GitHub Enterprise


mstachniuk:
### Problem

Currently mockito-release-tools works only with github.com. A path to it is hardcoded in the code :(. This project needs to works with [GitHub Enterprise API](https://developer.github.com/v3/enterprise/) otherwise it cannot be used in enterprise setting. A property to configure GitHub Enterprise needs to optional and by default it should works with github.com.

### Suggested implementation

ReleaseConfiguration.GitHub should have an optional setting for the url base host:

```Gradle
releasing {
    github.apiUrl = "https://github.in.your.company.com/api/v3"
    github.url = "https://github.in.your.company.com"
}
```

- [ReleaseConfiguration.GitHub](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/gradle/ReleaseConfiguration.java#L148) class that needs new properties.
- Example of [hardcoded GitHub API url](https://github.com/mockito/mockito-release-tools/blob/master/src/main/groovy/org/mockito/release/notes/improvements/GitHubTicketFetcher.java#L153), there are other places but this one is a good starting point.

szczepiq:
@mstachniuk, can you confirm suggested implementation above?

mstachniuk:
I suggest to add 2 options to the configuration:
```
releasing {
    github.api = "https://github.in.your.company.com/api/v3"
    github.url = "https://github.in.your.company.com"
    // others ...
}
```

The first one to the API endpoint for generation API requests and second one for generating links for  developers in pom (see: PomCustomizer.groovy)

szczepiq:
Nice proposal! I updated "api" -> "apiUrl". Let's make sure they are defaulted so that we don't have to configure it for typical OS builds.

PR #217, opened 2017-06-09
Add support for GitHub Enterprise

slq:
Implementation for #61.

Added support for GitHub Enterprise with default parameter values
```
gitHub.url = "https://github.in.your.company.com"
gitHub.apiUrl = "https://github.in.your.company.com/api/v3"
```

`gitHub.url` is used for release notes generation mostly. We need to keep it separated from `gitHub.apiUrl` as these two might have more differences than just relative path `/api/v3`. 

Once this pull request is merged, I will update shipkit-examples


gradle/shipkit.gradle
```diff
@@ -1,4 +1,6 @@
 shipkit {
+    // TODO uncomment once version is updated
+//    gitHub.apiUrl = "https://api.github.com/"
```
slq: I will uncomment this property when PR is merged

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -119,6 +121,22 @@ public String getPreviousReleaseVersion() {
...
+         */
+        public String getApiUrl() {
+            return getString("gitHub.apiUrl");
+        }
+
+        /**
+         * See {@link #getApiUrl()}
+         *
+         * @param apiUrl GitHub (or gitHub enterprise) API endpoint address, for example: https://api.github.com/
+         */
```
wwilk: This "/" at the end of example URL is important. In GitHubProjectContributorsBuilder you concatenate "/repos" to apiUrl so it has to not end with slash. 
See:
https://api.github.com//repos/mockito/mockito/contributors?access_token=e7fe8fcdd6ffed5c38498c4c79b2a68e6f6ed1bb

This one doesn't work. 

At first we can include information that URL should not be ended with slash, but it would be good to have a code that handles URL with slash and without it.

src/main/groovy/org/shipkit/internal/notes/contributors/Contributors.java
```diff
@@ -10,11 +10,12 @@
     /**
      * Fetches contributors from GitHub. Needs GitHub auth token.
      *
+     * @param apiUrl address of GitHub api endpoint, for example: "https://api.github.com/"
      * @param repository name of GitHub repository, for example: "mockito/mockito"
```
wwilk: As in the comment above - if we set URL without slash at the end, we shouldn't suggest to do so in javadoc.

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
slq: Thanks, I missed that. In code I expect url without ending slash, but in javadoc, I have just forgotten to remove it.

src/main/groovy/org/shipkit/internal/notes/improvements/GitHubTicketFetcher.java
```diff
@@ -151,7 +153,7 @@ public GitHubIssuesBuilder labels(String labels) {
             GitHubIssues browse() {
                 // see API doc: https://developer.github.com/v3/issues/
                 String nextPageUrl = String.format("%s%s%s%s%s%s%s",
-                        "https://api.github.com/repos/" + repository + "/issues",
+                        apiUrl + "/repos/" + repository + "/issues",
```
wwilk: I don't like the usage of String.format here, it doesn't give us anything when it only contains variables. Would be better to have a simple String concatenation and any more complex logic could be extracted into private methods, like baseUrl(), state(), filter(), direction().

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
slq: And indeed, it might be problematic, if someone used url with ending slash

gradle/shipkit.gradle
wwilk: Thanks!

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
wwilk: It would be good to resolve the endpoint URL, instead of a simple concatenation. Either by using URL class in Java (not sure if it would work in this case, I didn't test it), or a simple custom class.

wwilk: Maybe a class for all GitHub URLs. It may be useful, since some parameters are repeated, like access_token or per_page

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -111,9 +111,9 @@ public void execute(final Task t) {
 
         TaskMaker.execTask(project, COMMIT_CLEANUP_TASK, new Action<Exec>() {
             public void execute(final Exec t) {
-                t.setDescription("Removes last commit, using 'reset --hard HEAD~'");
+                t.setDescription("Removes last commit, using 'reset --soft HEAD~'");
```
wwilk: Why this change? TODO below suggests a combination of git reset --soft HEAD~ and git stash.

slq: This wasn't supposed to go with this change.

szczepiq: I feel you ;) Happened to me multiple times! We should go ahead an implement the TODO, ```git reset --hard``` is pretty brutal.

src/main/groovy/org/shipkit/internal/gradle/NotesGeneration.java
```diff
@@ -9,6 +9,7 @@
 //TODO expose as public API
 public class NotesGeneration {
```
szczepiq: NotesGeneration along with NotableReleaseNotesFetcherTask is dead code. Can you get rid of it? This will help keeping the codebase clean! Please also remove other stuff connected to those 2 classes (if you find any :)

src/test/groovy/org/shipkit/internal/notes/contributors/RecentContributorsFetcherTest.groovy
```diff
@@ -10,25 +10,26 @@ class RecentContributorsFetcherTest extends Specification {
...
-        def c = fetcher.fetchContributors("mockito/mockito", readOnlyToken,
+        def c = fetcher.fetchContributors(defaultGitHubApiEndpoint, "mockito/mockito", readOnlyToken,
                 parseDate("2017-05-02 00:00:00 -0000"), parseDate("2017-05-03 00:00:00 -0000"))
         expect:
         c.empty
     }
 
     def "null until date smoke test"() {
-        def c = fetcher.fetchContributorsSinceYesterday("mockito/mockito-release-tools", readOnlyToken)
+        def c = fetcher.fetchContributorsSinceYesterday(defaultGitHubApiEndpoint, "mockito/mockito-release-tools", readOnlyToken)
```
szczepiq: Thank you for updating the tests! 'mockito-release-tools' should not be used any more. Shipkit is the new name!

gradle/shipkit.gradle
szczepiq: Since it has a default value, why do you want to uncomment it?

mstachniuk:
Please don't forget about `PomCustomizer`

szczepiq:
@slq, is it ready to merge according to you? I was not following up with changes requested by other team members. If it is done let us know so that we can merge.

src/main/groovy/org/shipkit/internal/notes/improvements/GitHubTicketFetcher.java
slq: Replaced `String.format` with simple builder. It resembles URL nature a little bit better. 
I didn't want to rewrite the logic from scratch here, it might be better to do this in subsequent pull request.

slq:
@szczepiq, yes I think it's ready.

 I believe I have applied all suggestions. It should be better now. Please review and let me know if it's accepted.

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -119,6 +122,36 @@ public String getPreviousReleaseVersion() {
...
+         * GitHub API endpoint address, for example:  https://api.github.com
+         */
+        public String getApiUrl() {
+            return getStringUrl("gitHub.apiUrl");
+        }
+
+        /**
+         * See {@link #getApiUrl()}
+         *
+         * @param apiUrl GitHub (or gitHub enterprise) API endpoint address, for example: https://api.github.com
```
mstachniuk: You can remove this line with `@param` because you link above to getter.

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -119,6 +122,36 @@ public String getPreviousReleaseVersion() {
...
+         */
+        public void setUrl(String url) {
+            configuration.put("gitHub.url", url);
+        }
+
+        /**
+         * GitHub API endpoint address, for example:  https://api.github.com
+         */
+        public String getApiUrl() {
+            return getStringUrl("gitHub.apiUrl");
```
mstachniuk: Good point to be sure that url will always end without slash!

src/main/groovy/org/shipkit/internal/gradle/E2ETestingPlugin.java
```diff
@@ -24,8 +25,10 @@
 
     public void apply(final Project project) {
         E2ETest e2eTest = project.getExtensions().create("e2eTest", E2ETest.class, project);
+        final ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
+
         // TODO hardcoded for now
-        e2eTest.create("https://github.com/mockito/shipkit-example");
+        e2eTest.create(conf.getGitHub().getUrl() + "/mockito/shipkit-example");
```
mstachniuk: Major. Here should be `https://github.com`, please revert changes in this file.

src/main/groovy/org/shipkit/internal/notes/improvements/GitHubTicketFetcher.java
```diff
@@ -107,35 +107,36 @@ boolean hasNextPage() {
...
-            GitHubIssuesBuilder(String repository, String readOnlyAuthToken) {
+            private Map<String, String> parameters;
+
+            GitHubIssuesBuilder(String apiUrl, String repository, String readOnlyAuthToken) {
+                this.apiUrl = apiUrl;
                 this.repository = repository;
-                this.readOnlyAuthToken = readOnlyAuthToken;
+
+                parameters = new HashMap<String, String>();
+                parameters.put("access_token", readOnlyAuthToken);
```
mstachniuk: Great improvement!

slq:
@mstachniuk, thanks, I reverted this unwanted change.

PR #218, opened 2017-06-10
speed up fetchAllContributors task

epeee:
for real world projects like mockito.

Currently, `fetchAllContributors` task takes ~22s while running `gradlew build` in mockito project.

The changes in this PR will end up in a duration of ~7s for the same task executed on the same project.

Note: this is still a very first version which might need some improvements (e.g the splitting up into sublists can be done smarter and also the error handling should be enhanced if we decide to merge the pr).

Actual profiling results:

 - using the currently released version:
`:fetchAllContributors 	22.492s` 	
 - using the version from this PR:
`:fetchAllContributors 	7.665s` 	

Furthermore, I did a diff on the resulting json files and it turned out that they are identical.


src/main/groovy/org/shipkit/internal/notes/contributors/AllContributorsFetcher.java
```diff
@@ -40,16 +43,36 @@ ProjectContributorsSet fetchAllContributorsForProject(String repository, String
...
+                Callable<Set<ProjectContributor>> callable = new FetcherCallable<JsonObject, ProjectContributor>(subList, projectContributorFetcherFunction);
+
+                futures.add(executor.submit(callable));
+            }
         }
+
+        for (Future<Set<ProjectContributor>> future: futures) {
+            try {
+                result.addAll(future.get());
+            } catch (InterruptedException e) {
```
szczepiq: For simplicity, I suggest single catch(Exception) followed by throwing a RuntimeException with a decent message

src/main/groovy/org/shipkit/internal/notes/contributors/AllContributorsFetcher.java
```diff
@@ -40,16 +43,36 @@ ProjectContributorsSet fetchAllContributorsForProject(String repository, String
...
-        for (JsonObject contributor : page) {
-            //Since returned contributor does not have 'name' element, we need to fetch the user data to get his name
-            //TODO add static caching of this. Names don't change that often, let's just cache this forever in build cache.
-            String url = (String) contributor.get("url");
-            GitHubObjectFetcher userFetcher = new GitHubObjectFetcher(url, readOnlyAuthToken);
-            JsonObject user = userFetcher.getPage();
-            result.add(GitHubAllContributorsJson.toContributor(contributor, user));
+        //Since returned contributor does not have 'name' element, we need to fetch the user data to get his name
+        //TODO add static caching of this. Names don't change that often, let's just cache this forever in build cache.
+        ExecutorService executor = Executors.newFixedThreadPool(4);
```
szczepiq: It uses Java7 features. At the moment we still want to support Java6 for old projects that for various reasons cannot modernize Java version. How much complexity we need to add to support this solution if we remained Java6 compatible? We can revisit the Java6 compatibility requirement or decide to user Java7 as minimum required version to run Shipkit.

src/main/groovy/org/shipkit/internal/notes/contributors/ProjectContributorFetcherFunction.java
```diff
@@ -0,0 +1,37 @@
...
+        this.objectFetcher = objectFetcher;
+    }
+
+    @Override
+    public ProjectContributor apply(JsonObject contributor) {
+        String url = (String) contributor.get("url");
+        JsonObject user;
+        try {
+            user = objectFetcher.getPage(url);
+        } catch (IOException e) {
```
szczepiq: Single catch clause with 'Exception' gives us simpler code and less duplication.

src/main/groovy/org/shipkit/internal/notes/contributors/AllContributorsFetcher.java
```diff
@@ -40,16 +43,36 @@ ProjectContributorsSet fetchAllContributorsForProject(String repository, String
...
+        int maxSizePerCallable = 25;
+
+        List<Future<Set<ProjectContributor>>> futures = new ArrayList<Future<Set<ProjectContributor>>>();
+        if (page.size() > 0) {
+            for (int i = 0; i < Math.max(page.size() / maxSizePerCallable, 1); i++) {
```
szczepiq: I think we can come up with a simpler technical solution, with less math for paging.

High level guidelines:
 - decouple concurrency handling as much as possible from AllContributorsFetcher
 - decouple concurrency in a way it is unit testable without hitting GitHub

Suggested approach:
1. In AllContributorsFetcher 'page' object is a list of contributors 
2. We kick off threads, each thread: 
 - removes contributor from page
 - hits GitHub for more data for this contributor
 - converts data to the user
 - puts user into shared result list
3. we need to use locks or collections that are thread safe

Feel free to implement a solution as you see fit. Above is a suggestion :)

szczepiq:
Ping us when you finish so that we can merge. Let's make this fast :)

epeee:
@szczepiq sure, will do so. Most likely tomorrow.

src/main/groovy/org/shipkit/internal/notes/contributors/FetcherCallable.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+import java.util.concurrent.Callable;
+
+/**
+ * Basic function which applies values from a {@link List} of {@link V} to a given {@link Function}.
+ */
+class FetcherCallable<V, R> implements Callable<Set<R>> {
```
szczepiq: This class is neatly abstracted out and should very easy to unit test, too. Nice!

Since it contains conditional logic and some control statements like looping, can you write a unit test for it?

src/main/groovy/org/shipkit/internal/notes/contributors/ContributorsDispatcher.java
```diff
@@ -0,0 +1,45 @@
...
+
+    public Set<ProjectContributor> dispatch(List<JsonObject> page, String readOnlyAuthToken) {
+        Set<ProjectContributor> result = new HashSet<ProjectContributor>();
+        ExecutorService executor = Executors.newFixedThreadPool(N_THREADS);
+        List<JsonObject> synchronizedPageList = Collections.synchronizedList(page);
+
+        List<Future<Set<ProjectContributor>>> futures = new ArrayList<Future<Set<ProjectContributor>>>();
+        if (page.size() > 0) {
+            for (int i = 0; i < N_THREADS; i++) {
+                GitHubObjectFetcher objectFetcher = new GitHubObjectFetcher(readOnlyAuthToken);
```
szczepiq: This method has sufficient meat that is worth unit testing :)

I was thinking how to make it neatly unit testable (btw. my favorite thing about engineering is figure out how I can cut code so that it is easy to test :). fetcher and fetcherFunction are immutable so you can pass fetcherFunction as parameter to "dispatch" method. This should make "dispatch" method unit testable easily.

src/test/groovy/org/shipkit/internal/notes/contributors/ProjectContributorFetcherFunctionTest.groovy
```diff
@@ -0,0 +1,28 @@
+package org.shipkit.internal.notes.contributors
+
+import org.json.simple.JsonObject
+import org.shipkit.internal.notes.model.ProjectContributor
+import org.shipkit.internal.notes.util.GitHubObjectFetcher
+import spock.lang.Specification
+
+
+class ProjectContributorFetcherFunctionTest extends Specification {
```
szczepiq: Very nice test - small, clear, comprehensive, thank you!

src/main/groovy/org/shipkit/internal/notes/contributors/FetcherCallable.java
epeee: thx, just added some tests.

PR #224, opened 2017-06-10
added animalsniffer plugin (Java6 compatibility validation)

### LINKED ISSUE #222: Automated Java6 compatibility validation
### Problem

At the moment Shipkit is compatible with Java6 and we want 1.0 to be runnable on Java6. To ensure that we meet this requirement we should add validation for it.

### Suggested solution

- add matrix build in Travis configuration and also run build on Java6
- use animal sniffer, in similar way Mockito does it

Using Java6 for builds on Travis CI is problematic. In Mockito project we encountered weird errors that were gone when we upgraded to Java7. I suspect that animal sniffer is a better option.

For reference how the solution was implemented in Mockito, see https://github.com/mockito/mockito/pull/1021

epeee:
fixes #222 

Currently, animalsniffer plugin is always applied (in mockito project it is just applied if a certain property is passed or if release is skipped).
What do you think about executing animalsniffer plugin with every build.
If it does not take a lot if time I'd prefer to execute it with every build because of fail fast.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -7,6 +7,7 @@ buildscript {
     dependencies {
         classpath "com.gradle.publish:plugin-publish-plugin:0.9.6"
         classpath "gradle.plugin.org.shipkit:shipkit:0.8.82"
+        classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.3.0'
     }
 }
 
@@ -19,6 +20,7 @@ apply plugin: 'codenarc'
 apply plugin: 'org.shipkit.gradle-plugin-releasing'
 apply plugin: 'org.shipkit.release-needed'
 
+apply from: 'gradle/java6-compatibility.gradle'
 apply from: 'gradle/precommit.gradle'
 
 sourceCompatibility = 1.6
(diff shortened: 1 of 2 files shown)


szczepiq:
Have you tried using Java7 API and running the build? I want to make sure it actually fleshes out the problems.

I think it's fine to include it in every build.

Nice! Supporting Java6 in 1.0 will make our tool easier to use for some teams / projects.

szczepiq:
I've looked at Travis CI console and it seems the sniffer does not run as part of the build. Let's hook it up and merge!

epeee:
yes, I have tried it locally using `java.nio.file.Paths#get`:

> [Undefined reference] org.shipkit.gradle.(Test.java:15)
>   \>\> java.nio.file.Path java.nio.file.Paths.get(String, String[])
> 
> :animalsnifferMain FAILED

Ad Travis CI, thx, good catch! I will have a look at it.

epeee:
ad Travis CI: Actually, I the new task gets executed (just had a look at the travis CI output from the pr build):

> :animalsnifferMain
> file or directory '/home/travis/build/mockito/shipkit/src/main/java', not found
> Executing task ':animalsnifferMain' (up-to-date check took 0.043 secs) due to:
>   No history is available.
> :animalsnifferMain (Thread[Daemon worker,5,main]) completed. Took 9.537 secs.

Issue #225, opened 2017-06-10
initShipkit produces incorrect github link


szczepiq:
Running ```./gradlew initShipkit``` with no 'gradle/shipkit.gradle' file produces incorrect github link:

```
gitHub.repository = "//github.com/mockito/shipkit-bootstrap"
```

PR #226, opened 2017-06-10
Fixed getting git origin repo for https #225

wwilk:


src/main/groovy/org/shipkit/internal/notes/vcs/GitOriginRepoProvider.java
```diff
@@ -14,9 +14,17 @@ public GitOriginRepoProvider(ProcessRunner processRunner){
...
-        return processRunner.run("git", "remote", "get-url", "origin") // for GitHub returns git@github.com:user/repo.git
-                    .trim()
+        // for GitHub returns git@github.com:user/repo.git or https://github.com/user/repo.git
+        String remote = processRunner.run("git", "remote", "get-url", "origin").trim();
+
+        if(remote.startsWith("git")){
+            return remote
                     .replaceFirst("^.*?:", "") // remove everything before ':'
                     .replaceFirst(".git$", ""); // remove .git suffix
+        } else{
```
szczepiq: I was thinking if we should throw an exception when output for process is unfamiliar. However, I think it's easier and safer if we don't throw anything. Just like in my case, I could just tweak the repo by hand.

PR #229, opened 2017-06-11
Avoid git reset --hard when cleaning up commits

### LINKED ISSUE #221: Avoid git reset --hard when cleaning up commits
### Problem

Currently we issue ["git reset --hard"](https://github.com/mockito/shipkit/blob/3f90104d0740604cfbcd318896fe6f2908b94747/src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java#L116) to clean up commits. This is pretty brutal and may potentially cause losing commits.

### Suggested solution

Instead of single invocation of "git reset --hard", we can issue 2 git commands:

```
git reset --soft
git stash
```

1. gitCommitCleanUp task performs 'git stash'
2. gitResetCommit task performs 'git reset --soft HEAD~'
3. gitCommitCleanUp.dependsOn gitResetCommit

epeee:
fixes #221 

introduced `gitResetCommit` task, using git reset --soft now + git stash now

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
@@ -28,12 +28,14 @@ import static org.shipkit.internal.gradle.util.GitUtil.getTag;
  *     <li>performGitPush</li>
  *
  *     <li>gitCommitCleanUp</li>
+ *     <li>gitResetCommit</li>
  *     <li>gitTagCleanUp</li>
  * </ul>
  */
 public class GitPlugin implements Plugin<Project> {
 
     static final String COMMIT_CLEANUP_TASK = "gitCommitCleanUp";
+    static final String RESET_COMMIT_TASK = "gitResetCommit";
     static final String TAG_CLEANUP_TASK = "gitTagCleanUp";
     static final String GIT_TAG_TASK = "gitTag";
     static final String GIT_PUSH_TASK = "gitPush";
@@ -111,9 +113,16 @@ public class GitPlugin implements Plugin<Project> {
 
         TaskMaker.execTask(project, COMMIT_CLEANUP_TASK, new Action<Exec>() {
             public void execute(final Exec t) {
-                t.setDescription("Removes last commit, using 'reset --hard HEAD~'");
-                //TODO replace with combination of 'git reset --soft HEAD~ && git stash' so that we don't lose commits
-                t.commandLine("git", "reset", "--hard", "HEAD~");
+                t.setDescription("Stashes current changes");
+                t.commandLine("git", "stash");
+                t.dependsOn(RESET_COMMIT_TASK);
+            }
+        });
+
+        TaskMaker.execTask(project, RESET_COMMIT_TASK, new Action<Exec>() {
+            public void execute(final Exec t) {
+                t.setDescription("Removes last commit, using 'reset --soft HEAD~'");
+                t.commandLine("git", "reset", "--soft", "HEAD~");
             }
         });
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -111,9 +113,16 @@ public void execute(final Task t) {
...
-                //TODO replace with combination of 'git reset --soft HEAD~ && git stash' so that we don't lose commits
-                t.commandLine("git", "reset", "--hard", "HEAD~");
+                t.setDescription("Stashes current changes");
+                t.commandLine("git", "stash");
+                t.dependsOn(RESET_COMMIT_TASK);
+            }
+        });
+
+        TaskMaker.execTask(project, RESET_COMMIT_TASK, new Action<Exec>() {
+            public void execute(final Exec t) {
```
wwilk: It would be better to create two separate low-level tasks - gitSoftResetCommit and gitStash and one performCommitCleanUp that would depend on them. This is the approach that we use in a few other places, see performGitPush or performRelease. Description of the COMMIT_CLEANUP_TASK is a little confusing and also misleading - it doesn't only stash changes, but also soft-resets them first.

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -28,12 +28,14 @@
  *     <li>performGitPush</li>
  *
  *     <li>gitCommitCleanUp</li>
+ *     <li>gitResetCommit</li>
```
wwilk: Thanks for updating javadoc!

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
epeee: Thx for your feedback! Yes, indeed, we'll end up in a cleaner solution after applying your suggestions. I will update the PR soon.

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -109,11 +113,25 @@ public void execute(final Task t) {
...
+            public void execute(final Task t) {
+                t.setDescription("Performs " + SOFT_RESET_COMMIT_TASK + " and " + GIT_STASH_TASK + " tasks.");
+                t.dependsOn(SOFT_RESET_COMMIT_TASK, GIT_STASH_TASK);
+            }
+        });
+
+        TaskMaker.execTask(project, GIT_STASH_TASK, new Action<Exec>() {
+            public void execute(final Exec t) {
+                t.setDescription("Stashes current changes");
+                t.commandLine("git", "stash");
```
mstachniuk: What do you think about using explicitly `save` as 3rd argument?

epeee: I'd prefer the version without `save` as 3rd argument (since git stash without any arguments is equivalent to git stash save).

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -27,13 +27,17 @@
  *     <li>gitPush</li>
  *     <li>performGitPush</li>
  *
- *     <li>gitCommitCleanUp</li>
+ *     <li>performGitCommitCleanUp</li>
```
szczepiq: Might be good to start documenting the tasks a bit more, writing why we need this task and what it does. For example: "cleans up commit operation so that we can test release conveniently by cleaning up the state of the working copy after the test", "this task will delegate to git soft reset task and then git stash...". Something like that :) This Javadoc will be our public API documentation soon! 1.0 is coming :D

szczepiq:
Thanks @mstachniuk and @wwilk for review!

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
mstachniuk: OK, fine

PR #237, opened 2017-06-14
Moved the responsibility of declaring task relationships

szczepiq:
Into a better plugin, so that we keep the same complexity in the same spot


szczepiq:
Friendly ping :) Any feedback? If not, just merge!

Issue #232, opened 2017-06-13
Autodiscover plugins for Gradle Plugin Portal publication


szczepiq:
### Problem

Currently, we need to [manually configure all plugins](https://github.com/mockito/shipkit/blob/master/build.gradle#L56) published to Plugin Portal in the build.gradle file. This causes the list to be constantly out of date. For example, right now our "core" plugin, "org.shipkit.java" is not in the Plugin Portal :)

### Suggested implementation

- [x] rename "org.shipkit.gradle-plugin-releasing" to "org.shipkit.gradle-plugin" according to our [Gradle plugin reference](https://github.com/mockito/shipkit/wiki/Shipkit-Gradle-Plugin-Reference).
- in "org.shipkit.gradle-plugin":
  - [x] autodiscover plugins and configure automatically what [currently is configured manually](https://github.com/mockito/shipkit/blob/master/build.gradle#L56) and hence it is constantly out of date. To autodiscover we probably need to scan the directory for the *.properties files that contain plugin information.
  - [x] document in the javadoc the use case (e.g. "why") and the new behavior. This will be one of the selling points of our plugin!

szczepiq:
>I have a question, how to proceed with displayName and tags of plugins?
we can't autodiscover them easily. well, we have to add the current data in some other place.
I see 2 possible solutions:
> - add this meta info to the properties file
> - do some config/mapping in the gradle script

Good questions!

Adding metadata to properties file is an interesting idea. Problem is that we would introduce a brand new convention that needs to be documented and learned. Here's what I suggest:

- how about we get "displayName" from the first sentence of the Javadoc? This way, we feed of an existing, well known convention for documenting Java classes. To make progress and benefit from new features you work on, can we just leave display name empty or just use the class name for now? If we use class name for now, we can still override the display name in build.gradle files and provide neat, human-readable description.
- for tags, let's just not add any plugin-specific tags by default. We can set plugins at the "pluginBundle" level in build.gradle file (see the documentation https://plugins.gradle.org/docs/publish-plugin). This way, our implementation is simple and it still provides a way to specify tags for all plugins. I don't have long term solution idea of plugin-specific tags. Perhaps we can use the properties file. Perhaps we just move on, I don't think those tags are that much useful. Let's go for simple implementation!

epeee:
leaving the displayName empty caused the build to fail, I decided to go for the `implementation-class` value from the properties file.
shipkit v0.8.99 contains this feature.

PR #239, opened 2017-06-16
rename org.shipkit.gradle-plugin-releasing -> org.shipkit.gradle-plugin

epeee:
a very first part of #232


szczepiq:
Absolutely! Thank you for clearing the TODO!

PR #240, opened 2017-06-16
make use of spock-core 1.1-groovy-2.4

epeee:


szczepiq:
Nice. Newer versions of Spock offer interesting improvements.

PR #243, opened 2017-06-17
make use of shipkit v0.8.92

epeee:
- new gradle-plugin name


szczepiq:
Absolutely!!! Thank you.

PR #251, opened 2017-06-21
make use of shipkit v0.8.96 (autodiscover plugins feature)

epeee:
... will enable use to make use of new autodiscover feature in shipkit, see #232

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -6,7 +6,7 @@ buildscript {
     }
     dependencies {
         classpath "com.gradle.publish:plugin-publish-plugin:0.9.6"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -55,48 +55,6 @@ pluginBundle {
     description = 'Release tools and Gradle plugins that automate Mockito continuous delivery.'
     tags = ['shipkit', 'continuous delivery', 'release']
 
-    plugins {
```
szczepiq: I love the smell of removed complexity in the evening :)

PR #252, opened 2017-06-21
Leveraged new IdentifyGitBranchTask

szczepiq:
- This simplifies the code. The previous idea with GitStatus plugin had some limitations. The code now is more conventional and easier to work with.
- I haven't tested with real project yet but sending pull request for early review :)


src/main/groovy/org/shipkit/internal/gradle/git/GitBranchPlugin.java
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Adds tasks:
+ * <ul>
+ *     <li>identifyGitBranch - {@link IdentifyGitBranchTask}</li>
+ * </ul>
+ */
+public class GitBranchPlugin implements Plugin<Project> {
+
+    public static final String IDENTIFY_GIT_BRANCH = "identifyGitBranch";
```
epeee: just a minor one: it looks like we have two `IDENTIFY_GIT_BRANCH` now (`GitPlugin#IDENTIFY_GIT_BRANCH` and this one).
Let's just have one of them.

szczepiq: Nice catch!

PR #256, opened 2017-06-24
cleanup shipkit.gradle

epeee:
removed line since this configuration is default now since v0.8.94.
We are already using v0.8.99.


szczepiq:
Nice! The smaller configuration file, the better!

Issue #154, opened 2017-05-17
Release to Maven Central under org.shipkit group


mockitoguy:
Due to access to accounts and passwords, this feature needs to be implemented by core team member.

### Problem

We should ship shipkit to maven central in addition to Gradle Plugin portal because:
 - it works well for credibility, discoverability (maven central is considered source of truth)
 - we can build reusable plugin that helps other Gradle plugins that want to use Shipkit to publish to maven central (another use case - selling point)
 - we can support dynamic versions '+' (Gradle Plugin Portal [does not support dynamic versions](https://discuss.gradle.org/t/why-the-gradle-central-plugin-repository-does-not-support-dynamic-plugin-versions/11788)). Dynamic versions are super useful for bootstrapping scenarios, for auto testing, and for version discovery. Example use of dynamic version.

```
buildscript {
  repositories { mavenCentral() }
  dependencies {
    classpath "org.shipkit:shipkit:+"
  }
}
```

### Suggested implementation

- [x] rename "org.shipkit.gradle-plugin-releasing" to "org.shipkit.gradle-plugin" according to our [Gradle plugin reference](https://github.com/mockito/shipkit/wiki/Shipkit-Gradle-Plugin-Reference).
- [ ] tweak "org.shipkit.gradle-plugin" so that it:
  - publishes to Gradle Plugin Portal and to Maven Central under the "org.shipkit" group. This might be tricky, however we already publish to Bintray and have a way to publish to Maven Central (see the Mockito build) so it should be doable.
  - to start shipping to Maven Central, we need to complete the process started at https://issues.sonatype.org/browse/OSSRH-31217. The credentials to Sonatype account are stored in our team's accounts and passwords document.
- [x] touch base with Gradle team behind http://plugins.gradle.org and request to remove "gradle.plugin" prefix from "org.shipkit" group name. Currently, we configure the group as ["gradle.plugin.org.shipkit"](https://github.com/mockito/shipkit/blob/165bc4e718a5b13cd90fa3fccdcd058af57dbe35/build.gradle#L31) which looks unprofessional (it's how Gradle Plugin portal works by default).
  - Gradle Plugin Portal [documentation](https://plugins.gradle.org/plugin/org.shipkit.release-notes)
  - [Getting help](https://plugins.gradle.org/docs/get-help)

mstachniuk:
[...]

PR #258, opened 2017-06-25
#154 Change groupId from gradle.plugin.org.shipkit to org.shipkit

mstachniuk:


szczepiq:
Nice! Thank you for tidying it up!

PR #265, opened 2017-06-26
GitCommitTask works when version.properties is not tracked by git yet

wwilk:
"git commit" with a list of files to commit doesn't work when at least one of these files is not tracked by git yet. It is important for bootstraping - "ciPerformRelease" fails when run for the first time. It works if we use "git add" first, with the new CompositeExecTask we can do it in one task, without duplication of configuration.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
@@ -4,10 +4,9 @@ import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
-import org.gradle.api.logging.Logger;
-import org.gradle.api.logging.Logging;
 import org.gradle.api.tasks.Exec;
 import org.shipkit.gradle.ReleaseConfiguration;
+import org.shipkit.gradle.exec.ExecCommand;
 import org.shipkit.gradle.git.GitPushTask;
 import org.shipkit.gradle.git.IdentifyGitBranchTask;
 import org.shipkit.internal.gradle.git.GitBranchPlugin;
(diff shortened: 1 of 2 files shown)


mstachniuk:
Looks ok 👍  merge at will

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -52,28 +50,18 @@
...
-    private final static Logger LOG = Logging.getLogger(GitPlugin.class);
-
     public void apply(final Project project) {
         final ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
 
         TaskMaker.task(project, GIT_COMMIT_TASK, GitCommitTask.class, new Action<GitCommitTask>() {
             public void execute(final GitCommitTask t) {
                 t.setDescription("Commits all changed files using generic --author and aggregated commit message");
+                //TODO WW create unit tests
                 t.doFirst(new Action<Task>() {
```
szczepiq: I assume we use doFirst() so that the configuration can reflect the files to commit, specified by other plugins? Might be worth adding a code comment.

src/main/groovy/org/shipkit/internal/gradle/GitPlugin.java
```diff
@@ -154,6 +142,25 @@ public void execute(final Exec t) {
         });
     }
 
+    private List<String> getAddCommand(List<String> files) {
```
szczepiq: Nitpick (feel free to ignore ;). Using "asList()" gives more compact code that could be a tiny bit more readable.

wwilk: It's not that easy in this case, since we add list of files into the list with other arguments.

wwilk:
Applied some suggestions and merging

PR #267, opened 2017-06-27
Fixed Travis CI git branch handling

### LINKED ISSUE #264: Task assertReleaseNeeded fails because of wrong discovered branch name on Travis CI
In task `assertReleaseNeeded` we read current branch name using this git command:

```
git rev-parse --abbrev-ref HEAD
```

It's a valid command util we are not in detached head (then above command return `HEAD`)! Unfortunately Travis-CI does `git checkout -qf sha1` after clone:
https://travis-ci.org/mockito/shipkit-example/builds/247233789#L146

and branch name is other as expected in `assertReleaseNeeded`:
https://travis-ci.org/mockito/shipkit-example/builds/247233789#L408

The newest version of shipkit-example in bintray is 0.15.2 instead of 0.16.3 (13 revisions are missing).

szczepiq:
- Fixes #264. Reworked TravisPlugin so that it overrides branch on the IdentifyGitBranchTask in Travis CI scenarios
  - I'd rather not hardcode the handling of 'TRAVIS_BRANCH' env variable in the IdentifyGitBranchTask. This env variable handling should be a part of TravisPlugin.
- Renames, logging improvements, Javadoc

Tested by:
 - running ```./gradlew testRelease``` with and w/o 'TRAVIS_BRANCH' env variable exported, checked the build outputs. 
 - running ```./gradlew assertReleaseNeeded```, checked the build outputs.

Thank you for reporting the bug!

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/git/IdentifyGitBranchTask.java
+++ b/src/main/groovy/org/shipkit/gradle/git/IdentifyGitBranchTask.java
@@ -2,6 +2,8 @@ package org.shipkit.gradle.git;
 
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
 import org.gradle.api.tasks.TaskAction;
 import org.shipkit.internal.exec.Exec;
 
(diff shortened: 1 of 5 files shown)


mstachniuk:
I'm not sure it will solve https://github.com/mockito/shipkit/issues/264 issue... I just tried to run it in my head so I can be wrong ;-)
I think `ReleaseNeededTask` will still use `git rev-parse --abbrev-ref HEAD` for determine a branch name. I checked where `ReleaseNeededTask#setBranch()` is called.

What do you think about this possible solution:
* read a branch name from `TRAVIS_BRANCH` first
* if it's `null` let's use a value from `git rev-parse --abbrev-ref HEAD`
* propagate a correct branch name to the all tasks depend on branch

I this case on Travis the branch name will be correct and release from local machine or other CI server will be easier. And all tasks will have the same branch name.

szczepiq:
Thanks for review! The plan you outlined is essentially what we do now, the code is bit more decoupled (and probably more complicated, too ;p). I updated the description with more details.

szczepiq:
Since this is blocking @wwilk, I'll merge :)

PR #270, opened 2017-06-27
run animal sniffer only in CI

### LINKED ISSUE #266: Animal sniffer is slow, it would be nice to only run it on CI
Animal sniffer is slow, it would be nice to only run it on CI. Anyone wants to contribute? :) We already do something like that in Mockito project, perhaps we can implement the same idea?

epeee:
fixes #266

One can also pass `-PcheckJava6Compatibility` to the build locally in order to run animal sniffer.
In CI we are passing the property mentioned to the build by default.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ install:
  - true
 
 script:
-  - ./gradlew build -s -i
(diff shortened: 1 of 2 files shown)


szczepiq:
Thank you! I like fast builds :)

PR #273, opened 2017-06-28
Cleaner output from processes forked from Gradle

szczepiq:
This helps understanding what's going on when we shell out Gradle build from Gradle build. We need to do it for certain scenarios unfortunately. This change makes the view much cleaner.

Example, notice the "[]" prefix:

```
:testRelease
  Executing:
    ./gradlew releaseNeeded performRelease releaseCleanUp -Pshipkit.dryRun
[./gradlew] Starting a Gradle Daemon, 1 busy Daemon could not be reused, use --status for details
[./gradlew]   Building version '0.16.4' (value loaded from 'version.properties' file).
[./gradlew]   'git push' does not use GitHub write token because it was not specified.
```

More output tidy-ups are pending. This is a good 1st step.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/exec/ExternalProcessStream.java
+++ b/src/main/groovy/org/shipkit/internal/exec/ExternalProcessStream.java
@@ -1,0 +1,58 @@
+package org.shipkit.internal.exec;
+
+import org.shipkit.internal.util.ArgumentValidation;
+
+import java.io.IOException;
+import java.io.OutputStream;
+import java.io.PrintStream;
+
+/**
+ * On top of standard output stream, this implementation adds prefixing of the output.
+ * It is super useful to make the log output clear.
+ */
+public class ExternalProcessStream extends OutputStream {
+
+    private final PrintStream output;
+    private final String outputPrefix;
+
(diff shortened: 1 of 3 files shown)


src/main/groovy/org/shipkit/internal/exec/ExternalProcessStream.java
```diff
@@ -0,0 +1,58 @@
...
+    public ExternalProcessStream(String outputPrefix, PrintStream output) {
+        ArgumentValidation.notNull(outputPrefix, "outputPrefix", output, "output");
+        this.outputPrefix = outputPrefix;
+        this.output = output;
+    }
+
+    @Override
+    public void write(int b) throws IOException {
+        //keep building the line
+        line.append((char) b);
```
epeee: Maybe I missed something, but why do we need the `line`?

szczepiq: You are exactly right! It is not needed. I will remove it in the subsequent RB.

szczepiq: Nice catch :)

PR #277, opened 2017-07-02
TRAVIS_BRANCH only used when it's available + decreased logging level

wwilk:
When run locally and without TRAVIS_BRANCH env var set, branch in IdentifyGitBranch task will be incorrect. I would set it only if TRAVIS_BRANCH is set which always happen on Travis.

About logging in AllContributorsSerializer - I think this information is not needed on info level, especially that for projects with a lot of contributors it will obfuscate the log, see example from Mockito below: 
screen shot 2017-07-02 at 19 39 04

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/TravisPlugin.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/TravisPlugin.java
@@ -1,5 +1,6 @@
 package org.shipkit.internal.gradle;
 
+import org.apache.commons.lang.StringUtils;
 import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/shipkit/internal/gradle/TravisPlugin.java
```diff
@@ -40,7 +41,9 @@ public void apply(final Project project) {
             @Override
             public void execute(GitBranchPlugin p) {
                 IdentifyGitBranchTask identifyBranch = (IdentifyGitBranchTask) project.getTasks().getByName(IDENTIFY_GIT_BRANCH);
-                identifyBranch.setBranch(branch);
+                if(StringUtils.isNotEmpty(branch)) {
```
szczepiq: Can you use our own StringUtil instead? This way we avoid dependency on apache commons (less dependencies -> less problems for our users).

szczepiq:
Merging from my phone ;)

wwilk:
Thanks, great that you caught this apache-commons dependency, I missed it!

PR #278, opened 2017-07-03
Removed configuration logic that I think we don't really need

szczepiq:
- I'm trying to simplify the public API as much as possible before 1.0.
- I don't think we really need the configuration option for publishing all java subprojects. I suspect that we can have simpler implementation and still be able to satisfy the use case.
- If someone does not want to publish all java subprojects, he can apply a different set of plugins, instead of applying opinionated "org.shipkit.java". He can also disable bintray upload task to avoid publications. E.g. instead of us providing configuration option, we can relay on standard Gradle techniques to avoid publishing of some modules.

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
@@ -1,7 +1,6 @@
 package org.shipkit.gradle;
 
 import org.gradle.api.GradleException;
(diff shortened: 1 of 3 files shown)


szczepiq:
Friendly ping :)

szczepiq:
>just a quick note: we are currently using publishAllJavaSubprojects = false in mockito project

I know :) I'll figure it out!

szczepiq:
Thank you for review!

PR #211, opened 2017-06-06
E2E improve javadocs, add shallow clone, rename dogfooding plugin name to shipkit

mstachniuk:
e2e is not finished yet and not integrated with main build, so this PR is safe to merge.

@szczepiq I need your tips how to connect this code and Groovy DSL to define with repositories should be tested by e2e plugin. The code has changed a little bit since https://github.com/mockito/mockito-release-tools/pull/100
Feel free to commit to my branch.

PR #279, opened 2017-07-03
E2ETest based on Gradle TestKit

wwilk:
I created a separate pull request because my idea is a little different from the direction suggested in #211 .

Please only see the commit "E2ETest based on Gradle TestKit"!

These are the important changes. Basically the idea is to use Gradle TestKit, so that we can run end to end tests with ./gradlew e2eTest task. The first step is to clone project (eg. shipkit-example) to a temp test folder, then run performRelease on it with Shipkit as a composite build and then verify if all of the tasks (including shipkit ones) were run and if none of them finished with failure. Currently it skips gitPush and bintrayUpload, the next step would be to include them.

I only did it for shipkit-example to show you the idea, but I will make it generic and create a test for mockito too before merging, if this PR is approved.


szczepiq:
We talked about this idea at length on slack. I'm mentioning it for visibility ;) @wwilk has really good traction on this ticket!

wwilk:
Closing it for now, since we decided to go another way about e2eTests

PR #281, opened 2017-07-04
Plugins tidy-up before 1.0

szczepiq:
- tidied up plugin relationships before 1.0 release. Removed some plugins, reorganized how we apply some other plugins. the goal was to have the plugin architecture more decoupled, easier to understand
- renamed plugins and placed them in more correct packages
- added missing *.properties files for plugins - updated javadoc for many plugins

This change does not affect the behavior of Shipkit. Please review! The change is quite big because I worked on this in the train from Warsaw yesterday :)


docs/plugin-hierarchy.md
```diff
@@ -0,0 +1,28 @@
+## Plugin hierarchy
+
```
wwilk: Awesome!

src/main/groovy/org/shipkit/internal/gradle/release/CiReleasePlugin.java
```diff
@@ -46,11 +53,11 @@ public void apply(Project project) {
             public void execute(ShipkitExecTask task) {
                 task.setDescription("Checks if release is needed. If so it will prepare for ci release and perform release.");
                 task.getExecCommands().add(execCommand(
-                        "Checking if release is needed", asList("./gradlew", ASSERT_RELEASE_NEEDED_TASK), stopExecution()));
+                        "Checking if release is needed", asList("./gradlew", ReleaseNeededPlugin.ASSERT_RELEASE_NEEDED_TASK), stopExecution()));
```
wwilk: 👍 this is much more readable

docs/plugin-hierarchy.md
wwilk: Can we include here the information which plugins are public and which internal? Or maybe also which are external (not owned by Shipkit)

szczepiq: Sure we can! I'm not sure if we can call any of our plugins "internal" because they introduce tasks and stuff that clients depend on. All plugin classes are internal but the behavior they introduce (like tasks they add) is public and we need to keep it compatible.

szczepiq:
Thank you for review!

PR #288, opened 2017-07-04
Improvements to comparing publications

szczepiq:
- Made the task incremental by storing the diffs in the file which is an output from the task
- Renamed the task so that it is more consistent in how we name tasks (usually it is a verb)

Tested with example project for various combinations - changed the source code and observed releaseNeeded report correct output. Also validated the task is correctly incremental, and it is up-to-date if the inputs don't change. It works great!


szczepiq:
Friendly ping ;) I have 2 more PRs in the line :D

src/main/groovy/org/shipkit/gradle/ReleaseNeededTask.java
```diff
@@ -24,23 +19,15 @@
  *  - the current Git branch does not match release-eligibility regex ({@link #getReleasableBranchRegex()}.
  *  - binaries have not changes since the previous release
  */
+//TODO move to 'release' subpackage
 public class ReleaseNeededTask extends DefaultTask {
```
szpak: A good candidate for being [@CacheableTask])https://docs.gradle.org/4.0/javadoc/org/gradle/api/tasks/CacheableTask.html) once only newer Gradle 3.x versions are supported.

szczepiq: This is a great idea to use Gradle's build cache! Thanks!!!

We will probably use programatic API instead of the annotation to avoid compile time dependency on latest Gradle. This also allows us to conditionally apply the setting based on the Gradle version. Also, we need to add it to more tasks, most of our tasks are incremental :)

Thanks for suggestion and the review!

szczepiq:
Thank you @epeee for review!!! Merging.

Issue #280, opened 2017-07-03
Failover for publicationsComparatorTask


wwilk:
## Problem
PublicationsComparatorTask downloads artifacts of previous version, sometimes they are not available because bintrayUpload task failed (silently or not). Since gitPush task runs before bintrayUpload, version could be bumped anyway. When download fails -> assertReleaseNeeded fails too. The only solution in this case is manual change of previousVersion to the last one that was actually released.

## Proposed solution
In case there is a fail during download we should not fail assertReleaseNeeded. Additionally we can try to automatically choose the last version actually released on Bintray. It has to be checked if Bintray latest URL can be used, or alternatively we can try to download artifact for the few latest tags, but in the latter case we have to remember about limiting these retries.

szczepiq:
Great idea. At the very least we should not fail assertReleaseNeeded!

szczepiq:
I'm working a lot in publications comparison. This is very nice code and works well! I'm updating the Gradle part of the API and should be able to fix this problem soon.

szczepiq:
This is done!

PR #289, opened 2017-07-05
Release needed task - made comparing publications safe

szczepiq:
Fixes issue #280

- when downloading of previous publications fails we will not stop the build
- there are legit use cases where there are no previous publications - on initial bootstrap of shipkit, when previous release failed for some reason
- added unit tests for new functionality, refactored the code a bit to make testing easier

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/java/ComparePublicationsTask.java
+++ b/src/main/groovy/org/shipkit/gradle/java/ComparePublicationsTask.java
@@ -24,16 +24,20 @@ import java.io.File;
  */
 public class ComparePublicationsTask extends DefaultTask {
 
+    @OutputFile private File comparisonResult;
+
     @Input private String projectGroup;
     @Input private String currentVersion;
     @Input @Optional private String previousVersion;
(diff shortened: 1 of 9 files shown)


src/main/groovy/org/shipkit/gradle/java/ComparePublicationsTask.java
```diff
@@ -24,16 +24,20 @@
...
-    @InputFile private File previousPom;
-    @InputFile private File previousSourcesJar;
+    @Input private String pomTaskName;
 
-    @OutputFile private File comparisonResult;
+    //Not using @InputFile annotation on purpose below. @InputFile makes Gradle fail early
+    // when the file path specified but file does not exist (@Optional does not help).
+    // Using @Input is enough for this use case
+    @Input @Optional private File previousPom;
+    @Input @Optional private File previousSourcesJar;
```
wwilk: Great, an improvement on the way to fail-safe ComparePublications!

wwilk:
Merging because I need these changes for E2E tests

PR #295, opened 2017-07-06
Minor style improvements

szpak:
A side effect of getting through those files/classes.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -44,7 +44,7 @@ dependencies {
     testCompile("org.spockframework:spock-core:1.1-groovy-2.4") {
         exclude module: "groovy-all"
     }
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -44,7 +44,7 @@ dependencies {
     testCompile("org.spockframework:spock-core:1.1-groovy-2.4") {
         exclude module: "groovy-all"
     }
-    testCompile "cglib:cglib-nodep:2.2.2"
+    testCompile "net.bytebuddy:byte-buddy:1.7.1"
```
szczepiq: I didn't know spock can work with byte-buddy instead of cglib. Thank you!

Issue #300, opened 2017-07-07
0.8.120 was released instead of 0.8.119


szczepiq:
For some reason version "0.8.120" [was published](https://plugins.gradle.org/search?term=org.gradle.shipkit) to Plugin Portal, instead of "0.8.119". The tag and release notes were created for "0.8.119". Will investigate.

szczepiq:
This should be fixed.

PR #301, opened 2017-07-07
Fixed problem with org.shipkit.gradle-plugin

szczepiq:
Addresses bug #300

- I have not tested the previous implementation end-to-end and it had a fundamental flaw. We cannot really run 'publishPlugins' in a forked Gradle build because we would be shipping a different version. We tagged and prepared release notes for version 'n' and forked 'publishPlugins' released version 'n+1'. Initially, I used the forked Gradle invocation to tackle configurability deficiencies in the Gradle's official plugin.
- New implementation sets project properties based on env variables. This should work just fine for us. We don't need publish key/secret properties on the task API at this moment.
- Added initial integration test for 'org.shipkit.gradle-plugin' plugin.

Tested by doing 'testRelease' locally:

```
[./gradlew] :gitSoftResetCommit
[./gradlew]   Running:
[./gradlew]     git reset --soft HEAD~
[./gradlew] :gitStash
[./gradlew]   Running:
[./gradlew]     git stash
[./gradlew] Saved working directory and index state WIP on sf2: 966d8ee testing only
[./gradlew] HEAD is now at 966d8ee testing only
[./gradlew] :performGitCommitCleanUp
[./gradlew] 
[./gradlew] BUILD SUCCESSFUL
[./gradlew] 
[./gradlew] Total time: 16.502 secs

BUILD SUCCESSFUL
```


szczepiq:
Friendly ping :)

szczepiq:
Thank you for review!

Issue #302, opened 2017-07-08
Create a pull request in dependent repositories with a Shipkit version bump


wwilk:
## Problem
We have to update Shipkit's version in Mockito and Shipkit-example every time we release, which is quite often.

## Solution
We should automatically create a pull request with a version bump whenever there is release. Proposed flow:
- e2eTest
- Shipkit release
- send a PR

And sending a PR would be divided into:
- cloning a fork of the dependent repo
- checking out a new branch with the new version in the name
- bumping version of Shipkit in the provided config file (defaults to build.gradle)
- commiting config file
- pushing it and creating a pull request

wwilk:
The 1.0 version of this feature is done. I will create another issue for improving it further.

PR #304, opened 2017-07-08
AutoUpdateConsumerPlugin added #302

wwilk:
I added AutoUpdateCustomerPlugin, that adds a task "performAutoUpdate", which in turn does the following:
- checkouts branch "shipkit-version-bumped-$VERSION"
- replaces version in given config file 
- commits and pushes changes

This is the first step of automated pull requests. 

Tested so far on shipkit-example:

./gradlew performAutoUpdate -PshipkitNewVersion=0.1.5 -s

  ```
Building version '0.16.7' (value loaded from 'version.properties' file).
  'git push' uses GitHub write token specified by GH_WRITE_TOKEN env variable.
:autoUpdateCheckoutBranch
  Running:
    git checkout -b shipkit-bumped-version-0.1.5
Switched to a new branch 'shipkit-bumped-version-0.1.5'
:autoUpdateReplaceVersion
  Replacing version in /Users/wilk/Projects/shipkit-example/build.gradle using pattern org.shipkit:shipkit:{VERSION} and version 0.1.5.
:autoUpdateGitCommit
  Running:
    git commit -m Shipkit version updated to 0.1.5 build.gradle
[shipkit-bumped-version-0.1.5 526127b] Shipkit version updated to 0.1.5
 1 file changed, 1 insertion(+), 1 deletion(-)
:autoUpdateGitPush
  Running:
    git push -u origin shipkit-bumped-version-0.1.5
To github.com:mockito/shipkit-example.git
 * [new branch]      shipkit-bumped-version-0.1.5 -> shipkit-bumped-version-0.1.5
Branch shipkit-bumped-version-0.1.5 set up to track remote branch shipkit-bumped-version-0.1.5 from origin.
:performAutoUpdate
```

### CODE REVIEW

**Comment by szczepiq** [comment]
I really like the idea and the first step!

Things to consider / discuss:
 - projects that we don't have commit rights and cannot push branches
 - how are we hooking this up to the release workflow
 - good naming convention. "auto update" seem ambiguous, this is more about "version upgrade".

I can give more feedback when I have a chance to look at the code.

Thanks for making nice progress on this!

File: src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java [hunk 1]
```diff
@@ -0,0 +1,126 @@
...
+    public static final String SHIPKIT_NEW_VERSION = "shipkitNewVersion";
+    public static final String SHIPKIT_DEPENDENCY_FILE = "shipkitDependencyFile";
+    public static final String SHIPKIT_DEPENDENCY_PATTERN = "shipkitDependencyPattern";
+    public static final String DEPENDENCY_PATTERN_DEFAULT = "org.shipkit:shipkit:{VERSION}";
+    public static final String DEPENDENCY_FILE_DEFAULT = "build.gradle";
+
+    @Override
+    public void apply(final Project project) {
+
+        TaskMaker.execTask(project, AUTO_UPDATE_CHECKOUT_BRANCH, new Action<Exec>() {
```
szczepiq: You can use existing "GitCheckOutTask"

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+            }
+        });
+
+        TaskMaker.task(project, AUTO_UPDATE_REPLACE_VERSION, ReplaceVersionTask.class, new Action<ReplaceVersionTask>() {
+            @Override
+            public void execute(final ReplaceVersionTask task) {
+                task.setDescription("Replaces Shipkit's dependency version in config file.");
+                task.mustRunAfter(AUTO_UPDATE_CHECKOUT_BRANCH);
+
+                LazyConfiguration.lazyConfiguration(task, new Runnable() {
```
szczepiq: Is there a good reason to lazily configure it? It's best to avoid laziness if it is not needed for simplicity.

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+                    @Override
+                    public void run() {
+                        String file = getDependencyFile(project);
+                        exec.commandLine("git", "commit", "-m", "Shipkit version updated to " + getShipkitNewVersion(project), file);
+                    }
+                });
+            }
+        });
+
+        TaskMaker.execTask(project, AUTO_UPDATE_GIT_PUSH, new Action<Exec>() {
```
szczepiq: Does it make sense to use existing "GitPushTask"? The only difference seems to be using '-u' parameter here, which I don't think we really need.

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+    public static final String AUTO_UPDATE_CHECKOUT_BRANCH = "autoUpdateCheckoutBranch";
+    public static final String AUTO_UPDATE_REPLACE_VERSION = "autoUpdateReplaceVersion";
+    public static final String AUTO_UPDATE_GIT_COMMIT = "autoUpdateGitCommit";
+    public static final String AUTO_UPDATE_GIT_PUSH = "autoUpdateGitPush";
+
+    public static final String SHIPKIT_NEW_VERSION = "shipkitNewVersion";
+    public static final String SHIPKIT_DEPENDENCY_FILE = "shipkitDependencyFile";
+    public static final String SHIPKIT_DEPENDENCY_PATTERN = "shipkitDependencyPattern";
+    public static final String DEPENDENCY_PATTERN_DEFAULT = "org.shipkit:shipkit:{VERSION}";
+    public static final String DEPENDENCY_FILE_DEFAULT = "build.gradle";
```
szczepiq: "dependency file" name is somewhat confusing - I think this is simply "build file" :)

src/main/groovy/org/shipkit/internal/gradle/autoupdate/ReplaceVersionTask.java
```diff
@@ -0,0 +1,82 @@
...
+
+    /**
+     * See {@link #getNewVersion()}
+     */
+    public void setNewVersion(String newVersion) {
+        this.newVersion = newVersion;
+    }
+
+    /**
+     * Config file where dependency will be updated
```
szczepiq: "config file" is somewhat misleading, I suggest to call it "build file" or just "target file"

src/main/groovy/org/shipkit/internal/gradle/autoupdate/ReplaceVersionTask.java
```diff
@@ -0,0 +1,82 @@
...
+import java.io.File;
+
+/**
+ * Replaces version of given dependency with {@link ReplaceVersionTask#newVersion}
+ * in the given {@link ReplaceVersionTask#configFile}
+ * Use {@link ReplaceVersionTask#dependencyPattern} to tell Shipkit how it should
+ * replace the dependency. You should set it to regex pattern that contains
+ * {@link ReplaceVersionTask.VERSION_PLACEHOLDER} instead of version number.
+ */
+public class ReplaceVersionTask extends DefaultTask{
```
szczepiq: I love this task! Nice docs and well encapsulated functionality, easy to reuse!

src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
```diff
@@ -124,7 +124,7 @@ public static void writeFile(File target, String content) {
             p = new PrintWriter(new OutputStreamWriter(new FileOutputStream(target), "UTF-8"));
             p.write(content);
         } catch (Exception e) {
-            throw new RuntimeException("Problems writing text to file: " + target);
+            throw new RuntimeException("Problems writing text to file: " + target, e);
```
szczepiq: Well spotted, thank you!

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
wwilk: Unfortunately in the use cases I thought of new version will be passed as a project property in CLI invocation, so it won't be always present. For example if we use this task from CI, we will have to call it on the clone of client project repo, with version passed as parameter (shipkit version != client project version). Maybe later we can have a service that would call bintray to see what's the latest version, but in this case lazy configuration would also apply

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
wwilk: Seems like it, I will try it!

src/main/groovy/org/shipkit/internal/gradle/autoupdate/ReplaceVersionTask.java
wwilk: The thing is that it doesn't have to a build file, in case of mockito shipkit dependency is kept in gradle.properties.

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
```diff
@@ -0,0 +1,126 @@
...
+import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.gradle.api.tasks.Exec;
+import org.shipkit.internal.gradle.autoupdate.ReplaceVersionTask;
+import org.shipkit.internal.gradle.configuration.LazyConfiguration;
+import org.shipkit.internal.gradle.util.TaskMaker;
+
+public class AutoUpdateConsumerPlugin implements Plugin<Project> {
+
+    public static final String AUTO_UPDATE_CHECKOUT_BRANCH = "autoUpdateCheckoutBranch";
```
wwilk: 👍

src/main/groovy/org/shipkit/internal/gradle/AutoUpdateConsumerPlugin.java
wwilk: 👍

wwilk:
I added checking out base branch (branch which dependency will be upgraded on eventually) at the beginning and creating pull request at the end. Test run on shipkit-example:

./gradlew performVersionUpgrade -PdependencyNewVersion=0.5.8

  ```
Building version '0.16.7' (value loaded from 'version.properties' file).
  'git push' uses GitHub write token specified by GH_WRITE_TOKEN env variable.
Applying VersionUpgradeConsumerPlugin, beware that it's is INCUBATING state, so its API may change!
  'git push' uses GitHub write token specified by GH_WRITE_TOKEN env variable.
  'git push' uses GitHub write token specified by GH_WRITE_TOKEN env variable.
:versionUpgradeCheckoutBaseBranch
  Executing:
    git checkout master
:versionUpgradeCheckoutVersionBranch
  Executing:
    git checkout -b shipkit-version-bumped-0.5.8
:versionUpgradeReplaceVersion
  Replacing version in '/Users/wilk/Projects/shipkit-example/build.gradle' using pattern 'org.shipkit:shipkit:{VERSION}' and version '0.5.8'.
:versionUpgradeGitCommit
  Running:
    git commit -m Shipkit version updated to 0.5.8 build.gradle
[shipkit-version-bumped-0.5.8 0bb2fe9] Shipkit version updated to 0.5.8
 1 file changed, 1 insertion(+), 1 deletion(-)
:versionUpgradeGitPush
  Executing:
    git push https://shipkit:[SECRET]@github.com/mockito/shipkit-example.git shipkit-version-bumped-0.5.8
:versionUpgradeCreatePullRequest
  Creating a pull request of title 'Shipkit version bumped to 0.5.8' in repository 'mockito/shipkit-example' between base = 'master' and head = 'shipkit-version-bumped-0.5.8'.
:performVersionUpgrade

BUILD SUCCESSFUL
```

**Comment by szczepiq** [comment]
I've pulled the changes, will review on the way from Krakow to Bielsko :)

File: src/main/groovy/org/shipkit/internal/gradle/git/GitCheckOutTask.java [hunk 9]
```diff
@@ -27,8 +36,45 @@ public String getRev() {
...
+
+    private ProcessRunner getProcessRunner(){
+        if(processRunner == null){
+            return new DefaultProcessRunner(getProject().getProjectDir());
+        }
+        return processRunner;
+    }
+
+    @ExposedForTesting
+    protected void setProcessRunner(ProcessRunner processRunner){
```
anicos: `@ExposedForTesting` is not good idea. Why should avoid using this annotation i new code. Why you don't set `ProcessRunner` by constructor?

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequestTask.java
```diff
@@ -0,0 +1,142 @@
...
+            "  \"title\": \"" + title + "\"," +
+            "  \"body\": \"Please pull this in!\"," +
+            "  \"head\": \"" + headBranch + "\"," +
+            "  \"base\": \"" + baseBranch + "\"" +
+            "}";
+
+        try {
+            gitHubApi.post("/repos/" + repositoryUrl + "/pulls", body);
+        } catch(IOException e){
+            LOG.error("  Creating a pull request failed.\n  {}", e.getMessage());
```
anicos: Maybe you should throw RuntimeException or something which extened RuntimeException? The question is. If post throw exception the task should be interrupt?

src/main/groovy/org/shipkit/internal/gradle/git/GitCheckOutTask.java
wwilk: It's not so easy with Gradle tasks, they cannot be instantiated directly (they throw TaskInstantiationException if you try to call one of the constructors). See constructors of org.gradle.api.internal.AbstractTask for more details. I didn't find a better way of testing than the one above. If you find something better, please let me know.

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequestTask.java
wwilk: Yeah this post should throw an exception, it doesn't make sense to go forward if this task fails. I forgot to rethrow it, thanks for catching it!

szczepiq:
I read the code and have bunch of feedback. Regardless, let's ship it and keep iterating!

- Thank you for diligently writing great Javadocs and unit tests!!! The quality of shipkit code is really inspiring!!!
-  versionUpgradeCheckoutBaseBranch - I’m wondering if we need this task. We can create branch off origin/master (and make it configurable).
- I suggest new task names that are shorter and hence easier to use:
    - checkoutVersionUpgradeBranch
    - replaceVersion
    - commitVersionUpgrade
    - pushVersionUpgrade
    - createPullRequest
    - preformVersionUpgrade
- lazyConfiguration is used eagerly but I don’t think it is really needed (less lazy configuration -> simpler implementation)
- it would be great to have the intended use of the plugin documented in the javadoc.
- it would be great to support -PdryRun parameter so that we can performVersionUpgrade the process without making git push and PR creation. We could even create an integration test in similar way as ShipkitJavaPluginIntegTest.
- branch name suggestion for consistency and clarity: “upgrade-[dependency]-to-[version]”
- DEPENDENCY_PATTERN_DEFAULT - given that this plugin is intended for general use, I think we should avoid the default dependency pattern. Let’s require dependency pattern from the user. This kind of use would be most convenient: “./gradlew performVersionUpgrade -Pdependency=org.shipkit:shipkit:0.9.2”
- I’m not sure we need all the project properties - less properties -> smaller the public API we need to maintain. Given that we expose those settings at the task level, users can easily configure anything using various ways - env variables, project properties, etc. - but we don’t have to maintain the configurability. I suggest we keep the “supported” project properties at minimum. It feels that we only need one property to get started with: “-Pdependency=org.shipkit:shipkit:0.9.2”.
- in few places I see that we’re using "version bump” vocabulary when describing “version upgrades”. I suggest that we use same vocabulary consistently: "version upgrades”. Version bumps are something else - we use the to bump our own version during the release process.
- I suggest to consistently use “version upgrade” vocabulary. E.g. “VersionUpgradePlugin”, “org.shipkit.version-upgrade”, etc.
- I like your idea to to move GitHub auth token env variable handling to ReleaseConfiguration. Let’s do it 😃

Take my feedback as guidance and ship at will. I am nitpicky at times but I love progress and I'm happy to move on.

src/main/groovy/org/shipkit/internal/gradle/git/GitCheckOutTask.java
szczepiq: I suggest to push out all implementation details out of GitCheckOutTask. See how it is done in ReleaseNeededTask. To me, it feels that  "@ExposedForTesting" is not really needed given that the type that holds the method is internal anyway. We can also keep methods exposed for testing as package-protected.

Good discussion!

wwilk:
OK, then I'm merging it now since it grew quite big. And I will copy-paste the last suggestions you made to a new PR.

PR #306, opened 2017-07-11
Logging tweaks to improve the clarity of the build log

szczepiq:
Tested by doing ```./gradlew testRelease``` and inspecting the build log:

```
[./gradlew] :gitStash
[./gradlew]   Running:
[./gradlew]     git stash
[./gradlew] Saved working directory and index state WIP on sf: 400b90d Bumped Shipkit
[./gradlew] HEAD is now at 400b90d Bumped Shipkit
[./gradlew] :performGitCommitCleanUp
[./gradlew] 
[./gradlew] BUILD SUCCESSFUL
[./gradlew] 
[./gradlew] Total time: 28.313 secs
  External process [[./gradlew] ] completed.
  The release test was successful. Ship it!

BUILD SUCCESSFUL
```

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
@@ -31,7 +31,7 @@ public class ShipkitExec {
                     LOG.lifecycle("  " + execCommand.getDescription() + ":\n    " + StringUtil.join(execCommand.getCommandLine(), " "));
                 }
             });
-            LOG.lifecycle("  External process [{}] completed.", execCommand.getLoggingPrefix());
+            LOG.lifecycle("  External process {} completed.", execCommand.getLoggingPrefix().trim());
             execCommand.getResultAction().execute(result);
         }
     }
--- a/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
@@ -31,6 +31,7 @@ public class ShipkitExec {
                     LOG.lifecycle("  " + execCommand.getDescription() + ":\n    " + StringUtil.join(execCommand.getCommandLine(), " "));
                 }
             });
+            LOG.lifecycle("  External process [{}] completed.", execCommand.getLoggingPrefix());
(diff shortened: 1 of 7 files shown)


src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
```diff
@@ -31,6 +31,7 @@ public void execute(ExecSpec spec) {
                     LOG.lifecycle("  " + execCommand.getDescription() + ":\n    " + StringUtil.join(execCommand.getCommandLine(), " "));
                 }
             });
+            LOG.lifecycle("  External process {} completed.", execCommand.getLoggingPrefix().trim());
```
szpak: Btw, does Gradle log automatically the error message (and return code) in a case of failure?

src/main/groovy/org/shipkit/internal/gradle/release/CiReleasePlugin.java
```diff
@@ -58,6 +59,8 @@ public void execute(ShipkitExecTask task) {
                         "Preparing working copy for the release", asList("./gradlew", GitSetupPlugin.CI_RELEASE_PREPARE_TASK)));
                 task.getExecCommands().add(execCommand(
                         "Performing the release", asList("./gradlew", ReleasePlugin.PERFORM_RELEASE_TASK)));
+
+                TaskSuccessfulMessage.logOnSuccess(task, "  Release " + project.getVersion() + " was shipped! Thank you for using Shipkit!");
```
szpak: :)

src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
szczepiq: >Gradle log automatically the error message
What do you mean by 'error message'? When external process fails, there is no error message - there is error output that may or may not have had information. By default, Gradle logs to 'lifecycle' level all output from standard and error of the process. Note that in our case, we replace std output/error with stream implementation that emits a prefix. Hence, we are not strictly using the defaults :)

>(and return code)
Yes, Gradle tell the return code by default. In case of ShipkitExecTask, we overwrite the default exception and produce our own, which also includes the error code.

szczepiq:
Thank you very much for review!

PR #308, opened 2017-07-14
Public API tidy up pre 1.0

szczepiq:
This will be something I will focus in the next sometime. 1.0 public API polishing.

- elevated some tasks to public API
- pushed down some implementation away from public types
- updated Javadocs
- rename jobs for clarity and consistency
- refactoring, test code refactoring

Review by commit might be useful :)

Much more work is pending, there are lots of task types to be elevated up.


szczepiq:
Friendly ping :)

src/main/groovy/org/shipkit/internal/gradle/ShipkitJavaPlugin.java
```diff
@@ -20,6 +19,12 @@
  *     <li>{@link BintrayReleasePlugin}</li>
  *     <li>{@link TravisPlugin}</li>
  * </ul>
+ *
+ * Adds behavior:
+ * <ul>
+ *     <li>Applies {@link JavaBintrayPlugin} to all Java projects in a multi-project Gradle build
+ *          (all projects that use Gradle's "java" plugin).</li>
```
szpak: [`java-library`](https://docs.gradle.org/current/userguide/java_library_plugin.html) plugin would be also covered by that?

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+import org.shipkit.internal.notes.vcs.GitOriginRepoProvider
+import spock.lang.Specification
+
+class InitShipkitFileTest extends Specification {
+
+    @Rule TemporaryFolder tmp = new TemporaryFolder()
+
+    def "does not modify shipkit file if it already exists"() {
+        given:
+        def shipkitFile = tmp.newFile()
```
szpak: Why not just `File` instead of `def` (aka `Object`)? :)
(I know that you just renamed the file)

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
```diff
@@ -0,0 +1,70 @@
...
+        when:
+        InitShipkitFile.initShipkitFile(shipkitFile, tmp.root)
+
+        then:
+        shipkitFile.text == "foo"
+    }
+
+    def "uses fallback repo if git call fails"() {
+        def shipkitFile = tmp.newFile()
+        def repoProvider = Mock(GitOriginRepoProvider)
```
szpak: Maybe `GitOriginRepoProvider repoProvider = Mock()` ?

src/main/groovy/org/shipkit/internal/gradle/ShipkitJavaPlugin.java
szczepiq: Yes, "java-library" applies "java"

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
szczepiq: It's a groovy test, IDEA knows it's a File, 'def' is shorter :)

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
szczepiq: hmmm, I've tried that but the code is subjectively harder to read when some local variables use formal types and some not. Why do you push for formal types for local variables?

szczepiq:
Thank you very much for comprehensive review and interesting feedback!!!

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
szpak: Is it also more readable (easier for developers reading the code to get know it's `File`)?

src/test/groovy/org/shipkit/internal/gradle/init/tasks/InitShipkitFileTest.groovy
szpak: I personally prefer to have all variables defined with their formal types. It usually easier to read and understand. Your syntax was preferred in up to Spock 1.0 as in the one suggested by me there was a type warning in Idea. However, I tune it up in Spock 1.1 and it works fine with Idea, still be the same short (just type is where is should to be :) )

Issue #290, opened 2017-07-06
Javadoc shows 100 warnings and spoils the build log


mockitoguy:
Let's figure out what to do with 100 warnings from Javadoc that spoil the build log. Can we suppress the warnings while still show Javadoc errors? Alternatively, we could just fix the warnings (not ideal, the warnings don't seem to be useful) + treat warnings as errors so that we will not forget to clean them up in the future.

Steps to reproduce: ```./gradlew cleanJavadoc javadoc```

Build log contains 100 warnings such as:

```
/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/notes/format/MultiReleaseNotesFormatter.java:15: warning: no @param for data
    String formatReleaseNotes(Collection<ReleaseNotesData> data);
           ^
/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/notes/format/MultiReleaseNotesFormatter.java:15: warning: no @return
    String formatReleaseNotes(Collection<ReleaseNotesData> data);
           ^
/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/notes/format/ReleaseNotesFormatters.java:22: warning: no @return
    public static MultiReleaseNotesFormatter notableFormatter(String introductionText,
                                             ^
```

janpetryk:
Those warnings are there only if you use Java 8. Further info here: http://blog.joda.org/2014/02/turning-off-doclint-in-jdk-8-javadoc.html 

Adding option `Xdoclint:none` to gradle config (as mentioned in above link) fixes this issue. 

But if we suppress those warnings other ones emerge: 
`Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamMember.java:9: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getDevelopers()
/Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamMember.java:9: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getContributors()
/Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamParser.java:14: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getDevelopers()
`

Those warnings are actually useful, when fixed, those references create proper anchors in doc file and are clickable in IntelliJ providing easier navigation. 

My starting point would be adding 
[...]

PR #309, opened 2017-07-15
Remove javadoc warnings

janpetryk:
Suggested fix for issue #290


wwilk:
What about javadoc errors when using Xdoclint:none? Do they still appear in the log and break the build? I think they are valuable

janpetryk:
Those messages that appear when using Xdoclint:none are warnings and look like that: 
```
/Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamMember.java:9: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getDevelopers() 
/Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamMember.java:9: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getContributors() 
/Users/jpetryk/Projects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/team/TeamParser.java:14: warning - Tag @link: reference not found: ReleaseConfiguration.Team#getDevelopers()
```

They are indeed valuable so I fixed the issue that was causing those warnings. New warnings will not break the build though.

build.gradle
```diff
@@ -61,6 +61,16 @@ pluginBundle {
     }
 }
 
+
```
szczepiq: Can you add short code comment why we are doing this? It's very useful to grasp the implementation quickly, without navigating git history + github issue description.

src/main/groovy/org/shipkit/internal/gradle/contributors/AllContributorsFetcherTask.java
```diff
@@ -32,7 +32,7 @@
     @OutputFile private File outputFile;
 
     /**
-     * See {@link ReleaseConfiguration.GitHub#getApiUrl()}
+     * See {@link org.shipkit.gradle.ReleaseConfiguration.GitHub#getApiUrl()}
```
szczepiq: Thank you for fixing the link! I assume that this can also be fixed by adding a formal import.

build.gradle
szczepiq: Can you add short code comment why we are doing this? It's very useful to grasp the implementation quickly, without navigating git history + github issue description.

szczepiq:
Thank you! Looking forward to merging. Please take a look at the review I gave :)

wwilk:
I think it would be valuable to keep errors breaking the build. This way we always keep correct documentation. Who looks at warnings if build passes? Is there any way this could be done without all these irrelevant warnings? @szczepiq What do you think about it?

szczepiq:
>I think it would be valuable to keep errors breaking the build. This way we always keep correct documentation. Who looks at warnings if build passes? Is there any way this could be done without all these irrelevant warnings? @szczepiq What do you think about it?

Removing warnings from the build output is a good incremental step so I think we should merge the PR given code review feedback is applied.

I absolutely agree that ideally we show errors instead of warnings.

Does it answer your question?

janpetryk:
@szczepiq changes you've requested are applied.

janpetryk:
@szczepiq changes that you've requested have been applied.

szczepiq:
Thank you! Merging!

Issue #286, opened 2017-07-04
Refactor release notes generation to also work with git services other than github


epeee:
## Goal
Let's decouple release note generation from github and furthermore make it possible to use other git services.

## Steps
- [ ] move github related code from `org.shipkit.internal.notes.contributors` to `org.shipkit.internal.notes.contributors.github`
- [ ] introcude interface for fetching contributors  
- [ ] introduce `GithubPlugin` which will provide tasks for fetching contributor infos and release notes data from github
- [ ] `ReleaseNotesPlugin` should just take care about updating release notes and therefore should just create `updateReleaseNotes` task and configure it accordingly.

szczepiq:
This sound great! Thank you for looking into it! Feel free to submit PRs with some preliminary design / thoughts how to make it work.

PR #310, opened 2017-07-16
Refactor release notes generation GitHub

epeee:
very first steps (#286)


epeee:
hmm, will resolve conflicts later on today.

src/main/groovy/org/shipkit/internal/gradle/GithubPlugin.java
```diff
@@ -0,0 +1,16 @@
+package org.shipkit.internal.gradle;
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.shipkit.internal.gradle.contributors.github.GithubContributorsPlugin;
+import org.shipkit.internal.gradle.java.GithubPomContributorsPlugin;
+
+
+public class GithubPlugin implements Plugin<Project> {
```
szczepiq: When introducing new plugins / refactoring how existing plugins are used together, please update the Javadoc accordingly. See the Javadoc of surrounding plugins for reference. 1.0 is coming soon and we need the Javadoc to be comprehensive and accurate! :)

szczepiq: When introducing new plugins / refactoring how existing plugins are used together, please update the Javadoc accordingly. See the Javadoc of surrounding plugins for reference. 1.0 is coming soon and we need the Javadoc to be comprehensive and accurate! :)

src/main/groovy/org/shipkit/internal/gradle/contributors/ContributorsPlugin.java
```diff
@@ -16,7 +16,7 @@
  *     <li>fetchAllContributors - {@link AllContributorsFetcherTask}</li>
  * </ul>
  */
-public class ContributorsPlugin implements Plugin<Project> {
+public abstract class ContributorsPlugin implements Plugin<Project> {
```
szczepiq: Can you explain in the description of the PR what is the motivation to create abstract classes? Abstract classes make the hierarchy complicated and usually it is better to replace the type hierarchy with composition. Plugin hierarchy is also harder to document. Can you try making the implementation use composition (e.g. one plugin applies another plugin)? Keeping the code simple! :)

We can also play along and see what happens - e.g. how your next PR demonstrates the need for plugin type hierarchies.

src/main/groovy/org/shipkit/internal/gradle/GithubPlugin.java
```diff
@@ -0,0 +1,16 @@
...
+
+    @Override
+    public void apply(Project project) {
```
szczepiq: This looks good. Plugin that wraps other plugins for ease of use.

epeee:
@szczepiq thx for your feedback!
rebasing and applying your suggestions now.

src/main/groovy/org/shipkit/gradle/notes/FetchContributorsTask.java
```diff
@@ -25,9 +26,15 @@
     @Input private String readOnlyAuthToken;
     @OutputFile private File outputFile;
 
+    private ContributorsProvider contributorsProvider;
+
+    public void setContributorsProvider(ContributorsProvider contributorsProvider) {
```
szczepiq: Let's avoid adding this method to the public API because it leaks internal "ContributorsProvider" type. I scrutinize all changes to public API before 1.0 release. If you want, you can put "contributorsProvider" as a field in the "FetchContributors" class. However, I don't think you really need this. When you implement new kind of ContributorsProvider, you can just create a new task type, for example "GitLabContributorsTask" (using GitLab as an example). In the implementation of the task you can use new ContributorsProvider and just reuse serialization logic.

I'm trying to avoid making the public API leaking internal API (or making the public API expose some complicated API).

Hope it makes sense :) Kind of hard to explain in text!

src/main/groovy/org/shipkit/internal/gradle/GithubPlugin.java
```diff
@@ -0,0 +1,27 @@
...
+ * <p>
+ * Applies following plugins:
+ *
+ * <ul>
+ *     <li>{@link GithubContributorsPlugin}</li>
+ *     <li>{@link GithubPomContributorsPlugin}</li>
+ * </ul>
+ *
+ */
+public class GithubPlugin implements Plugin<Project> {
```
szczepiq: Thank you for adding documentation!

src/main/groovy/org/shipkit/gradle/notes/FetchContributorsTask.java
epeee: k, thx.
I will change the implementation.

szczepiq:
Let us know when you're ready for the next round of review. Thank you for working on this!

epeee:
@szczepiq will rebase later on today, but other than that I think I'm ready.

src/main/groovy/org/shipkit/gradle/notes/GithubContributorsTask.java
```diff
@@ -19,21 +20,16 @@
  * https://developer.github.com/v3/repos/commits/
  * This way, we can also fetch the most recent contributors, necessary for correct release notes information.
  */
-public class FetchContributorsTask extends DefaultTask {
+public class GithubContributorsTask extends DefaultTask {
```
szczepiq: Name feedback:
- The convention for task names that we currently have in Shipkit is that verb comes first. It's not great convention but that's how we do it currently so keeping it consistent will make the code easier to navigate :)
- We use "GitHub" instead of "Github" in all other places in code.

I suggest to use "FetchGitHubContributorsTask" name instead "GithubContributorsTask".

src/main/groovy/org/shipkit/internal/gradle/notes/ReleaseNotesPlugin.java
```diff
@@ -106,6 +104,6 @@ private static void configureDetailedNotes(final UpdateReleaseNotesTask task,
         task.setPreviousVersion(project.getExtensions().getByType(VersionInfo.class).getPreviousVersion());
 
         task.setReleaseNotesData(releaseNotesFetcher.getOutputFile());
-        task.setContributorsDataFile(contributorsFetcher.getOutputFile());
+        task.setContributorsDataFile(contributorsFetcher.getOutputs().getFiles().getSingleFile());
```
szczepiq: Is there any particular reason you use "outputs.files.singleFile" instead of "outputFile"? I think the latter is more conventional and simpler.

epeee: I changed `contributorsFetcher` to type `org.gradle.api.Task` since we might integrate more systems in the future (and furthermore the output is available via the gradle api).
You are right, using `outputFile` is way simpler.

PR #311, opened 2017-07-16
Another round of polishing public API before 1.0

szczepiq:
The main goal is to polish the public API, specifically the task API before public 1.0.

- renames, package rework
- pushing out complexity out of tasks to internal impl classes
- test refactoring for simplicity
- removed dead code
- added documentation
- refactorings for consistency

Review per commit might be useful.

Thank you very much in advance for review!


src/test/groovy/org/shipkit/internal/gradle/init/InitVersioningTaskTest.groovy
```diff
@@ -3,6 +3,7 @@ package org.shipkit.internal.gradle.init
 import org.gradle.testfixtures.ProjectBuilder
 import org.junit.Rule
 import org.junit.rules.TemporaryFolder
+import org.shipkit.gradle.init.InitVersioningTask
```
wwilk: Please move the test to the same package as the corresponding class.

szczepiq:
>Awesome! org.shipkit.gradle looks nice and clean :) Maybe ReleaseNeededTask can be moved to "java" package?

Possibly, in the next round of 1.0 polishing, I'll take a look :)

Thank you for review!

PR #312, opened 2017-07-16
Refactoring of VersionUpgradeConsumerPlugin #302

wwilk:
As discussed by email and in the previous pull request #304 I applied a few changes to VersionUpgradeConsumerPlugin: 

- Choosing between env variable and project property for gitHubWriteToken takes place in ReleaseConfiguration now
- I created an extension VersionUpgrade that contains all of the configuration properties shared between tasks of VersionUpgradeConsumerPlugin
- I removed LazyConfig usages in the plugin but in a few cases I had to replace them with DeferredConfig, because some of the properties can be set by user after aplying the plugin.
- Javadoc improved
- CreatePullRequestTask logic extracted into another class, similarly as with ReleaseNeededTask
- DryRun can be used with the tasks of the plugin
- Integration test for the order of tasks in the plugin


wwilk:
See the intended use in javadoc of VersionUpgradeConsumerPlugin

src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -198,8 +207,18 @@ public void setReadOnlyAuthToken(String token) {
          * if this value is not specified.
          */
         public String getWriteAuthToken() {
```
szczepiq: I suggest a simpler approach for testing this. Instead of adding @ExposedOnlyForTesting setter, we can just create a new package-protected method like:

```
String getWriteAuthToken(EnvVariables envVariables)
//or:
String getWriteAuthToken(String envVarValue)
```

Then, we can make the public ```getWriteAuthToken()``` delegate to the new package-protected method. We would only test package-protected method.

I don't think we necessarily have to use "ExposedForTesting" for package protected methods - they are not really exposed. They are package-private! :)

src/main/groovy/org/shipkit/internal/gradle/VersionUpgrade.java
```diff
@@ -0,0 +1,84 @@
+package org.shipkit.internal.gradle;
+
+import java.io.File;
+
+/**
+ * Configuration of {@link VersionUpgradeConsumerPlugin}
```
szczepiq: 1. Let's start adding stuff to new package, for example "upgrades"

org.shipkit.internal.gradle.upgrades.ConsumerVersionUpgradePlugin
org.shipkit.internal.gradle.upgrades.ConsumerVersionUpgrade

2. I'm wondering if we should

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
+ *     <li>commitVersionUpgrade - commits replaced version</li>
+ *     <li>pushVersionUpgrade - pushes the commit to the version branch</li>
+ *     <li>createPullRequest - creates a pull request between base and version branches</li>
  *     <li>performVersionUpgrade - task aggregating all of the above</li>
  * </ul>
  *
  * Plugin should be used in client projects that want to have automated version upgrades of some other dependency, that use the producer version of this plugin.
  * Project with the producer plugin applied would then clone a fork of client project and run './gradlew performVersionUpgrade -PdependencyNewVersion=${VERSION}' on it.
+ *
+ * Example of plugin usage:
```
szczepiq: Fantastic Javadoc - very easy for users to grasp!

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
         final ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
 
-        TaskMaker.task(project, VERSION_UPGRADE_CHECKOUT_BASE_BRANCH, GitCheckOutTask.class, new Action<GitCheckOutTask>() {
+        versionUpgrade = project.getExtensions().create("versionUpgrade", VersionUpgrade.class);
+
+        // set defaults
+        versionUpgrade.setBuildFile(project.file("build.gradle"));
+        versionUpgrade.setBaseBranch("master");
+
+        String dependencyNewVersion = (String) project.getProperties().get(DEPENDENCY_NEW_VERSION);
```
szczepiq: FYI: you can also use 'project.findProperty' method.

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
+        if(dependencyNewVersion != null) {
```
szczepiq: I suggest to push out all that complexity elsewhere, for ease of testing. Perhaps to the DependencyNewVersionParser?

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
-    }
+    private VersionUpgrade versionUpgrade;
 
     @Override
     public void apply(final Project project) {
         LOG.lifecycle("Applying VersionUpgradeConsumerPlugin, beware that it's is INCUBATING state, so its API may change!");
         final ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
 
-        TaskMaker.task(project, VERSION_UPGRADE_CHECKOUT_BASE_BRANCH, GitCheckOutTask.class, new Action<GitCheckOutTask>() {
+        versionUpgrade = project.getExtensions().create("versionUpgrade", VersionUpgrade.class);
```
szczepiq: Since more extensions are probably coming, what do think about creating some semantics / conventions around them. For example:
- plugin id "version-upgrade-customer", extension id "version-upgrade-customer", extension class name "VersionUpgradeCustomerExtension"

No need to change it at this point given that the plugin is incubating - food for thought. We would do the same with "shipkit" extension and rename "ReleaseConfiguration" to "ShipkitExtension".

Strong conventions will make the code consistent, easy to work with and easy to use by end users.

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
+ *      apply plugin: 'org.shipkit.version-upgrade-consumer'
+ *
+ *      versionUpgrade{
+ *          baseBranch = 'release/2.x'
+ *          buildFile = file('gradle.properties')
+ *      }
+ *
+ * and then call it:
+ *
+ * ./gradlew performVersionUpgrade -PdependencyNewVersion=org.shipkit:shipkit:1.2.3
```
szczepiq: So far, we have been using short names for project properties ("dryRun", "preview") for ease of use. I suggest to call this project property in similar fashion, e.g. "dependency".

src/main/groovy/org/shipkit/internal/gradle/configuration/DeferredConfiguration.java
```diff
@@ -45,4 +51,11 @@ public void execute(Project project) {
             }
         });
     }
+
+    @ExposedForTesting
+    public static void forceConfiguration(Project project){
```
szczepiq: So far, we have been using ```project.evaluate()``` to simulate this. Why do we need forceConfiguration?

src/main/groovy/org/shipkit/internal/gradle/git/GitPush.java
```diff
@@ -55,19 +55,6 @@ static void setPushUrl(GitPushTask pushTask, String writeTokenEnvValue, String g
         }
     }
 
-    public static String getWriteToken(ReleaseConfiguration conf, String writeTokenEnvValue) {
-        String token = conf.getLenient().getGitHub().getWriteAuthToken();
```
szczepiq: It looks like we may no longer need ```getLenient()``` method on ReleaseConfiguration. Given that it is a public type and hence a liability can you deprecate ```getLenient()``` method? This way it will force us to figure out what to do with it later on :)

src/main/groovy/org/shipkit/internal/gradle/git/GitPush.java
```diff
@@ -55,19 +55,6 @@ static void setPushUrl(GitPushTask pushTask, String writeTokenEnvValue, String g
...
-        if (token != null) {
-            LOG.lifecycle("  'git push' uses GitHub write token specified in shipkit configuration.");
```
szczepiq: It was kind of nice that the build log indicated how we are doing git push - with the auth token or without it. It gives the user understanding what's going on, reduces "magic" :)

We can move on for now, not a blocker!

src/test/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPluginTest.groovy
```diff
@@ -1,139 +1,133 @@
...
 import org.shipkit.gradle.git.GitPushTask
+import org.shipkit.internal.gradle.configuration.DeferredConfiguration
+import org.shipkit.internal.gradle.git.GitCheckOutTask
 import org.shipkit.internal.gradle.versionupgrade.CreatePullRequestTask
-import org.shipkit.internal.gradle.configuration.LazyConfiguration
-import org.shipkit.internal.util.EnvVariables
+import org.shipkit.internal.gradle.versionupgrade.ReplaceVersionTask
 import testutil.PluginSpecification
 
 class VersionUpgradeConsumerPluginTest extends PluginSpecification {
```
szczepiq: Very nice and comprehensive test coverage. Thanks a bunch!!!

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -0,0 +1,35 @@
...
+            LOG.lifecycle("  Skipping pull request creation due to dryRun = true");
+            return;
+        }
+
+        LOG.lifecycle("  Creating a pull request of title '{}' in repository '{}' between base = '{}' and head = '{}'.",
+            task.getTitle(), task.getRepositoryUrl(), task.getBaseBranch(), task.getHeadBranch());
+
+        String body = "{" +
+            "  \"title\": \"" + task.getTitle() + "\"," +
+            "  \"body\": \"Please pull this in!\"," +
```
szczepiq: Might be cool to write some more, something like: "This pull request was automatically created by Shipkit's "version-upgrade-customer" Gradle plugin (http://shipkit.org). Please merge it so that you are using fresh version of "xxx" dependency."

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
```diff
@@ -9,170 +9,164 @@
...
+
+        String dependencyNewVersion = (String) project.getProperties().get(DEPENDENCY_NEW_VERSION);
+        if(dependencyNewVersion != null) {
+            DependencyNewVersionParser dependencyNewVersionParser = new DependencyNewVersionParser(dependencyNewVersion);
+            if (dependencyNewVersionParser.isValid()) {
+                versionUpgrade.setDependencyGroup(dependencyNewVersionParser.getDependencyGroup());
+                versionUpgrade.setDependencyName(dependencyNewVersionParser.getDependencyName());
+                versionUpgrade.setNewVersion(dependencyNewVersionParser.getNewVersion());
+            } else {
+                throw new IllegalArgumentException("");
```
szczepiq: Friendly reminder :)

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
wwilk: OK, I'll do it!

src/main/groovy/org/shipkit/internal/gradle/VersionUpgrade.java
wwilk: I think you didn't finish the second comment? About the first one - there is already a package like that, I will move all the classes there.

src/main/groovy/org/shipkit/internal/gradle/VersionUpgradeConsumerPlugin.java
wwilk: Thanks!

wwilk:
Merging now, I will address issues related to GitPush and Lenient config in a separate pull request. Other suggestions applied.

PR #313, opened 2017-07-19
GitAuthPlugin introduced #302

wwilk:
This is not yet ready, with missing tests and documentation. I wanted to present the idea before cleaning everything up.

The idea is to push the complexity of creating repository urls with authentication to a separate plugin. This way:
- we have it all in one place and we can share it between all gitPush , gitPull and possibly other tasks
- it's only initialized once, so we can have this nice message "git push does not use writeToken because it is not specified" or otherwise

I also introduced terms like "configRepositoryUrl" (taken from ReleaseConfig) and "originRepositoryUrl'  (taken from git remote origin), since in case of version-upgrade-consumer, we will run these tasks on fork of repo, so we can't use the repository specified in ReleaseConfiguration. 

I tested it on my fork of shipkit-example and the whole version upgrade works. See https://github.com/wwilk/shipkit-example/pull/1


src/main/groovy/org/shipkit/internal/gradle/git/GitAuthPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+    @Override
+    public void apply(Project project) {
+        ReleaseConfiguration conf = project.getPlugins().apply(ReleaseConfigurationPlugin.class).getConfiguration();
+        String ghUser = conf.getGitHub().getWriteAuthUser();
+        String writeToken = conf.getGitHub().getWriteAuthToken();
+
+        String configUrl;
+        String originUrl;
+        String secretValue = null;
+        String originRepositoryName = getOriginRepositoryName(project);
```
szczepiq: This method triggers an external process so we need to do it in the execution time.

There is no easy way to model this in Gradle. I suggest to try out the approach I did with the IdentifyGitBranchTask. Perhaps a new task IdentifyGitOriginTask? If you find a better solution - let me know!

src/main/groovy/org/shipkit/internal/gradle/git/GitPullTask.java
```diff
@@ -0,0 +1,50 @@
...
+public class GitPullTask extends DefaultTask{
+
+    @Input private String url;
+    @Input private String rev;
+    @Input private boolean dryRun;
+    private String secretValue;
+
+    @TaskAction
+    public void gitPull(){
+        new GitPull().gitPull(this);
```
szczepiq: Very nice, thank you!!! We push the internal implementation out of the public type.

src/main/groovy/org/shipkit/internal/gradle/git/GitAuthPlugin.java
```diff
@@ -0,0 +1,113 @@
...
+        return MessageFormat.format("https://github.com/{0}.git", ghRepo);
+    }
+
+    private String getOriginRepositoryName(Project project) {
+        DefaultProcessRunner runner = new DefaultProcessRunner(project.getProjectDir());
+        GitOriginRepoProvider repoProvider = new GitOriginRepoProvider(runner);
+        try{
+            return repoProvider.getOriginGitRepo();
+        } catch(Exception e){
+            LOG.debug("Failed to get local git remote origin", e);
```
szczepiq: I would probably do "info" level message and tweak it a bit (we should not use keywords like "failed" or "error" at the info level - they are very confusing when there are other real errors). What is the consequence if the origin git repo is null? Will the task fail saying that @Input is not satisfied? The best developer experience is when failures are easy to figure out :)

src/main/groovy/org/shipkit/internal/gradle/git/GitAuthPlugin.java
wwilk: That makes it a little hard, because everywhere you use you have to take it lazily, so if you want to create a repositoryUrl out of repositoryName taken from remote origin, you have to do this concatenation inside taskAction. But yeah I think we can do it like with IdentifyGitBranchTask and throw exception if accessed too soon.

szczepiq: I know it is tricky. Gradle does not have a good model for this. Think about it, perhaps you will come to some interesting solution.

We could also make the IdentifyX task drop a file with information. Other tasks would read the file.

wwilk:
I added pullUpstream task because we need it to do versionUpgrade on fork. Also I encapsulated all of the logic related to remote origin in GitRemoteOriginPlugin.

wwilk:
I tested it on initShipkit task, with remote origin set and null. I will test it on performVersionUpgrade task when it's released and fix bugs in case there are, since versionUpgrade is ongoing work and I don't want to break shipkit-example project. I would have to break it since I would need to use  nonexistent version (my local one).

src/main/groovy/org/shipkit/internal/gradle/git/GitPullTask.java
```diff
@@ -0,0 +1,50 @@
+package org.shipkit.internal.gradle.git;
+
+import org.gradle.api.DefaultTask;
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+
+public class GitPullTask extends DefaultTask{
```
szczepiq: Although we would keep the task internal for now (in internal package), let's add javadoc soon so that it's easier and faster to move it to public space when we are ready.

src/main/groovy/org/shipkit/internal/gradle/git/GitRemoteOriginPlugin.java
```diff
@@ -0,0 +1,99 @@
...
+
+    /**
+     * Configures some task that needs git origin information
+     * and an action that is executed when the git origin is available.
+     * This information is not available in configuration phase.
+     *
+     * @param needsOrigin some task that needs git origin information. Necessary 'dependsOn' will be added.
+     * @param resultHandler executed when git origin info is ready. Hooked up as 'doLast' action.
+     */
+    public void provideOriginTo(final Task needsOrigin, final ResultHandler<GitOriginAuth> resultHandler) {
```
szczepiq: This is very nice way of handling the use case in Gradle.

src/main/groovy/org/shipkit/internal/notes/vcs/IdentifyGitOriginRepoTask.java
```diff
@@ -0,0 +1,72 @@
+package org.shipkit.internal.notes.vcs;
```
wwilk: Right, I'll move it!

wwilk:
Suggestions applied and merging.

Issue #291, opened 2017-07-06
Package and rename job tidy-up


szczepiq:
Suggested rename jobs / package work:

- ReleaseConfiguration -> ShipkitConfiguration
- BumpVersionFileTask -> versioning.BumpVersionFileTask
- ReleaseNeededTask -> release.ReleaseNeededTask
- ReleaseNotesFetcherTask -> notes.ReleaseNotesFetcherTask
- UpdateReleaseNotesTask -> notes.UpdateReleaseNotesTask
- InitTravisTask -> travis.InitTravisTask

We should move classes that currently reside directly under “internal.gradle” to appropriate subpackages (and create subpackages if necessary :)

szczepiq:
Most of that is complete. The reset we will fix as we go!

PR #316, opened 2017-07-25
Public API clean up before 1.0

szczepiq:
Detailed information is in the commit messages, it might be easier to review per commit.

- moved types to better packages
- pushed out implementation to the internal types
- documentation update
- moved logic to more appropriate types

It should fix most of #291


szczepiq:
Friendly ping :)

szczepiq:
Thank you for review! Merging!

PR #317, opened 2017-07-25
Fixes #315. Returning null instead of throwing exception when ghWriteToken not specified

### LINKED ISSUE #315: master branch does not build without env variables
At the moment, master branch does not build without env variables and some tests fail in IDEA. Let's fix it. It's related to @wwilk's changes so I'm assigning this ticket to him :) We can also revert temporarily to fix the build.

wwilk:
I can't think of a better solution for now. Maybe I'll figure it out when I'm working on https://github.com/mockito/shipkit/pull/313

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
+++ b/src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
@@ -1,6 +1,8 @@
 package org.shipkit.gradle;
 
 import org.gradle.api.GradleException;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
 import org.shipkit.internal.util.EnvVariables;
 import org.shipkit.internal.util.ExposedForTesting;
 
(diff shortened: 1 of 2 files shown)


src/main/groovy/org/shipkit/gradle/ReleaseConfiguration.java
```diff
@@ -218,11 +222,12 @@ public String getWriteAuthToken() {
             if(envVar != null){
                 return envVar;
             }
-            throw new GradleException("Please export 'GH_WRITE_TOKEN' env variable first!\n" +
-                    "  The value of that variable is automatically used by Shipkit.\n" +
-                    "  It is highly recommended to keep write token secure and store env variable with your CI configuration.\n" +
-                    "  Alternatively, you can configure the write token explicitly:\n" +
-                    "    shipkit.gitHub.writeAuthToken = 'secret'");
+            LOG.lifecycle("  BEWARE! Shipkit.gitHub.writeAuthToken is unspecified.\n" +
```
szczepiq: Let's check this in so that the build is happy. However, we should find a better solution very soon. I suspect that now the build will show this hearty warning with every invocation for most normal builds. It gives ungraceful experience to the end users of Shipkit (we will have lots of them soon :).

szczepiq:
I'm merging, please consider my review feedback!

PR #318, opened 2017-07-30
Public API polish before 1.0

szczepiq:
Tweaks to the documentation, moved classes to better packages.


szczepiq:
Thank you for review!

PR #324, opened 2017-08-04
Public API clean up

szczepiq:
- pushed complexity out of public type to an internal type (ShipkitConfiguration -> ShipkitConfigurationStore, GitCommitTask -> GitCommitImpl)
 - moved GitCommitTask to the public API and made it more robust, easier to use. Updated documentation and tests.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/configuration/ShipkitConfigurationStore.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/configuration/ShipkitConfigurationStore.java
@@ -1,0 +1,81 @@
+package org.shipkit.internal.gradle.configuration;
+
+import org.gradle.api.GradleException;
+import org.shipkit.internal.util.EnvVariables;
+
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Map;
+
+public class ShipkitConfigurationStore {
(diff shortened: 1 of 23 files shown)


src/main/groovy/org/shipkit/internal/gradle/configuration/ShipkitConfigurationStore.java
```diff
@@ -0,0 +1,81 @@
...
+            if (value != null) {
+                return value;
+            }
+        }
+
+        if (lenient) {
+            return null;
+        }
+
+        throw new GradleException(message);
```
wwilk: This will throw the exception in case of gitPushWriteToken when env_variable and config one are missing, right? We changed this behaviour a few days ago to only log the message and return null.

wwilk: OK, I saw below that all accesses to this token use lenient configuration

szczepiq: Yeah, exactly!

szczepiq:
Thank you for review! Merging...

PR #325, opened 2017-08-04
Plugin-publish-plugin version upgraded

wwilk:
Since 0.9.14 (inclusive) shipkit upload to gradle plugins doesn't work. It seems that there may be an issue with plugin-publish-plugin version, according to https://discuss.gradle.org/t/published-plugin-not-found-404/23565.


wwilk:
Merging now to see if it shows why upload fails.

PR #329, opened 2017-08-06
Dogfooded version of Shipkit bumped

wwilk:


wwilk:
Merging to see if it helps with publishing plugins.

PR #330, opened 2017-08-06
Travis logging level decreased

wwilk:
I increased it to debug the problem with releasing. It's already fixed so I'm decreasing it back.


wwilk:
Merging because it's a trivial change

PR #332, opened 2017-08-06
VersionUpgradeProducerPlugin introduced #302

wwilk:
Introduced a plugin for producers of versionUpgrade functionality.

I tested it e2e on Shipkit as the producer and shipkit-example as the consumer. It worked and created a correct pull request in shipkit-example repo.


src/main/groovy/org/shipkit/internal/gradle/versionupgrade/VersionUpgradeConsumerPlugin.java
```diff
@@ -41,7 +41,7 @@
  * </ul>
  *
  * Plugin should be used in client projects that want to have automated version upgrades of some other dependency, that use the producer version of this plugin.
- * Project with the producer plugin applied would then clone a fork of client project and run './gradlew performVersionUpgrade -PdependencyNewVersion=${VERSION}' on it.
+ * Project with the producer plugin applied would then clone a fork of client project and run './gradlew performVersionUpgrade -Pdependency=${group:name:version}' on it.
```
szczepiq: Nice! Much simpler interface for the user!!!

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/VersionUpgradeProducerPlugin.java
```diff
@@ -0,0 +1,125 @@
...
+        });
+    }
+
+    private Task createProduceUpgradeTask(final Project project, final String consumerRepository){
+        return TaskMaker.execTask(project, "produceVersionUpgrade" + cleanRepositoryName(consumerRepository), new Action<Exec>() {
+            @Override
+            public void execute(final Exec task) {
+                task.setDescription("Performs dependency upgrade in " + consumerRepository);
+                task.setWorkingDir(getConsumerRepoTempDir(project, consumerRepository));
+                task.commandLine("./gradlew", "performVersionUpgrade", getDependencyProperty(project));
```
wwilk: Yeah, version-upgrade-consumer. Do you think we can do better about error message, when plugin is not applied, than "Task :performVersionUpgrade not found"?

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/VersionUpgradeProducerPlugin.java
```diff
@@ -0,0 +1,125 @@
...
+        return new File(project.getBuildDir().getAbsolutePath() + "/" + cleanRepositoryName(consumerRepository));
+    }
+
+    private String getDependencyProperty(Project project){
+        VersionInfo info = project.getExtensions().getByType(VersionInfo.class);
+        return String.format("-Pdependency=%s:%s:%s", project.getGroup().toString(), project.getName(), info.getPreviousVersion());
+    }
+
+    private String cleanRepositoryName(String repository){
+        return "_" + repository.replace('/', '_');
```
wwilk: Sure I will do that.

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/VersionUpgradeProducerPlugin.java
```diff
@@ -0,0 +1,125 @@
...
+ * <ul>
+ *     <li>{@link ShipkitConfigurationPlugin}</li>
+ * </ul>
+ *
+ * and adds following tasks:
+ *
+ * <ul>
+ *     <li>cloneConsumerRepo{consumerRepository} - clones consumer repository into temporary directory</li>
+ *     <li>produceVersionUpgrade{consumerRepository} - runs task performVersionUpgrade on consumerRepository</li>
+ *     <li>produceVersionUpgrade - task aggregating all of the produceVersionUpgrade{consumerRepository} tasks</li>
```
wwilk: If you run it without "clean" it will fail on clone, since the temporary directory won't be empty. With clean if you run it and the version of project doesn't change it will fail on "git commit" since the file with dependency (build.gradle) will not change and there will be nothing to commit. It's won't be clear to the user what happened, so I'm thinking about checking if there was any actual change made, and configure subsequent tasks (git commit, push, create pull request) with .onlyIf condition based on it. What do you think about it?

wwilk:
I applied all suggestions so merging.

PR #335, opened 2017-08-08
Public API tidy-up before 1.0

szczepiq:
Made public API tidy-up. The list of plugins I plan to tidy up is shorter every day. Exciting! :)

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/git/IdentifyGitBranchTask.java
+++ b/src/main/groovy/org/shipkit/gradle/git/IdentifyGitBranchTask.java
@@ -2,10 +2,8 @@ package org.shipkit.gradle.git;
 
 import org.gradle.api.DefaultTask;
 import org.gradle.api.GradleException;
-import org.gradle.api.logging.Logger;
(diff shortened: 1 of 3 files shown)


wwilk:
@szczepiq IdentifyGitBranchTask may be easy to misuse since branch is set at the specific time. The recommended usage would be by calling GitBranchPlugin#provideBranchTo. So maybe we can keep GitBranchPlugin public and the task internal? At least at the beginning.

szczepiq:
>@szczepiq IdentifyGitBranchTask may be easy to misuse since branch is set at the specific time. The recommended usage would be by calling GitBranchPlugin#provideBranchTo. So maybe we can keep GitBranchPlugin public and the task internal? At least at the beginning.

It's true that it can be misused. That's why we throw a meaningful exception message :) Instead of provideBranchTo, users can just set the task dependencies correctly. Given that we have actionable exception message and easy way to set this task correctly (task dependencies), I feel that we don't need to expose extra public API and the task implementation as public API.

Approved? :)

Issue #333, opened 2017-08-07
PowerMock CR:  customisable tag name.


thekingnothing:
It will be nice to have ability to customisable tag name for releases. For example, PowerMock uses another tag naming pattern: `powermock-`. It is not customisable at the current moment according to comment in https://github.com/powermock/powermock/issues/826 

> @thekingnothing Starting from the end, at the moment only thing about tag that you can customize is its prefix.
```
        git.setTagPrefix("v"); //so that tags are "v1.0", "v2.3.4"
```

szczepiq:
@thekingnothing, thank you for using Shipkit!!!

Can you try the tag prefix as suggested by @wwilk? It seems that the tags in powermock are versions prefixed with "powermock-".

This should do the trick:
```Gradle
shipkit {
  git.tagPrefix = "powermock-"
}
```

szczepiq:
@wwilk, it seems that in the code we use hardcoded "v" literal in few places, instead of using the tag prefix from configuration.

szczepiq:
After @epeee improvements, I think this one can be fixed.

PR #336, opened 2017-08-08
get rid of hardcoded tag prefix 'v'

epeee:
... the ones I found during a quick search. (#333)


epeee:
@wwilk I tested the release notes generation (which still looks good) and the change is pretty much the same for ShipkitBintrayPlugin. 
I will update shipkit-example after merging to make sure everything still does work.

szczepiq:
Thank you for fixing this! Powermock needs it ;)

PR #337, opened 2017-08-09
VersionUpgradeConsumerPlugin renamed to UpgradeDependencyPlugin

wwilk:
Additionally added documentation for GitPull and GitPullTask. Also I added ReplaceVersionTask.buildFileUpdated so that I can skip GitCommit, GitPush and CreatePullRequest tasks if version wasn't actually changed (newVersion == previousVersion). I tested it e2e on shipkit-example.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/ReplaceVersionTask.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/ReplaceVersionTask.java
@@ -19,6 +19,7 @@ public class ReplaceVersionTask extends DefaultTask{
     public static final String VERSION_REGEX = "[0-9.]+";
 
     private UpgradeDependencyExtension versionUpgrade;
+    private Boolean buildFileUpdated;
 
     @TaskAction
     public void replaceVersion(){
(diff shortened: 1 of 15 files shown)


src/main/groovy/org/shipkit/internal/gradle/versionupgrade/ReplaceVersionTask.java
```diff
@@ -30,14 +31,30 @@ public void replaceVersion(){
 
         String content = IOUtil.readFully(versionUpgrade.getBuildFile());
         String updatedContent = content.replaceAll(versionPattern, replacement);
-        IOUtil.writeFile(versionUpgrade.getBuildFile().getAbsoluteFile(), updatedContent);
+
+        buildFileUpdated = !content.equals(updatedContent);
+
+        if(buildFileUpdated) {
```
szczepiq: This is nice. Avoids touching file if not needed.

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -47,21 +48,21 @@
  *
  * Configure your 'shipkit.gradle' file like here:
  *
- *      apply plugin: 'org.shipkit.version-upgrade-consumer'
+ *      apply plugin: 'org.shipkit.upgrade-dependency'
  *
- *      versionUpgrade{
+ *      upgradeDependency{
  *          baseBranch = 'release/2.x'
  *          buildFile = file('gradle.properties')
```
szczepiq: Let's use 'build.gradle' in the Javadoc example instead of 'gradle.properties'. The latter is very unconventional and I plan to get rid of the from Mockito.

src/test/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPluginIntegTest.groovy
```diff
@@ -14,7 +14,7 @@ class VersionUpgradeConsumerPluginIntegTest extends GradleSpecification {
         """
 
         buildFile << """
-            apply plugin: "org.shipkit.version-upgrade-consumer"
+            apply plugin: "org.shipkit.upgrade-dependency"
```
szczepiq: This looks now way cleaner. Thanks!

PR #338, opened 2017-08-10
Public API tidy-up, logging improvements

szczepiq:
Tidied up the API before 1.0 release, replaced the use of Exec task with ShipkitExecTask to improve the logging offered to the end user.

The shortlist of public API tidy-up is getting really small :) Soon, we should be able to focus on documentation and adoption!


szczepiq:
Friendly ping :)

szczepiq:
Thank you for review!

PR #339, opened 2017-08-12
Fixed bug with releaseNeeded, replaced Exec use with ShipkitExecTask

szczepiq:
- fixed issue with releaseNeeded. Previously we were incorrectly failing the entire build when assertReleaseNeeded failed. Instead of failing the entire build, we should stop executing the task.
 - refactored remaining usages of Exec task to ShipkitExecTask (this improves logging when we fork of external processes). After 1.0, it would be hard to change the task types due to the compatibility contract.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/git/GitSetupPlugin.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/git/GitSetupPlugin.java
@@ -6,7 +6,6 @@ import org.gradle.api.Project;
 import org.gradle.api.Task;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
-import org.gradle.api.tasks.Exec;
 import org.gradle.process.ExecResult;
 import org.shipkit.gradle.configuration.ShipkitConfiguration;
 import org.shipkit.gradle.exec.ShipkitExecTask;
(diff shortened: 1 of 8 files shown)


szczepiq:
Friendly ping ;)

src/main/groovy/org/shipkit/internal/gradle/git/GitSetupPlugin.java
```diff
@@ -54,22 +57,20 @@
     public void apply(Project project) {
         final ShipkitConfiguration conf = project.getPlugins().apply(ShipkitConfigurationPlugin.class).getConfiguration();
 
-        TaskMaker.execTask(project, UNSHALLOW_TASK, new Action<Exec>() {
-            public void execute(final Exec t) {
+        TaskMaker.task(project, UNSHALLOW_TASK, ShipkitExecTask.class, new Action<ShipkitExecTask>() {
```
szczepiq: We could do that. I have not implemented it this way because the value is pretty small. Previously we had a factory method for Exec task because we needed to add an additional 'doFirst' action to all Exec tasks. With ShipkitExecTask, there is no need for this.

szczepiq:
Thank you for review!

PR #347, opened 2017-08-20
make use of shipkit v0.9.41

epeee:


szczepiq:
Nice! Thank you for bumping the version!

PR #348, opened 2017-08-20
Plugin upgrade-downstream applied in Shipkit #302

wwilk:
Tested locally, running "./gradlew upgradeDownstream" resulted in https://github.com/mockito/shipkit-example/pull/78

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,6 +23,3 @@ install:
 
 script:
   - ./gradlew build -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
-
-after_success:
-  - ./gradlew upgradeDownstream
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,3 +23,6 @@ install:
 
 script:
   - ./gradlew build -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
+
+after_success:
+  - ./gradlew upgradeDownstream
--- a/gradle/shipkit.gradle
+++ b/gradle/shipkit.gradle
@@ -10,3 +10,9 @@ afterEvaluate {
     //TODO let's create a property for this setting in 'shipkit' extension
     tasks.updateReleaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
 }
+
+apply plugin: 'org.shipkit.upgrade-downstream'
+
+upgradeDownstream{
+    repositories = ['mockito/shipkit-example']
+}
--- a/gradle/shipkit.gradle
+++ b/gradle/shipkit.gradle
@@ -10,9 +10,3 @@ afterEvaluate {
     //TODO let's create a property for this setting in 'shipkit' extension
     tasks.updateReleaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
 }
-
-apply plugin: 'org.shipkit.upgrade-downstream'
-
-upgradeDownstream{
-    repositories = ['mockito/shipkit-example']
-}
--- a/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
@@ -55,7 +55,7 @@ import static org.shipkit.internal.util.ArgumentValidation.notNull;
  */
 public class UpgradeDownstreamPlugin implements Plugin<Project> {
 
-    private static final String UPGRADE_DOWNSTREAM_TASK = "upgradeDownstream";
+    static final String UPGRADE_DOWNSTREAM_TASK = "upgradeDownstream";
     private UpgradeDownstreamExtension upgradeDownstreamExtension;
 
     @Override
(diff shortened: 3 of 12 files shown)


.travis.yml
```diff
@@ -23,3 +23,6 @@ install:
 
 script:
   - ./gradlew build -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
+
+after_success:
+  - ./gradlew upgradeDownstream
```
szczepiq: Does take under account "release needed" factor? If we're building from tag we don't want to upgrade downstream.

Perhaps this is as simple as adding this for now:

```
after_success:
  - ./gradlew assertReleaseNeeded && ./gradlew upgradeDownstream
```

It's a great idea to add this to "after_success". I love the fact that PRs are automatically created!!

gradle/shipkit.gradle
```diff
@@ -10,3 +10,9 @@ afterEvaluate {
     //TODO let's create a property for this setting in 'shipkit' extension
     tasks.updateReleaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
 }
+
+apply plugin: 'org.shipkit.upgrade-downstream'
+
+upgradeDownstream{
+    repositories = ['mockito/shipkit-example']
```
szczepiq: Let's add 'shipkit-bootstrap', too.

Later on if this works well for us, let's add Mockito. This will be awesome!!! Thanks a lot for working on this plugin!

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
```diff
@@ -50,7 +50,7 @@
  *
  * and then call:
  *
- * ./gradlew produceVersionUpgrade
+ * ./gradlew upgradeDownstream
```
szczepiq: hah, nice catch

gradle/shipkit.gradle
wwilk: Sure, I wanted to start with example to check if it works as expected and then add other projects

.travis.yml
wwilk: The thing is that "assertReleaseNeeded" is ran with "ciPerformRelease" too. So if it fails, everything is fine, it would fail again in "after_success" ("after_success" is run because "ciperformRelease" is set to stop execution if "assertReleaseNeeded" fails, without failing the build). But if it passes, release would be executed and the version changed and pushed. So if we check again in "after_success" we would fail on "assertReleaseNeeded" AFAIK. Maybe we can somehow pass information between executed tasks? Maybe through creating a file in the workspace, with information if "assertReleaseNeeded" failed or not?

wwilk:
I added "upgradeDownstream" to "ciPerformRelease" dynamically, as discussed with @szczepiq on Slack. Unfortunately it introduced a small temporal coupling, you have to apply "ci-release" plugin before applying "upgrade-downstream", to maintain the correct order of tasks in "ciPerformRelease". For now it should be OK, later we can think how to do it better, maybe we can add this execCommand in deferredConfig or introduce sth like doLast for ShipkitExecTask.

BTW merging this PR shouldn't have any effect before upgrading version of Shipkit, since it needs a new version of "upgrade-downstream" which will add the task to "ciPerformRelease"

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
```diff
@@ -63,13 +64,22 @@ public void apply(final Project project) {
...
             }
         });
 
+        ShipkitExecTask ciPerformReleaseTask = (ShipkitExecTask) project.getTasks().findByName(CiReleasePlugin.CI_PERFORM_RELEASE_TASK);
+
+        if(ciPerformReleaseTask != null){
+            ciPerformReleaseTask.getExecCommands().add(execCommand(
+                "Upgrading downstream projects", asList("./gradlew", UPGRADE_DOWNSTREAM_TASK)));
+        } else {
+            throw new IllegalStateException("Please apply org.shipkit.ci-release or org.shipkit.java plugin before applying upgrade-downstream.");
```
szczepiq: There is an alternative way of configuring this and it is also more idiomatic Gradle, avoiding complexity and the need for additional unit testing:

```Groovy
project.plugins.withType(CiReleasePlugin) {
   //here you know that ciPerformRelease task exists
}
```

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
```diff
@@ -63,13 +64,22 @@ public void apply(final Project project) {
...
 
-        final Task performAllUpdates = TaskMaker.task(project, "upgradeDownstream", new Action<Task>() {
+        final Task performAllUpdates = TaskMaker.task(project, UPGRADE_DOWNSTREAM_TASK, new Action<Task>() {
             @Override
             public void execute(final Task task) {
                 task.setDescription("Performs dependency upgrade in all downstream repositories.");
             }
         });
 
+        ShipkitExecTask ciPerformReleaseTask = (ShipkitExecTask) project.getTasks().findByName(CiReleasePlugin.CI_PERFORM_RELEASE_TASK);
```
szczepiq: We can also bust this plugin into 2:
 - upgrade-downstream - adds task, extension, etc.
 - ci-upgrade-downstream - "glues" upgrade-downstream with ci-release

This would offer better function decomposition and make plugins more usable in general

src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDownstreamPlugin.java
wwilk: I wanted to throw this exception, if there is no task, to tell the user what exactly the problem is. But with the split to "ci-upgrade-downstream" and "upgrade-downstream" that you suggested, it is not needed any more.

wwilk:

PR #349, opened 2017-08-21
Public API tidy-up before 1.0 ...

epeee:
...  plugin package.

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/plugin/PluginDiscoveryTask.java
+++ b/src/main/groovy/org/shipkit/gradle/plugin/PluginDiscoveryTask.java
@@ -7,6 +7,10 @@ import org.shipkit.internal.gradle.plugin.tasks.PluginDiscovery;
 
 /**
  * Discovers gradle plugins and adds them to the {@link PluginBundleExtension}.
+ * <p>
+ * This task will automatically pick up available gradle plugins (discovered via properties files in
+ * META-INF/gradle-plugins) during execution time and will configure the pluginBundle extension (provided via
+ * 'com.gradle.plugin-publish' plugin) accordingly.
  */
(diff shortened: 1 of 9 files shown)


wwilk:
Why only PluginDiscoveryTask is made public? It's just a first step or there is another reason for keeping PluginDiscoveryPlugin or ShipkitGradlePlugin internal?

src/main/groovy/org/shipkit/gradle/plugin/PluginDiscoveryTask.java
```diff
@@ -1,8 +1,13 @@
-package org.shipkit.internal.gradle.plugin;
+package org.shipkit.gradle.plugin;
 
+import com.gradle.publish.PluginBundleExtension;
 import org.gradle.api.DefaultTask;
 import org.gradle.api.tasks.TaskAction;
+import org.shipkit.internal.gradle.plugin.tasks.PluginDiscovery;
 
+/**
+ * Discovers gradle plugins and adds them to the {@link PluginBundleExtension}.
```
szczepiq: Can you write a little bit more how it discovers the plugins? I think it would be really useful for our users to understand how exactly it happens.

src/main/groovy/org/shipkit/gradle/plugin/PluginDiscoveryTask.java
```diff
@@ -1,8 +1,13 @@
...
+ */
 public class PluginDiscoveryTask extends DefaultTask {
```
szczepiq: Nice catch! This task should be public.

epeee:
thx for your feedback!

I just added some more javadoc.

@wwilk `PluginValidatorTask` needs some rework, therefore we are not moving it to public api yet.

PR #350, opened 2017-08-21
E2ETest renamed to DownstreamTest

wwilk:


settings.gradle
```diff
@@ -1,3 +1,3 @@
 rootProject.name="shipkit"
 
-include "e2eTests"
+include "downstreamTests"
```
szczepiq: Minor: I suggest to call it "downstreamTest" for consistency. Usually we use singular for such modules, for 'test' source set, etc. Alternatively, we just call the subproject "downstream" for simplicity

wwilk:
Renamed so merging

PR #351, opened 2017-08-22
Ci-upgrade-downstream enabled

wwilk:


wwilk:
Merging to check if ci-upgrade-downstream works correctly on Travis.

PR #357, opened 2017-08-23
Renamed plugin id from '*git-hub*' to 'github'

szczepiq:
'git-hub' looks off. Let's do this change now, before 1.0.


epeee:
It looks like the build is failing because a test is still using the old plugin id.
Let's also use the class in this case and plugin validation will do the other part in the future.
I will push a fix directly to your branch and merge afterwards.

PR #359, opened 2017-08-23
Downstream tests enabled

wwilk:


szczepiq:
This is awesome!!! Trying it out!

szczepiq:
I have some feedback about usability and logging, but that can be implemented later :)

Really good work getting this feature to completion, merging!

PR #361, opened 2017-08-24
Improve plugin validation

### LINKED ISSUE #356: Improve Gradle plugin validation
### Problem

Running ```./gradlew validatePlugins``` yields many false positives:

```
> no properties file found for plugin(s):
        'ShipkitJava' (/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/gradle/java/ShipkitJavaPlugin.java)
        'ShipkitConfiguration' (/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/gradle/configuration/ShipkitConfigurationPlugin.java)
        'ShipkitBintray' (/Users/sfaber/mockito/release/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java)        
```

- some plugins have slightly different naming convention, with optional 'Shipkit' prefix
- some plugins don't have plugin id intentionally

### Suggested approach

How about we change the implementation into something like this. Given example resource files:

```
implementation-class=org.shipkit.internal.gradle.java.GitHubPomContributorsPlugin
implementation-class=org.shipkit.internal.gradle.java.ShipkitJavaPlugin
```

For every resource file we: 

1. Ensure naming convention:
	a. convert plugin id to acceptable names, examples
		- "org.shipkit.bintray" -> "OrgShipkitBintrayPlugin", "ShipkitBintrayPlugin", "BintrayPlugin"
		- "org.shipkit.git-hub-pom-contributors" -> "OrgShipkitGitHubPomContributors", "ShipkitGitHubPomContributors", "GitHubPomContributors"
	b. implementation-class must match one of the acceptable names for given plugin id
        c. I would ignore case for simplicity
2. Validate if the implementation class exists
	org.shipkit.internal.gradle.java.GitHubPomContributorsPlugin -> file org/shipkit/internal/gradle/java/GitHubPomContributorsPlugin.java|groovy exists

epeee:
fixes  #356


epeee:
Good news, plugin validation does pass for shipkit project :-)
Anyway, if the validation fails it might look like:

```
:validatePlugins FAILED
:validatePlugins (Thread[Daemon worker Thread 16,5,main]) completed. Took 0.006 secs.

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':validatePlugins'.
> plugin validation failed for plugin(s):
        'org.shipkit.github-pom-contributors': Implementation class org.shipkit.internal.gradle.java.AnotherPlugin does not match one of the acceptable names [ContributorsPlugin, PomContributorsPlugin, GithubPomContributorsPlugin, ShipkitGithubPomContributorsPlugin, OrgShipkitGithubPomContributorsPlugin]
```

src/main/groovy/org/shipkit/gradle/plugin/PluginValidatorTask.java
```diff
@@ -0,0 +1,24 @@
...
+ * The following constraints are validated:
+ * <ul>
+ *     <li>the naming convention of a plugin:</li>
+ *     acceptable names are calculated based on the plugin id and the implementation class has to match one of them, e.g:
+ *     <br>
+ *     "org.shipkit.bintray" -> "OrgShipkitBintrayPlugin", "ShipkitBintrayPlugin", "BintrayPlugin"
+ *     <li>the implementation class specified in a plugin properties file exists</li>
+ * </ul>
+ */
+public class PluginValidatorTask extends DefaultTask {
```
wwilk: Thanks for clearly stating the constraints of validation!

src/main/groovy/org/shipkit/internal/util/PropertiesUtil.java
```diff
@@ -0,0 +1,24 @@
...
+public class PropertiesUtil {
+
+    public static Properties readProperties(File properties) {
+        Properties p = new Properties();
+        FileReader reader = null;
+        try {
+            reader = new FileReader(properties);
+            p.load(reader);
+        } catch (Exception e) {
+            throw new RuntimeException("Problems reading properties file: " + properties);
```
wwilk: It would be good to rethrow the original exception. Properties object won't give us much if it fails due to eg. nonexistent file.

wwilk:
I like this error message, it clearly says what is wrong and how to fix it!

src/main/groovy/org/shipkit/internal/gradle/plugin/tasks/PluginValidator.java
```diff
@@ -1,95 +1,120 @@
...
-import java.util.HashMap;
-import java.util.Map;
-import java.util.Set;
+import java.util.*;
 
+import static org.shipkit.internal.gradle.plugin.tasks.PluginUtil.getImplementationClass;
 import static org.shipkit.internal.gradle.plugin.tasks.PluginUtil.DOT_PROPERTIES;
 
 
 public class PluginValidator {
```
wwilk: No unit test for this class?

src/main/groovy/org/shipkit/internal/gradle/plugin/tasks/PluginValidator.java
```diff
@@ -1,95 +1,120 @@
...
-                sb.append(pluginName);
-                sb.append("' (");
-                sb.append(missingPropertiesFiles.get(pluginName));
-                sb.append(")");
-            }
-            throw new GradleException(sb.toString());
+    private List<String> getClassCandidates(String pluginId) {
+        List<String> candidates = new ArrayList<String>();
+
+        String[] candidate1 = pluginId.split("\\.|-");
```
wwilk: Maybe "pluginIdParts" instead of "candidate1"?

epeee:
@wwilk thx for your feedback!
I applied your suggestions and also added some unit tests for `PluginValidator`.

src/main/groovy/org/shipkit/gradle/plugin/PluginValidatorTask.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @TaskAction
+    public void validate() {
+        new PluginValidator().validate(getProject());
```
szczepiq: Nice! Thank you for following our current pattern to move out complexity from public types!

src/main/groovy/org/shipkit/internal/gradle/plugin/tasks/PluginDiscovery.java
```diff
@@ -27,7 +23,7 @@ public void discover(Project project) {
         for (File pluginPropertyFile : pluginPropertyFiles) {
             PluginConfig config = new PluginConfig(generatePluginName(pluginPropertyFile.getName()));
             config.setId(pluginPropertyFile.getName().substring(0, pluginPropertyFile.getName().lastIndexOf(DOT_PROPERTIES)));
-            config.setDisplayName(getImplementationClass(pluginPropertyFile));
+            config.setDisplayName(PluginUtil.getImplementationClass(pluginPropertyFile));
```
szczepiq: +1 for pushing complexity to separate objects for cleaner codebase!

src/main/groovy/org/shipkit/internal/gradle/plugin/tasks/PluginValidator.java
```diff
@@ -1,95 +1,129 @@
...
 
     private static final Logger LOG = Logging.getLogger(PluginValidator.class);
+    private static final String[] PLUGIN_EXTENSIONS = new String[]{".groovy", ".java"};
 
-    public void validate(Set<File> gradlePlugins, Set<File> gradleProperties) {
-        ensurePluginsHavePropertiesFile(gradlePlugins, gradleProperties);
+    private Set<File> sourceDirs;
+
+    public void validate(Project project) {
+        SourceSet sourceSet = JavaPluginUtil.getMainSourceSet(project);
```
szczepiq: In general, during the task execution, the task should not interrogate "project" object for the inputs. Instead, it's best if the task has a property that is set by the plugin. This way, the task is more robust, more useful, the user can reconfigure it easily.

You had it before - there was a "sourceSet" property on the task that you were setting in the plugin. That was really, good, any particular reason you removed it?

src/main/groovy/org/shipkit/internal/gradle/plugin/tasks/PluginValidator.java
szczepiq: @wwilk, you are probably right and more balanced coverage, with more explicit unit tests would drive slightly cleaner codebase.

PluginValidationPluginTest is sort of an integration test that covers this functionality. So we have the coverage, but necessarily as an explicit unit test to PluginValidator. I'd say that we move on, we don't necessarily have to have 1:1 class to test class in the codebase for every functionality. It's important that we have coverage. Coarse-grained tests are sometimes useful because refactorings are easier :) Let's move on and learn on the way if this way of writing tests works in our favor.

epeee:
@szczepiq thx a lot for your feedback! 
Makes perfectly sense, I just re-added the sourceset property to the task.

PR #366, opened 2017-08-26
make check task dependsOn validatePlugins task

epeee:


wwilk:
It would be good to set this dependency for Shipkit (when dogfooding), but I don't know if it should be the default for all users. In this case maybe we can inform the user that they can disable VALIDATE_PLUGINS task if they don't want to obey the validation users (naming rules are quite strict). Not sure about it, what do you think @epeee ?

epeee:
@wwilk yes, let's just use it for shipkit for now.

gradle/shipkit.gradle
```diff
@@ -9,4 +9,6 @@ shipkit {
 afterEvaluate {
     //TODO let's create a property for this setting in 'shipkit' extension
     tasks.updateReleaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
+
+    tasks.findByName(JavaBasePlugin.CHECK_TASK_NAME).dependsOn(validatePlugins)
```
szczepiq: Typically in Gradle build script file you would see:

```
check.dependsOn validatePlugins
//or:
tasks.check.dependsOn validatePlugins
```

This way it is shorter and easier to read. In the plugin code, in java, we would be more explicit and use 'findByName'.

Build script files and plugin code are both optimized for maintainability. In build.gradle file readability gives most maintainability whereas in plugin code, the best is explicit calls with compile-time safety.

Issue #360, opened 2017-08-24
Inconsistent project directory layout


szczepiq:
## Problem

Currently, the layout of subprojects in Shipkit is inconsistent because the subproject is a part of main project

```
- src
- downstreamTest
- build.gradle
- settings.gradle
```

## Suggested change

How about we make it consistent and rework Gradle subprojects to the following

```
- subprojects
   - shipkit
   - downstreamTest
- build.gradle
- settings.gradle
```

## Implementation hints

Check out "settings.gradle" file of Mockito project to see how "subprojects" can be implemented.

PR #381, opened 2017-09-04
set working dir for ShipkitExec

epeee:
e.g. if shipkit is applied to a subproject we currently can't find the gradlew.

> Caused by: java.io.IOException: Cannot run program "./gradlew" (in directory "~/shipkit/subprojects/shipkit"): error=2, No such file or directory

I thought it might be a good idea to use `rootDir` as a `workingDir` default.
I headed into it while working on #360 and applying shipkit just on a subproject.

### DIFF
--- a/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
+++ b/src/main/groovy/org/shipkit/internal/gradle/exec/ShipkitExec.java
@@ -16,7 +16,7 @@ public class ShipkitExec {
 
     private final static Logger LOG = Logging.getLogger(ShipkitExec.class);
 
-    public void execCommands(Collection<ExecCommand> execCommands, Project project) {
+    public void execCommands(Collection<ExecCommand> execCommands, final Project project) {
         for (final ExecCommand execCommand : execCommands) {
             ExecResult result = project.exec(new Action<ExecSpec>() {
                 @Override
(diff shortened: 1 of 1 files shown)


szczepiq:
>I headed into it while working on #360 and applying shipkit just on a subproject.

Shipkit is intended to be applied to the root project. Why do you need to apply it to a subproject? #360 ticket is about putting "code" to the submodules but still can keep shipkit plugin applied to the root project.

I think that Shipkit needs to be applied to root, there are likely other assumptions about this in the code :)

epeee:
It looks like `ShipkitGradlePlugin` only applies plugins and therefore configures certain tasks just to the current project. Therefore e.g. plugindiscovery won't work properly if we e.g. move shipkit src to subprojects/shipkit and apply ShipkitGradlePlugin just to the root project. 
ShipkitJava plugin in fact can handle subprojects.

szczepiq:
>It looks like ShipkitGradlePlugin only applies plugins and therefore configures certain tasks just to the current project.

You are right. It looks like ShipkitGradlePlugin has to be applied to root. Seems inflexible. Do you want to take a stab at it or this is too much of a deep Gradle change? Let me know how I can help!

szczepiq:
I reworked the code to enable submodule use case cleanly: #384 (I hope it works ;)

epeee:
@szczepiq thx, we can close this one now.

PR #383, opened 2017-09-05
expose release notes url

epeee:
in order to be able to log the release notes url (see ##364).

I already did some local testing:

>   Release shipped!
>     - Files in Bintray:   https://dl.bintray.com/shipkit-bootstrap/bootstrap/org/shipkit/bootstrap/ep/shipkit-bootstrap/0.0.3
>     - Bintray repository: https://bintray.com/shipkit-bootstrap/bootstrap/maven/0.0.3
>     - Release notes:      https://github.com/mockito/shipkit-bootstrap/blob/master/docs/release-notes.md

### DIFF
--- a/src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
+++ b/src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
@@ -46,7 +46,7 @@ public class UpdateReleaseNotesTask extends DefaultTask {
      */
     @TaskAction
     public void updateReleaseNotes() {
-        new UpdateReleaseNotes(new HeaderProvider()).updateReleaseNotes(this);
+        new UpdateReleaseNotes().updateReleaseNotes(this, new HeaderProvider());
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
```diff
@@ -129,6 +129,13 @@ public String getGitHubRepository() {
     }
 
     /**
+     * @return a link to the generated release notes file hosted on Github.
+     */
+    public String getReleaseNotesUrl(String branch) {
+        return  getGitHubUrl() + "/" + getGitHubRepository() + "/blob/" + branch + "/" + getProject().relativePath(getReleaseNotesFile());
```
szczepiq: Nice new property!

Let's push all the implementation to the "UpdateReleaseNotes" class. This way, the public API types have only a reason to change when public API changes!

src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
```diff
@@ -129,6 +129,13 @@ public String getGitHubRepository() {
     }
 
     /**
+     * @return a link to the generated release notes file hosted on Github.
+     */
+    public String getReleaseNotesUrl(String branch) {
```
szczepiq: I'm wondering if this is unit testable easily. If so, let's a add a test. String concat is a complexity worth unit testing.

epeee: yes, indeed, we should add a test. I will do so.

Issue #390, opened 2017-09-08
Build fails due to upgradeDownstream exception


szczepiq:
Shipkit build has [failed on Travis](https://travis-ci.org/mockito/shipkit/builds/273171764) due to:

```
[./gradlew] :upgradeMockitoShipkit
[./gradlew]   Upgrading dependency:
[./gradlew]     ./gradlew performVersionUpgrade -Pdependency=org.shipkit:shipkit:0.9.60
[./gradlew] [./gradlew]   Building version '0.9.61' (value loaded from 'version.properties' file).
[./gradlew] [./gradlew]   [INCUBATING] downstream-testing plugin is incubating and may change in any version.
[./gradlew] [./gradlew] 
[./gradlew] [./gradlew] FAILURE: Build failed with an exception.
[./gradlew] [./gradlew] 
[./gradlew] [./gradlew] * What went wrong:
[./gradlew] [./gradlew] Task 'performVersionUpgrade' not found in root project 'shipkit'.
[./gradlew] [./gradlew] 
[./gradlew] [./gradlew] * Try:
[./gradlew] [./gradlew] 
[./gradlew] [./gradlew] BUILD FAILED
[./gradlew] [./gradlew] Run gradlew tasks to get a list of available tasks. Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
[./gradlew] [./gradlew] 
[./gradlew] [./gradlew] Total time: 4.044 secs
[./gradlew]   External process [./gradlew] completed.
[./gradlew] :upgradeMockitoShipkit FAILED
[./gradlew] 
[./gradlew] FAILURE: Build failed with an exception.
[./gradlew] 
[./gradlew] * What went wrong:
[./gradlew] Execution failed for task ':upgradeMockitoShipkit'.
[./gradlew] > External process failed with exit code 1
[./gradlew]   Please inspect the command output prefixed with '[./gradlew]' the build log.
[./gradlew] 
[./gradlew] * Try:
[./gradlew] Run with --stacktrace option to get the stack trace. Run with --info or --debug option to get more log output.
[./gradlew] 
[./gradlew] BUILD FAILED
[./gradlew] 
[./gradlew] Total time: 8.967 secs
  External process [./gradlew] completed.
:ciPerformRelease FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':ciPerformRelease'.
```

szczepiq:
This has been fixed! Thank you @epeee fore reviews!

PR #391, opened 2017-09-09
Fixed problem with downstream upgrade, improved logging

szczepiq:
- Fixed problem with 'downstreamUpgrade' task of Shipkit. Adding this plugin allows Shipkit project to automatically send PRs to this project with version bumps of Shipkit.  See: #390
- Improved the default prefix of forked processes …I found it not easy to debug if the default prefix was the same for every invocation. Now, instead of [./gradlew] we will see [performRelease], instead of [git] (git status) we will see [status]. I think it is a better default, it will be easier to debug Travis CI builds. Example of a build log to debug is in #390

### DIFF
--- a/gradle/shipkit.gradle
+++ b/gradle/shipkit.gradle
@@ -12,3 +12,5 @@ afterEvaluate {
 
     tasks.check.dependsOn validatePlugins
 }
+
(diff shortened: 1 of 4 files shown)


szczepiq:
Thank you for review!

PR #393, opened 2017-09-09
Improved the task relationships

szczepiq:
This should help the scenario when we apply shipkit in a subproject.

Inspired by previous good change made by @epeee. Thank you!

I haven't tested this change in anger, yet :)


szczepiq:
Thanks for review!

PR #396, opened 2017-09-09
set previous version to 0.9.60

epeee:
.... in order to get a new release


epeee:
merging (in order to get a new version published).

Issue #403, opened 2017-09-10
upgradeDownstream fails when shipkit moved to subproject


szczepiq:
Relevant line from failed [build log](https://travis-ci.org/mockito/shipkit/builds/273772641):

```
...
[upgradeDownstream]     ./gradlew performVersionUpgrade -Pdependency=:shipkit:0.9.66
...
[upgradeDownstream] [performVersionUpgrade] * What went wrong:
[upgradeDownstream] [performVersionUpgrade] A problem occurred evaluating script.
[upgradeDownstream] [performVersionUpgrade] > Failed to apply plugin [id 'org.shipkit.upgrade-dependency']
[upgradeDownstream] [performVersionUpgrade]    >   Incorrect format of property 'dependency', it should match the pattern '[A-Za-z0-9.\-_]+:[A-Za-z0-9.\-_]+:[0-9.]+', eg. 'org.shipkit:shipkit:1.2.3'.
```

Problems:
- The dependency should be: "0.9.66", not ":shipkit:0.9.66"
- The exception message should tell the input version, currently it tells that the version does not match the pattern :)

szczepiq:
I think this is fixed so I'll try to reenable upgrade downstream

szczepiq:
Here's what I think we need to do to fix it:
 - fix #436. It could be as easy as changing CiUpgradeDownstreamPlugin so that it
    - applies CiReleasePlugin to root project
    - but still applies UpgradeDownstreamPlugin to current project
 - in Shipkit, move the 'upgrade-downstream' plugin so that it is applied in the subproject
 - test in shipkit by calling './gradlew upgradeDownstream'

@epeee, do you want to take a stab at it? I think this would also fix #436.

PR #404, opened 2017-09-10
Commented out downstream until bug #403 is fixed

szczepiq:
@epeee, can you take a look at bug #403? Thanks :)


epeee:
@szczepiq yes, I will have a look at the bug mentioned.

Issue #363, opened 2017-08-26
Shipkit bootstrap pushes to "mockito/shipkit-bootstrap" repository


szczepiq:
## Problem

Performing Shipkit bootstrap pushes to "mockito/shipkit-bootstrap" repository, instead of user's repository:

```
:gitPush
  'git push' uses GitHub write token
  Executing:
    git push https://github.com/mockito/shipkit-bootstrap.git v0.0.2 master
```

This is undesired. During bootstrapping, we want to demonstrate how the release works on the user's repo, for example "szczepiq/shipkit-bootstrap" and not on our repo, e.g. "mockito/shipkit-bootstrap"

## Steps to reproduce

Follow the bootstrap steps: https://github.com/mockito/shipkit-bootstrap/wiki/Shipkit-Test-Drive-in-4-steps

szczepiq:
Fixed.

Issue #285, opened 2017-07-04
Autodiscover name of github repository


szczepiq:
### Problem

Shipkit needs to configure all defaults that can be automatically inferred. This way, the user only has to specify minimum amount of configuration. Currently we explicitly specify ```shipkit.gitHub.repository``` setting in ```shipkit.gradle``` file (see [Mockito's "shipkit.gradle"](https://github.com/mockito/mockito/blob/release/2.x/gradle/shipkit.gradle#L2) file). It would be great to infer the name of the repository automatically.

### Design

- We already have logic that identifies the current git repository, see InitShipkitFileTask
- Implementing this functionality is not trivial because getting repository information requires us to call external git command. We don't want to call git command during Gradle's configuration time. So we need to do it during task execution.
- One of the options to deal with above is to create task 'identifyGitRepository' and use it in similar way we currently use 'identifyGitBranch' task. However, it's going to be a bit of work because we use "repository" all over the place in Shipkit.
- Touch base if you want to work on this and you have more questions!!!

anicos:
@szczepiq can you assign me?

wwilk:
It seems that I can't assign someone not from the team, but it's ok, you can work on that, I think a comment is enough information for others.

szczepiq:
@anicos, are you still interested in working on this? Cheers!

anicos:
Hi. @szczepiq right now I can't take this task. If anyone wants to take it's free:)

szczepiq:
We need to implement this feature because it is blocking #363

I'm on it :)

szczepiq:
I identified where we need to fix it. In GitPlugin we use the repository name from shipkit configuration (org/shipkit/internal/gradle/git/GitPlugin.java:80), instead of inferring it from git origin.

szczepiq:
We cannot really implement this. Let's close.

There are use cases where it is very useful to have have the repo name checked in to the source control. For example, it makes creation of pull requests from forks easier. Also, it makes some useful tasks like preview release notes useful when origin repository is forked.

PR #405, opened 2017-09-11
Autodiscover GitHub repo name

szczepiq:
- Addresses #363 and makes progress on #285
- Started using existing logic we have for identifying GitHub repository. This way, we can configure shipkit.gitHub.repository automatically, which is nice - users don't have to configure it explicitly
- Merged GitOriginPlugin into GitAuthPlugin

Tested by running ```./gradlew testRelease``` in 2 projects: example and shipkit


szczepiq:
Thank you for review!

PR #411, opened 2017-09-13
Uploading downstream tests logs to gist after failure

wwilk:
Tested on Travis docker image, as explained here:
https://stackoverflow.com/questions/21053657/how-to-run-travis-ci-locally#answer-35972902

Here are the log files for success runs of our downstream projects (mockito and shipkit-example):
https://gist.github.com/c410d755d2fc0948d2fa2c26bacb88ee
https://gist.github.com/5c8f3102ad1042ea167662e6bb3f1f73


.travis.yml
```diff
@@ -23,3 +23,10 @@ install:
 
 script:
   - ./gradlew build testDownstream -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
+
+after_failure:
+  - ./after-failure.sh
+
+#TODO Added here only for tests purposes because failure is harder to reproduce. If it's still here - remove it!!
```
szpak: Shouldn't it be removed before merge (looking at the comment :) )?

wwilk:
It seems that:
"message": "API rate limit exceeded for 52.54.31.11. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
I will try the authenticated version tomorrow

.travis.yml
wwilk: Yes, I will remove it as soon as it works correctly on the pull request branch

szczepiq:
Chatted with @wwilk on slack. Gists feel like a complexity to me, so I'm -1. I assume that @wwilk is +1, @szpak, looks like you are the tie breaker and if you +1 it, I will play along :)

szczepiq:
If I go for gists, let's push all complexity to plugin code.

wwilk:
Added as a plugin/task, @szczepiq @szpak please let me know what you guys think

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/UploadGistsPlugin.java
```diff
@@ -0,0 +1,63 @@
...
+ */
+public class UploadGistsPlugin implements Plugin<Project> {
+
+    public static final String FILES_PATTERNS_PROPERTY = "filesPatterns";
+
+    private List<String> filesPatterns;
+
+    @Override
+    public void apply(final Project project) {
+        filesPatterns = extractLogDirPatternsFromProjProperty(project);
```
szczepiq: Why do we need to expose the project property? Can you describe how this plugin is intended to be used? I'm asking because project property makes the public API bigger.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/UploadGistsPlugin.java
```diff
@@ -0,0 +1,63 @@
...
+        final ShipkitConfiguration conf = project.getPlugins().apply(ShipkitConfigurationPlugin.class).getConfiguration();
+
+        TaskMaker.task(project, "uploadGists", UploadGistsTask.class, new Action<UploadGistsTask>() {
+            @Override
+            public void execute(UploadGistsTask uploadGistsTask) {
+                uploadGistsTask.setDescription("Uploads files matching patterns in property '" + FILES_PATTERNS_PROPERTY + "', to Gist");
+                uploadGistsTask.setFilesPatterns(filesPatterns);
```
szczepiq: I suggest to create FileCollection property on the task. This way, we can let Gradle handle the patterns. FileCollection instance can be easily configured with patterns (using API from memory, pardon mistakes):

```groovy
task.myFileCollection = project.files("file1", "file2").filter(...)
task.myFileCollection = project.fileTree("dir").include("*.gist")
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/UploadGistsPlugin.java
```diff
@@ -0,0 +1,63 @@
...
+                uploadGistsTask.setRootDir(project.getRootDir().getAbsolutePath());
+                uploadGistsTask.setGitHubApiUrl(conf.getGitHub().getApiUrl());
+                uploadGistsTask.setGitHubWriteToken(conf.getLenient().getGitHub().getWriteAuthToken());
+            }
+        });
+    }
+
+    private List<String> extractLogDirPatternsFromProjProperty(Project project) {
+        String property = (String) project.findProperty(FILES_PATTERNS_PROPERTY);
+        if (StringUtil.isEmpty(property)) {
```
szczepiq: If we keep the property, let's wrap it up with a unit test at some point. Separate PR ok.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/tasks/UploadGists.java
```diff
@@ -0,0 +1,64 @@
...
+        }
+    }
+
+    private void uploadForPattern(final String logFilePattern, GitHubApi gitHubApi, String rootDir) {
+        for (String logFile : FileUtil.findFilesByPattern(rootDir, logFilePattern)) {
+            uploadSingleFile(logFile, gitHubApi);
+        }
+    }
+
+    private void uploadSingleFile(String filePath, GitHubApi gitHubApi) {
```
szczepiq: For clarity and easy unit testing, I suggest to push all GitHub JSON + posting logic to a separate class. It could expose a single method that takes Strings as parameters (url, token, gist content).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/tasks/UploadGistsTask.java
```diff
@@ -0,0 +1,87 @@
...
+
+import java.util.List;
+
+/**
+ * Uploads files as gists to GitHub API.
+ * Uses {@link UploadGistsTask#getFilesPatterns()} to determine the files that need to be uploaded.
+ * Uploads each of them individually.
+ * Looks for files only in the {@link UploadGistsTask#getRootDir()}.
+ */
+public class UploadGistsTask extends DefaultTask {
```
szczepiq: Very nice documentation, thank you!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/FileUtil.groovy
```diff
@@ -22,4 +22,15 @@ class FileUtil {
...
         notesFile.text = content + existingWithoutHeader
     }
+
+    /**
+     * Finds all files matching {@param pattern} in {@param rootDir} and below it.
+     * @param pattern in Ant format @see <a href="https://ant.apache.org/manual/dirtasks.html#patterns">https://ant.apache.org/manual/dirtasks.html#patterns</a>
+     * eg. ["**.log", "**\/**.txt"]
+     * @param rootDir - absolute path to root for search
+     */
+    static List<String> findFilesByPattern(String rootDir, String pattern){
```
szczepiq: If you go for FileCollection idea, we won't need to keep this.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/release/tasks/UploadGistsTest.groovy
```diff
@@ -0,0 +1,35 @@
+package org.shipkit.internal.gradle.release.tasks
+
+import org.shipkit.internal.util.GitHubApi
+import testutil.PluginSpecification
+
+class UploadGistsTest extends PluginSpecification {
+
+    def "should upload gists correctly"() {
```
szczepiq: This is a really nice test! Good coverage!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/FileUtilTest.groovy
```diff
@@ -26,4 +26,17 @@ class FileUtilTest extends Specification {
         expect:
         f.text == "foo"
     }
+
+    def "finds files by pattern"(){
```
szczepiq: May not be needed!

wwilk:
@szczepiq 
The usage of this plugin I was thinking of would be eg.:
uploadGists -PfilesPatterns="*.txt,**/**.log" 

Which would look for all the files that match all of the patterns and create a separate gist for all of them. The typical use would be in "after_failure" section of Travis but maybe it can be used in other ways too (eg. uploading release information after a successful build?). 
That's why I need a property. Do you have another idea how to do it? I didn't want to have it in project configuration because it would make it harder to use this task in multiple sections of Travis build (like both "after_failure" and "after_success").

szczepiq:
>That's why I need a property. Do you have another idea how to do it? 

I don't think we need a property. For example, we can have:
 - gist plugin
 - test-downstream plugin applies gist plugin
   - also preconfigures gist task to upload specific files

In Travis, we don't configure anything special. We just have ```./gradlew ciPerformRelease```

The users can add support for project property if they want to. They can configure the task to take file collection based on an arbitrary property.

The less public API we expose the better. The users have full configurability on task level if they need to.

Thoughts?

wwilk:
I didn't want to use this uploadGists task with performRelease. Only as a separate process, like eg `./gradlew uploadGists -PfilesPatterns="**/**.log`. I think project property makes task more configurable, since we can use it in a few different gradlew calls in the same .travis.yml file. But maybe this is not needed, so I gave up the project property for now. Configuring it on task may be actually enough.

wwilk:
Thanks for the review! Please let me know if you are OK with the changes I made.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/tasks/UploadGists.java
```diff
@@ -0,0 +1,30 @@
...
+        uploadGists(uploadGistsTask, new GistsApi(gitHubApi));
+    }
+
+    public void uploadGists(UploadGistsTask uploadGistsTask, GistsApi gistsApi) {
+        for (File file : uploadGistsTask.getFilesToUpload()) {
+            try {
+                String url = gistsApi.uploadFile(file.getName(), IOUtil.readFully(file));
+                LOG.lifecycle("Gist for file '{}' created. You can find it here: {}", file.getAbsolutePath(), url);
+            } catch (Exception e) {
+                LOG.error("Creating a Gist for '" + file.getAbsolutePath() + "' failed.", e);
```
szczepiq: It's not a great UX if we log error with stack trace but we don't really fail the task. Think about it: the users sees "BUILD SUCCESSFUL" and a giant stack trace in the build log. "Error" means failure. If it is not a failure, we can log it as standard message: "Unable to create gist, ignoring. Run with --debug to see stack trace".

For the time being, I suggest to just throw an exception. We can think about the best UX for the failure scenarios.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/tasks/UploadGistsTask.java
```diff
@@ -0,0 +1,66 @@
...
+import org.gradle.api.tasks.Input;
+import org.gradle.api.tasks.TaskAction;
+import org.shipkit.gradle.configuration.ShipkitConfiguration;
+
+/**
+ * Uploads files as gists to GitHub API.
+ * Uses {@link UploadGistsTask#getFilesToUpload()} to determine the files that need to be uploaded.
+ * Uploads each of them individually.
+ */
+public class UploadGistsTask extends DefaultTask {
```
szczepiq: This task has great API!!!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/tasks/UploadGists.java
wwilk: Thanks, right, it doesn't make sense to pass the build if the task fails. 

I changed it a little. Since there may be a lot of these files uploaded, it doesn't make sense to not upload all of them if the first one fails. This is an API call so it can fail. But if any of them throws an exception then I'm throwing one at the end of the task. I will merge it for now to see how it works in practice and if we have a better idea for this error handling then I will improve it later.

PR #418, opened 2017-09-16
Made shipkit.gitHub.repository optional

szczepiq:
- The goal is to reduce amount of configuration, especially if that configuration can be inferred.
This makes the tool easier to use. The cost is extra complexity in the code.
- Instead of accessing the repository directly we use the pattern of "identify" task. If some task needs repository name, it needs to depend on "identifyRepository" task / relevant git plugin.
- Made git auth plugin applied to root project only. This way, we avoid triggering multiple git commands, per each submodule

Forgot to say how I tested it: I did ./gradlew testRelease in shipkit and in release-example


szczepiq:
Friendly ping :)

szczepiq:
I am reconsidering this change. If we autodiscover the repository some use cases are not handled well enough.

PR #420, opened 2017-09-17
[#420] Travis cache performance enahncements

szpak:
Travis cache should not be updated (re-uploaded) if there is no changes in project dependencies. With my change all those files are being deleted before cache change detection phase.

It's hard to make any measurements based on Travis builds, however, It should safe up to [60 seconds](https://travis-ci.org/mockito/shipkit/builds/274656727#L2328) per build (minus time needed do rebuild removed files which are updated anyway in every build):
> change detected (content changed, file is created, or file is deleted): - 59.03s

in a situation there are no changes in the project dependencies. Sample results after my change: https://travis-ci.org/mockito/shipkit/builds/276488043#L2108
> nothing changed, not updating cache - 4.70s

It has been working fine (without visible side effects) for a long time in a few my projects built on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ install:
  - true
 
 script:
-  - ./gradlew build testDownstream -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
+  - ./gradlew build testDownstream -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease -s
 
 # Remove often changing files to prevent cache re-upload on no changes in dependencies
 before_cache:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -25,5 +19,19 @@ install:
  - true
 
 script:
-  - ./gradlew build testDownstream -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease
+  - ./gradlew build testDownstream -PcheckJava6Compatibility --daemon -s -i && ./gradlew ciPerformRelease -s
```
wwilk: 👍

PR #422, opened 2017-09-18
Update readme with a better getting started section

StefMa:
Because every developer wants to start as soon as possible with some new libs I've increased the getting started section of the readme.
Then some new uses can directly copy/past some code and can directly start with Shipkit...


szczepiq:
Great thank you very much!

PR #428, opened 2017-09-23
Logging, documentation, init usability

szczepiq:
- Added initial guidelines about writing tests
- Improved logging
   - Previous behavior was telling the user incorrect information. If the version file was not present, the user saw "value loaded from version file". Existing unit tests provide coverage for this change, no need to add new. The change only updates logging
   - I found it useful to understand why identify git origin method has failed at '-i' level.
- Improved initialization experience
   - Changed fallback git repository to something obvious to the user to configure in shipkit file

### DIFF
--- a/docs/writing-tests.md
+++ b/docs/writing-tests.md
@@ -1,0 +1,10 @@
+# Testing guidelines
+
+## Write tests for...
+
+Write tests for all complexity. Period.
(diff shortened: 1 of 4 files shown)


szczepiq:
Need to publish it from a different branch

Issue #427, opened 2017-09-22
Make publication repository configurable in shipkit file


StefMa:
### Problem

When using 'org.shipkit.gradle-plugin' the user needs to remember to configure publication repository on the task: https://github.com/mockito/shipkit/blob/master/gradle/shipkit.gradle#L10 We want the shipkit configuration to be central point of the important configuration. Unimportant configuration settings are fine at the task level. Publication repository setting is important because without it we cannot generate release notes for 'org.shipkit.gradle-plugin' users.

### Suggested implementation

See the initial design at #430

szczepiq:
@StefMa, I updated the description and hacked my initial thoughts about the implementation. Do you want to take a stab at it? We love contributions!!! :)

StefMa:
Since #446 is merged. This can be closed.

PR #430, opened 2017-09-23
Initial design for configurable publication repository setting

szczepiq:
Does not compile but should lead the way :) For the details of the issue, see #427


StefMa:
Since #446 is merged. This can be closed.

Issue #432, opened 2017-09-25
Documentation for Shipkit 1.0 release


mockitoguy:
Let's sort out the documentation before 1.0 release.

 - [x] - move all documentation from wiki to checked-in markdown files to help with contributions
 - [x] - update history, conference links, meetup links
 - [x] - update README.md
 - [x] - update how-shipkit-works.md
 - [x] - make bootstrap use getting started guide, retire bootstrap wiki

After the release we will continue working on the documentation: #534

szczepiq:
I am going to chip away the TODOs until we are ready

wwilk:
This is related to #362

mockitoguy:
This ticket is done!

PR #433, opened 2017-09-25
Updated the documentation

szczepiq:
Chipping away TODOs in #432


szczepiq:
I'll merge it if that's ok. Happy to update the doc afterwards if there is feedback. I'm linking to the project page from my recent article and it would be useful to have it updated :)

PR #440, opened 2017-09-27
Make "upgrade-downstream" plugin work well when applied in subproject

### LINKED ISSUE #436: Make "upgrade-downstream" plugin work well when applied in subproject
The plugins should be flexible so that users can apply them in subprojects if needed. This will allow us to apply "upgrade-downstream" plugin in the same subproject as "test-downstream".

epeee:
fixes #436

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CiUpgradeDownstreamPlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CiUpgradeDownstreamPlugin.java
@@ -22,10 +22,10 @@ public class CiUpgradeDownstreamPlugin implements Plugin<Project> {
 
     @Override
     public void apply(Project project) {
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CiUpgradeDownstreamPlugin.java
```diff
@@ -22,10 +22,10 @@
 
     @Override
     public void apply(Project project) {
-        project.getPlugins().apply(CiReleasePlugin.class);
+        project.getRootProject().getPlugins().apply(CiReleasePlugin.class);
```
szczepiq: Great! Can you update the Javadoc for the CiUpgradeDownstreamPlugin? @mockito/shipkit-developers, let's make sure we update the Javadocs when we're making changes. We need good javadocs for 1.0 ;)

PR #446, opened 2017-09-29
configurable publication repository

StefMa:
Rest of the implementation of #430 :)

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -257,7 +257,12 @@ public class ShipkitConfiguration {
         }
 
         /**
-         * Set the Publication Repository where your project can be found
+         * Set the Publication Repository where we look for your published binary.
+         * E.g.
+         * <pre>
+         *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
+         * </pre>
+         * This will be used for adding and linking the repository in the release notes.
          *
          * @see #getPublicationRepository()
          */
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -140,7 +140,7 @@ public class ShipkitConfiguration {
 
         /**
          * GitHub repository name, for example: "mockito/shipkit"
-         *
+         * <p>
          * //TODO SF Javadoc about getRepository() throughout the project need to include information how we obtain this repo automatically
          */
         public String getRepository() {
@@ -254,6 +254,24 @@ public class ShipkitConfiguration {
          */
         public void setIgnoreCommitsContaining(Collection<String> commitMessageParts) {
             store.put("releaseNotes.ignoreCommitsContaining", commitMessageParts);
+        }
+
+        /**
+         * Set the Publication Repository where your project can be found
+         *
+         * @see #getPublicationRepository()
+         */
+        public void setPublicationRepository(String publicationRepository) {
+            store.put("releaseNotes.publicationRepository", publicationRepository);
+        }
+
+        /**
+         * Get the Publication Repository
+         *
+         * @see #setPublicationRepository(String)
+         */
+        public String getPublicationRepository() {
+            return store.getString("releaseNotes.publicationRepository");
         }
     }
 
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
@@ -64,7 +64,7 @@ public class BintrayReleasePlugin implements Plugin<Project> {
                         deferredConfiguration(subproject, new Runnable() {
                             public void run() {
                                 UpdateReleaseNotesTask updateNotes = (UpdateReleaseNotesTask) project.getTasks().getByName(ReleaseNotesPlugin.UPDATE_NOTES_TASK);
-                                boolean userSpecifiedRepo = conf.getLenient().getReleaseNotes().getPublicationRepository();
+                                String userSpecifiedRepo = conf.getLenient().getReleaseNotes().getPublicationRepository();
                                 if (userSpecifiedRepo != null) {
                                     updateNotes.setPublicationRepository(userSpecifiedRepo);
                                 } else {
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
@@ -21,16 +21,16 @@ import static org.shipkit.internal.gradle.java.JavaPublishPlugin.MAVEN_LOCAL_TAS
  * Configures Java project for automated releases with Bintray.
  * <p>
  * Applies following plugins:
- *
+ * <p>
  * <ul>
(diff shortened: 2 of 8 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
```diff
@@ -61,7 +64,12 @@ public void execute(JavaBintrayPlugin plugin) {
                         deferredConfiguration(subproject, new Runnable() {
                             public void run() {
                                 UpdateReleaseNotesTask updateNotes = (UpdateReleaseNotesTask) project.getTasks().getByName(ReleaseNotesPlugin.UPDATE_NOTES_TASK);
-                                updateNotes.setPublicationRepository(BintrayUtil.getRepoLink(bintray));
+                                String userSpecifiedRepo = conf.getLenient().getReleaseNotes().getPublicationRepository();
+                                if (userSpecifiedRepo != null) {
```
wwilk: Why lenient config here, while regular one in GradlePortalReleasePlugin?

StefMa: Sry. I don't really get it 🙃 
That was original designed by @szczepiq.

Should I change that code to
```
(UpdateReleaseNotesTask) project.getTasks().getByName(ReleaseNotesPlugin.UPDATE_NOTES_TASK);
```
as well? 🤔

wwilk: It would be also good to create unit tests for these two cases.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/GradlePortalReleasePlugin.java
```diff
@@ -59,6 +64,9 @@ public void execute(Plugin plugin) {
                         Task archivesTask = subproject.getTasks().getByName("buildArchives");
                         publishPlugins.dependsOn(archivesTask);
                         gitPush.mustRunAfter(archivesTask);
+
+                        UpdateReleaseNotesTask updateNotes = (UpdateReleaseNotesTask) project.getTasks().getByName(ReleaseNotesPlugin.UPDATE_NOTES_TASK);
+                        updateNotes.setPublicationRepository(conf.getReleaseNotes().getPublicationRepository());
```
wwilk: Also missing unit test here

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
StefMa: Hmm.. wait 🤔 
We have to get the `publicationRepository` from the config here.
Because we want to set it to the task :)
We can't use the task to get the `publicationRepository`. It is always `null` ...

wwilk: I'm only asking about lenient config here: 
`conf.getLenient().getReleaseNotes().getPublicationRepository()`
Why here it's lenient, and in GradlePortalReleasePlugin it's not lenient. 
In general we use lenient one when the property can be null. But you are right, it was like that in the first commit for this issue.
@szczepiq Why is it different for these two classes?

szczepiq: Sure!

lenient() is documented in the Javadoc, let me know if it is unclear (or suggest an improvement / rename).

In this case, we should be using lenient() because repo is optional - we can get the value automatically from bintray extension. In the other case (gradle portal plugin), it is not optional, we cannot get the value from bintray extension because bintray plugin is not used.

Does it sound right? Good question!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
```diff
@@ -61,7 +64,12 @@ public void execute(JavaBintrayPlugin plugin) {
...
+                                    updateNotes.setPublicationRepository(userSpecifiedRepo);
+                                } else {
```
szczepiq: Unit test please :) Let us know if you need help.

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -255,6 +255,24 @@ public void setLabelMapping(Map<String, String> labelMapping) {
         public void setIgnoreCommitsContaining(Collection<String> commitMessageParts) {
             store.put("releaseNotes.ignoreCommitsContaining", commitMessageParts);
         }
+
+        /**
+         * Set the Publication Repository where your project can be found
```
szczepiq: The publication repository is the place where to look for binaries. It could be MavenCentral, Bintray or some other repository. Can you update the javadoc accordingly? Can you add an example in the Javadoc? See an example if Shipkit project itself.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/GradlePortalReleasePlugin.java
szczepiq: +1

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePlugin.java
StefMa: A little bit help would be great 👍 
I've never used spock before. I've also never tested a gradle plugin 🙃 
I've tried it with something like that:
```
    def "applies"() {
        expect:
        project.plugins.apply(BintrayReleasePlugin)
        project.plugins.apply(JavaBintrayPlugin)
        //def repository = project.tasks.getByName("updateReleaseNotes").getPublicationRepository()
        //assert repository == null
        project.evaluate()
        def repositoryNonNull = project.tasks.getByName("updateReleaseNotes").getPublicationRepository()
        assert repositoryNonNull != null
    }
```
But unfortunately that isn't working..

StefMa:
Wow. Thanks for the help @szczepiq 💪 😗 
I've cherry-picked your example test and adapt it to the new needs.
Hope that don't lead into merge conflicts if #456 got merged 🙂

Hope everything is fine now 😇

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePluginTest.groovy
```diff
@@ -1,17 +1,45 @@
...
+        conf.releaseNotes.publicationRepository = "publicRepo"
 
-    def "applies"() {
-        expect:
         project.plugins.apply(BintrayReleasePlugin)
         project.plugins.apply(JavaBintrayPlugin)
+
+        project.bintray.pkg.userOrg = "some-org"
+        project.bintray.pkg.repo = "some-repo"
+        project.bintray.pkg.name = "some-pkg"
```
wwilk: Configuration of bintray extension seems irrelevant for this test. UpdateReleaseNotesTask will use conf.releaseNotes.publicationRepository to set the value of publicationRepository.

StefMa:
Because of merge conflicts I've rebased the branch.
I've also removed the configuration of the bintray plugin in the test (see https://github.com/mockito/shipkit/pull/446#discussion_r142036445)

wwilk:
Thanks! Merging

szczepiq:
@StefMa, Can you fix Shipkit so that it can consume new version? After this PR was merged, Shipkit no longer can consume Shipkit :) See the build failure: https://travis-ci.org/mockito/shipkit/builds/282839303?utm_source=github_status&utm_medium=notification

PR #448, opened 2017-10-01
Improved documentation

szczepiq:
- All wiki content is now moved!
- Tailored getting started guide for Shipkit Bootstrap


docs/getting-started.md
```diff
@@ -6,23 +6,37 @@ See also [documentation index](README.md#documentation)
...
 ### Adding dependency
 
-First step on the way to use Shipkit is adding a dependency on “org.shipkit:shipkit:${version}” which currently can be downloaded from “https://plugins.gradle.org/m2/”, so you need to add this Maven repository to your gradle configuration if you don’t already have it. Additionally you need to apply “org.shipkit.java” plugin, only in your root project.
+Add below to your root "build.gradle" file.
+Get the latest version of "org.shipkit.java" plugin from [Gradle Plugin Portal](https://plugins.gradle.org/plugin/org.shipkit.java)
 
-Afterwards your build file should look like this:
+```Gradle
+plugins {
+    id "org.shipkit.java" version "TODO"
```
epeee: should we make use of the same version as we are using in the traditional config?

szczepiq: That's a good idea.

It would be cool if Shipkit during the release updated the documentation automatically with the right version. However, it is a small improvement.

PR #454, opened 2017-10-01
fix broken link in documentation

epeee:


docs/getting-started.md
```diff
@@ -1,7 +1,7 @@
 ## Getting started
 
 Thank you for using Shipkit!
-See also [documentation index](README.md#documentation)
+See also [documentation index](../README.md#documentation)
```
szczepiq: Woa! I didn't know that we can do relative paths like that. Great!

PR #458, opened 2017-10-01
enable codenarc formatting rules and fix violations

### LINKED ISSUE #22: Enable consistent code style validation

epeee:
fixes #22


config/codenarc/codenarc.xml
```diff
@@ -15,4 +15,10 @@
 
     <ruleset-ref path='rulesets/unused.xml'/>
 
-</ruleset>
\ No newline at end of file
+    <ruleset-ref path='rulesets/formatting.xml'>
+        <exclude name="ClassJavadoc"/>
+        <exclude name="LineLength"/>
+        <exclude name="SpaceAroundMapEntryColon"/>
```
epeee: I disabled the 3 rules above because I was not sure if we should go for them.
@mockito/shipkit-developers what do you think?

wwilk: I don't think each class needs a Javadoc, for example internal implementation of tasks (like ReleaseNeeded). It seems that well-documented task is enough. Or for utility classes more important is to have method comments than class ones.

I'm OK with LineLength, default value 120 seems to be OK. Or maybe a little more

SpaceAroundMapEntryColon - I'm neutral.

szczepiq: - ClassJavadoc - ok to exclude. Let's use own judgement. In general, we should have Javadoc on all public types.
- LineLength - ok to exclude, I think that naturally when we care about the code we write / review we will keep line length reasonable
- SpaceAroundMapEntryColon - why we need to exclude it?

epeee: @wwilk, @szczepiq : thx for your feedack!
In a first version I excluded `SpaceAroundMapEntryColon` because it caused a lot of violations and according to goovy doc the map syntax we are using `[key: value]` seems fine. Therefore I decided to change the config of codenarc and make codenarc validating the syntax described.

Issue #419, opened 2017-09-17
org.shipkit.gradle-plugin releases even if there are no changes in binaries


szpak:
### Problem

Preventing the release happening when there are not changes in the binary files is not currently implemented for 'org.shipkit.gradle-plugin' users. It is only implemented for 'org.shipkit.java' users.

### Implementation

The plugin that compares publications is called "ComparePublicationsPlugin" and it is applied when 'org.shipkit.java' is in use. Let's figure out how can we make 'org.shipkit.gradle-plugin' leverage compare publications plugin.

### Original report from @szpak

I've just [tweaked](https://github.com/mockito/shipkit/commit/1c5a8727dfbc01f644db4c937c0dbb7b256ae20e) release notes and the new version has been released (as the publications was [not equal](https://travis-ci.org/mockito/shipkit/builds/276207813#L1551)). Is it the expected behavior?

wwilk:
It's not the expected behaviour. We compare poms and sources jars there and docs/ should not be a part of this. I will investigate it today, thanks for opening this issue!

BTW we have a strategy to create pull requests even for small changes like this one, and merge it only when someone else approves it or if it stays too long as open and is blocking us.

szczepiq:
I saw this in the logs:

```
[assertReleaseNeeded]   Publication comparison was skipped (no comparison result files found).
```

I will look into this

szczepiq:
We currently don't support skipping publications if binaries are the same for Gradle plugins. Our "org.shipkit.java" plugin supports it, but the "org.shipkit.gradle" does not. It should not be too difficult to add this. Who wants to give it a stab? :) The plugin that compares publications is called "ComparePublicationsPlugin"

szczepiq:
Updated the description. @szpak, do you want to take a stab at it?

wwilk:
OK I'm starting to work on this, since lack of it seems to create an issue #443

wwilk:
@szpak Please let me know if you already started working on it

szczepiq:
@wwilk, it seems the issue is not yet fixed, even though we added compare publications plugin. I saw the PR opened for a documentation-only change (markdown files). Pull request: https://github.com/mockito/shipkit/pull/492

wwilk:
Yeah it seems that ComparePublications finds differences in pom when there is none. I'll investigate

szpak:
@wwilk No, I haven't started working on it yet.

wwilk:
Fixed in version 0.9.139

PR #461, opened 2017-10-01
ComparePublications working for Gradle plugins #419

wwilk:
I tested it locally. It seems that there is always a difference in poms but since this is not a critical situation, we can test if the same problem occurs on Travis.

### DIFF
--- a/subprojects/shipkit/shipkit.gradle
+++ b/subprojects/shipkit/shipkit.gradle
@@ -8,6 +8,9 @@ apply plugin: 'com.gradle.plugin-publish'
 
 apply from: "$rootDir/gradle/upgrade-downstream.gradle"
 
+//TODO: WW uncomment when GradlePluginArtifactUrlResolver is released
+//apply plugin: 'org.shipkit.compare-publications'
+
 checkstyle {
     configFile = file("${rootDir}/config/checkstyle/checkstyle.xml")
 }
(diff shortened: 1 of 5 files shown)


subprojects/shipkit/shipkit.gradle
```diff
@@ -8,6 +8,9 @@ apply plugin: 'com.gradle.plugin-publish'
 
 apply from: "$rootDir/gradle/upgrade-downstream.gradle"
 
+//TODO: WW uncomment when GradlePluginArtifactUrlResolver is released
+//apply plugin: 'org.shipkit.compare-publications'
```
szczepiq: Why don't we fold it under 'org.shipkit.gradle-plugin'? It's already very opinionated by design.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/artifact/DefaultArtifactUrlResolverFactory.java
```diff
@@ -2,12 +2,15 @@
...
 import org.gradle.api.Project;
 import org.shipkit.internal.gradle.bintray.ShipkitBintrayPlugin;
+import org.shipkit.internal.gradle.plugin.GradlePortalPublishPlugin;
 
 public class DefaultArtifactUrlResolverFactory {
 
     public DefaultArtifactUrlResolver getDefaultResolver(Project project, String artifactBaseName, String previousVersion) {
         if (project.getPlugins().hasPlugin(ShipkitBintrayPlugin.class)) {
             return new BintrayDefaultArtifactUrlResolver(project, artifactBaseName, previousVersion);
+        } else if (project.getPlugins().hasPlugin(GradlePortalPublishPlugin.class)) {
```
szczepiq: Really nice idea, thank you!

PR #462, opened 2017-10-01
Uploading Gists for downstreamTests' log files enabled

wwilk:
I tested it locally. Generated gist:
https://gist.github.com/a3d9eeda7aae0d38776f98f8a8e77b4e


.travis.yml
```diff
@@ -35,3 +35,8 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
+after_success:
+  - ./gradlew uploadGists
```
szczepiq: Why not making it part of the test-downstream plugin? It would be easier to use by our customers!

gradle/shipkit.gradle
```diff
@@ -12,3 +12,6 @@ afterEvaluate {
 }
 
 apply plugin: 'org.shipkit.upgrade-dependency'
+
+apply plugin: 'org.shipkit.upload-gists'
```
szczepiq: I vote to fold this plugin under 'test-downstream' :)

.travis.yml
wwilk: You mean not only applying it inside TestDownstreamPlugin but also making testDownstream be finalizedBy uploadGists? Shouldn't uploading logs be done in after_failure part of Travis build? It may take some time since these are network calls with potentially big attachments, so you would get later an information if build is successful.

szczepiq: Let's avoid premature optimizations and focus on simplicity at first. If this indeed takes too long we can brainstorm how to make it better. There multiple options: do it in parallel, use after_failure/success.

Good question!

wwilk:
Applied in TestDownstreamPlugin. Tested by dogfooding and running "./gradlew testDownstream" in Shipkit. Gists properly generated.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -22,12 +27,30 @@
 
     public void apply(final Project project) {
         IncubatingWarning.warn("downstream-testing plugin");
-        TaskMaker.task(project, TEST_DOWNSTREAM_TASK, TestDownstreamTask.class, new Action<TestDownstreamTask>() {
+
+        project.getPlugins().apply(UploadGistsPlugin.class);
```
szczepiq: Can you update the javadoc of this class? We're changing the behavior and applying another plugin.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -22,12 +27,30 @@
...
             public void execute(TestDownstreamTask task) {
                 task.setDescription("Runs all downstream tests.");
             }
         });
+
+        final UploadGistsTask uploadGistsTask = (UploadGistsTask) project.getTasks().findByName(UploadGistsPlugin.UPLOAD_GISTS_TASK);
+
+        uploadGistsTask.setFilesToUpload(
+            project
+                .fileTree(testDownstreamTask.getLogDirectory())
```
szczepiq: Check the API of fileTree. It has some nice syntax for matching and including specific files. It will make this code more elegant. I don't remember the API but you probably can do something like:

```
fileTree(testDownstreamTask.getLogDirectory()).include("**/*.log")
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -22,12 +27,30 @@
...
+                .filter(new Spec<File>() {
+                    @Override
+                    public boolean isSatisfiedBy(File file) {
+                        return file.getName().endsWith(".log");
+                    }
+                })
+        );
+
+        testDownstreamTask.finalizedBy(uploadGistsTask);
```
szczepiq: Will this work?

testDownstreamTask depends on all TestReleaseTasks. If one of the TestReleaseTasks fails, testDownstreamTask will not be executed and hence it will not be finalized. Also if I run one testReleaseTask explicitly I will not get the gists.

I think that gist task needs to finalize every individual TestReleaseTask.

szczepiq:
Reviewed :) Found some potential issues to look at!

The direction is good - we want Shipkit to have nice default, minimal configuration and decent out-of-the box behavior.

Thank you for working on this!!!

wwilk:
Thanks for the feedback! After the changes I tested it on failing downstream test release task (changed 'releaseNeeded' to 'releaseNeededss' in task command line), it correctly uploaded a gist for the task.

About running it without GH_WRITE_TOKEN - I would like to create a separate CR for that. The first solution that comes to mind is to make gitHubWriteToken optional, and not to run the task if it's null. But probably it would be better to disable the task from TestDownstreamPlugin without any changes in UploadGists itself. WDYT @mockitoguy ?

wwilk:
BTW my commit doesn't introduce any failure, failed check is caused by failing test on master branch. You can see above that it passed the check with @Ignore flag for commit 4113337.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -22,12 +35,36 @@
...
+        final UploadGistsTask uploadGistsTask = (UploadGistsTask) project.getTasks().findByName(UploadGistsPlugin.UPLOAD_GISTS_TASK);
+
+        uploadGistsTask.setFilesToUpload((ConfigurableFileTree)
+            project
+                .fileTree(testDownstreamTask.getLogDirectory())
+                .include("*.log")
+        );
+
+        // it has to be deferred because TestDownstreamReleaseTask are added by user after applying this plugin
+        DeferredConfiguration.deferredConfiguration(project, new Runnable() {
```
mockitoguy: A couple of options:
1. Add this to TravisPlugin. E.g. travis plugin would configure 'finalizedBy' only when the build is invoked on Travis CI. If the build is invoked locally, we don't send gists
2. TravisPlugin checks if we are on CI machine and if so, we are making uploadGistsTask.enabled = false
3. Similar to previous point but flipping the default. By default uploadGists is disabled. Only TravisPlugin enables it when we're running on Travis.

I think I like option 3) most. There are more options of course :) please suggest!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -22,12 +35,36 @@
...
+            @Override
+            public void run() {
```
mockitoguy: What do you think about simplifying it? 

- we create uploadGists task before testDownstream task
- we pass an instance of uploadGists to testDownstream task
- each time the user calls addRepository method, we automatically add 'finalizedBy(uploadGists)'

This way we don't need deferred configuration block and we also don't need the marker interface.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPluginTest.groovy
```diff
@@ -15,4 +13,19 @@ class TestDownstreamPluginTest extends Specification {
         project.tasks.testDownstream
     }
 
+    def "should find log files for uploadGists in log directory"() {
```
mockitoguy: Nice test!

mockitoguy:
>About running it without GH_WRITE_TOKEN - I would like to create a separate CR for that.

Sure! Let's just not wait too long with this. Let's make the project build correctly for new devs who are not core contributors!

>The first solution that comes to mind is to make gitHubWriteToken optional, and not to run the task if it's null. But probably it would be better to disable the task from TestDownstreamPlugin without any changes in UploadGists itself.

Disabling the task works. See some of my suggestions in the code review.

Thank you for working on this!

wwilk:
Suggestions applied, except for the one with disabling UploadGists in TravisPlugin. I think it would be better to do it in TestDownsteamPlugin. This way we don't create a hidden dependency on TestDownstream in TravisPlugin. 

I tested it by running ./gradlew testDownstream with CI=true and CI missing.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/downstream/test/TestDownstreamPlugin.java
```diff
@@ -3,31 +3,85 @@
...
+        uploadGistsTask.setFilesToUpload((ConfigurableFileTree)
+            project
+                .fileTree(logsDirectory)
+                .include("*.log")
+        );
+
+        boolean shouldUploadLogs = ciContext.isCiBuild() && !StringUtil.isEmpty(conf.getLenient().getGitHub().getWriteAuthToken());
+
+        if (shouldUploadLogs) {
+            LOG.debug("  Skipping UploadGists task execution." +
```
mockitoguy: I suggest to bump it to 'info' and made the message a little bit user friendly, for example: "Skipping UploadGists task execution" -> "  test-downstream plugin will not send logs to Gist (setting uploadGists.enabled=false)."

wwilk:

Issue #368, opened 2017-08-26
Hardcoded pull request title and description


mockitoguy:
## Problem

Our really cool 'org.shipkit.upgrade-downstream' adds and configures CreatePullRequestTask. This task automatically creates pull requests to other projects with version bump. The task implementation [hardcodes pull request](https://github.com/mockito/shipkit/blob/a94f0a12f6070b1ba0be5127dfdab110f66ee154/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java#L39) 'title' and 'description' values. This makes the task less robust, less configurable. It also ties implementation details with configuration.

## Suggested solution

Let's expose 'title' and 'description' properties on the [CreatePullRequestTask](https://github.com/mockito/shipkit/blob/master/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java). Ideally, the values of those properties should be configured by the plugin.

This ticket is marked as 'please-contribute'. Please let us know if you have questions on how to get started!

ArturSkowronski:
@szczepiq I would like to take it :)

szczepiq:
@ArturSkowronski, go for it! We don't have a way to assign the ticket in the tracker for non-members. However, consider the ticket yours ;)

ArturSkowronski:
I have one question @szczepiq - do we want to allow users to use some kind of placeholders for dependency version and dependency name in their PR descriptions/titles that would be filled automatically - similar to current solution? Especially the version number would be something that I would like to have filled in title of PR.

Also, I would like to confirm - if somebody don't pass title/description in task configuration, we should fallback to current solution, am I right? As sensible default.

wwilk:
@ArturSkowronski No needs for placeholders or default values. The idea is to make this task more generic, so that it can be reused in other scenarios where creating a pull request would be useful ( example from the top of my head - performing automated migration in downstream projects).

[...]

PR #474, opened 2017-10-03
[SHIPKIT-368] Move pull request title and description to plugin configuration

ArturSkowronski:
Relates to issue #368 

In this implementation, this is not possible to set version and dependency name in description

I throw the exception when pull request description or title is not specified, but I'm a bit worried there about backward compatibility. In this way all previous configuration with fail. Still, I tried to add a direct suggestion how to resolve that as possible in error message.

Travis checks should pass after https://github.com/mockito/shipkit-example/pull/109 will be merged


subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -360,6 +360,34 @@ public void setCommitMessagePostfix(String commitMessagePostfix) {
             //TODO protect this setter and other relevant from invalid input (null value)
             store.put("git.commitMessagePostfix", commitMessagePostfix);
         }
+
+        /**
+         * Text will be used as description for pull request to downstream repositories.
+         */
+        public String getPullRequestDescription() {
+            return store.getString("git.pullRequestDescription");
```
wwilk: We don't need properties in ShipkitConfiguration at the moment. We don't want it to grow more than necessary. You can start with only task properties, so that you could modify task like that:
project.tasks.createPullRequest.pullRequestDescription = "Upgraded version"

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -2,11 +2,14 @@
 
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
+import org.gradle.internal.impldep.org.apache.commons.lang.StringUtils;
```
wwilk: There is StringUtil class internally in Shipkit so no need to make a dependency on Gradle internals.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -2,11 +2,14 @@
...
 import org.shipkit.internal.util.GitHubApi;
 import org.shipkit.internal.util.IncubatingWarning;
 
 import java.io.IOException;
 
+import static org.gradle.internal.impldep.org.apache.commons.lang.StringUtils.isBlank;
```
wwilk: The same as with the other import - you can use Shipkit's own StringUtil.isEmpty

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -37,16 +43,22 @@ public void createPullRequest(CreatePullRequestTask task, GitHubApi gitHubApi) t
...
-         " 'org.shipkit.upgrade-downstream' Gradle plugin (http://shipkit.org)." +
-        " Please merge it so that you are using fresh version of '%s' dependency.",
-            task.getVersionUpgrade().getDependencyName());
+        return task.getPullRequestDescription();
     }
 
     private String getTitle(CreatePullRequestTask task) {
-        UpgradeDependencyExtension versionUpgrade = task.getVersionUpgrade();
-        return String.format("Version of %s upgraded to %s", versionUpgrade.getDependencyName(), versionUpgrade.getNewVersion());
+        return task.getPullRequestTitle();
```
wwilk: It seems that these private methods getTitle and getMessage are no longer needed

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -37,16 +43,22 @@ public void createPullRequest(CreatePullRequestTask task, GitHubApi gitHubApi) t
         gitHubApi.post("/repos/" + task.getUpstreamRepositoryName() + "/pulls", body);
     }
 
+    private void checkPullRequestMetadata(CreatePullRequestTask task) {
+        if (isBlank(getTitle(task))) {
```
wwilk: Since CreatePullRequestTask#pullRequestDescription and CreatePullRequestTask#pullRequestTitle have @Input annotation on them, there is no need to throw an exception, Gradle will validate it for us when someone runs this task without setting these properties. You can take a look at: https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/Optional.html

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequestTask.java
```diff
@@ -123,4 +125,34 @@ public void setDryRun(boolean dryRun) {
     public boolean isDryRun() {
         return dryRun;
     }
+
+    /**
+     * Description of pull request for downstream repositories.
+     */
```
wwilk: Thanks for updating the docs! "Description of pull request" should be enough, this task can be used in different contexts than downstream repos.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequestTask.java
```diff
@@ -123,4 +125,34 @@ public void setDryRun(boolean dryRun) {
...
+    /**
+     * See {@link #getPullRequestDescription()}
+     */
+    public void setPullRequestDescription(String pullRequestDescription) {
+        this.pullRequestDescription = pullRequestDescription;
+    }
+
+    /**
+     * Title of pull request for downstream repositories.
+     */
```
wwilk: The same as above - "for downstream repositories" part is not needed

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -198,6 +198,8 @@ public void execute(final CreatePullRequestTask task) {
                 task.setAuthToken(conf.getLenient().getGitHub().getWriteAuthToken());
                 task.setVersionBranch(getVersionBranchName(upgradeDependencyExtension));
                 task.setVersionUpgrade(upgradeDependencyExtension);
+                task.setPullRequestTitle(conf.getGit().getPullRequestTitle());
+                task.setPullRequestDescription(conf.getGit().getPullRequestDescription());
```
wwilk: Since this configuration of **CreatePullRequest** task is used specifically for upgrading dependencies, we can use default values that we had previously hardcoded in **CreatePullRequest** class. So for **pullRequestTitle** it would be "Version of %s upgraded to %s". 

This way user can use the default value or override it like that:

apply plugin: "org.shipkit.upgrade-dependency"
project.tasks.createPullRequest.pullRequestTitle = "some title"

wwilk: This would be more complicated. The scope of this issue is only to allow task to be configurable. No need for complex smart defaults, just the default specified in the plugin at the moment would suffice. Since 'pullRequestTitle' and 'pullRequestDescription' will be now configurable user will still be able to set it to anything they want.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -23,43 +24,42 @@
...
- *
- *      upgradeDependency{
- *          baseBranch = 'release/2.x'
- *          buildFile = file('build.gradle')
- *      }
- *
+ * <p>
+ * apply plugin: 'org.shipkit.upgrade-dependency'
+ * <p>
+ * upgradeDependency{
```
wwilk: Probably your IDE removed whitespaces here. It would be useful to keep it, it's more readable this way.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -226,6 +228,26 @@ public void execute(Task task) {
         });
     }
 
+    private String getPullRequestDescription(CreatePullRequestTask task) {
+        if (StringUtil.isEmpty(task.getPullRequestDescription())) {
```
wwilk: This "if" is not needed here. Execute action will be executed right after creating a task so there is no way that user overrode one of the parameters in the meantime. User can still override it later.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -226,6 +228,26 @@ public void execute(Task task) {
...
+                    " 'org.shipkit.upgrade-downstream' Gradle plugin (http://shipkit.org)." +
+                    " Please merge it so that you are using fresh version of '%s' dependency.",
+                task.getVersionUpgrade().getDependencyName());
+        } else {
+            return task.getPullRequestDescription();
+        }
+    }
+
+    private String getPullRequestTitle(CreatePullRequestTask task) {
+        if (StringUtil.isEmpty(task.getPullRequestDescription())) {
```
wwilk: The same as for the "if" in getPullRequestDescription.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPluginTest.groovy
```diff
@@ -176,5 +176,30 @@ class UpgradeDependencyPluginTest extends PluginSpecification {
         task.authToken == "writeToken"
         task.versionUpgrade == versionUpgrade
         task.versionBranch == "upgrade-shipkit-to-1.2.30"
+        task.pullRequestDescription == "This pull request was automatically created by " +
+                "Shipkit's 'org.shipkit.upgrade-downstream' Gradle plugin (http://shipkit.org). " +
+                "Please merge it so that you are using fresh version of 'shipkit' dependency."
+        task.pullRequestTitle == "Version of shipkit upgraded to ${versionUpgrade.newVersion}"
+    }
+
+    def "should configure createPullRequest basing on overriden values"() {
```
wwilk: When we remove the if's in getPullRequestTitle and getPullRequestDescription we don't need this test anymore.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
ArturSkowronski: @wwilk Where should defaults be initialized then? In getter of this class in similar way I did in Upgrade Dependency Plugin but in CreatePullRequestTask? Cause I need to have access to  ${versionUpgrade.newVersion} while setting their default value

I will be thankful for hint there.

wwilk: Getter is fine here, this line:
`if (StringUtil.isEmpty(task.getPullRequestDescription()`
is redundant because it will always be empty at this point.

PR #479, opened 2017-10-04
ComparePublicationsPlugin applied in Shipkit

wwilk:


subprojects/shipkit/shipkit.gradle
```diff
@@ -8,8 +8,7 @@ apply plugin: 'com.gradle.plugin-publish'
 
 apply from: "$rootDir/gradle/upgrade-downstream.gradle"
 
-//TODO: WW uncomment when GradlePluginArtifactUrlResolver is released
-//apply plugin: 'org.shipkit.compare-publications'
+apply plugin: 'org.shipkit.compare-publications'
```
epeee: I'd vote for making it the default and not applying it in the build.gradle file.
I just had a quick look at the code and I think `GradlePortalReleasePlugin` might be the place to apply the plugin.

shipkit-org: +1 let's include it by default!

PR #482, opened 2017-10-06
fix broken links in how-we-work.md

epeee:


docs/how-we-work.md
```diff
@@ -1,8 +1,8 @@
 ## How we work
 
 We are a community project and we work as a team.
- - See [CONTRIBUTING.md](CONTRIBUTING.md) for how to contribute
- - See [documentation index](README.md#documentation)
+ - See [CONTRIBUTING.md](../CONTRIBUTING.md) for how to contribute
```
wwilk: I think using absolute URLs (/CONTRIBUTING.md instead of ../) is more flexible. If we move this file the link would still work.

epeee: thx! yep, does make sense for links to files which are in the root dir.
updated the pr

wwilk:
Thanks @epeee!

Issue #450, opened 2017-10-01
Emit nice summary at the end of the release


szczepiq:
### Problem

When release build ends, the user does not know what happened. We also use "performRelease" for demos / bootstrapping. Newbie users will use that task.

### Solution

It would be nice to emit some kind of summary message at the end end of the "performRelease" task.

Here's what we had in bootstrap project in the past. Let's add something like that to Shipkit performRelease task.

```Gradle
performRelease.doLast {
        println """
  Release shipped!
    - Files in Bintray:   https://dl.bintray.com/shipkit-bootstrap/bootstrap/org/shipkit/bootstrap/${System.getProperty("user.name")}/shipkit-bootstrap/$version
    - Bintray repository: https://bintray.com/shipkit-bootstrap/bootstrap/maven/$version
    - Release notes:      TODO"""
}
```

### Code

See how "performRelease" task is configured. Adding doLast action with logging should not be difficult. We can use TaskSuccessfulMessage class.

epeee:
This one seems to be related to #364.
I already started working on it some time ago (exposing the release notes).

szczepiq:
Oups, duplicate :) I will close the other.

wwilk:
@epeee  Do you want to finish this one or is still for "Please contribute"?

epeee:
@wwilk I will most likely start working on it tomorrow.

PR #499, opened 2017-10-08
emit nice summary at the end of the release (release notes)

epeee:
- very first version which includes a link to the release notes

first part of #450 

Did some testing locally:

```
:performRelease

Release shipped!
    - Release notes:      https://github.com/mockito/shipkit/blob/ep2/docs/release-notes.md
```

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/TaskSuccessfulMessage.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/TaskSuccessfulMessage.java
@@ -4,6 +4,7 @@ import org.gradle.api.Action;
 import org.gradle.api.Task;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
+import org.shipkit.internal.notes.util.Supplier;
 
 public class TaskSuccessfulMessage {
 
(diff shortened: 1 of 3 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/Supplier.java
```diff
@@ -0,0 +1,6 @@
+package org.shipkit.internal.notes.util;
+
+public interface Supplier<T> {
```
szczepiq: Can you add a TODO to remove when we migrate to Java8?

I like this new interface. I would introduce it myself :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/TaskSuccessfulMessage.java
```diff
@@ -21,4 +22,13 @@ public void execute(Task task) {
             }
         });
     }
+
+    public static void logOnSuccess(Task task, final Supplier<String> supplier) {
```
szczepiq: Can you add short Javadoc why we need this method (and also in the other method on this class)?

Good stuff. I'm happy that you have figured out how to defer logging the message, until we have all information needed.

szczepiq:
@epeee, do you mind addressing PR comments from my review?

Merged with pleasure!

epeee:
@szczepiq thx for your feedback!
I will apply your suggestions soon.

PR #504, opened 2017-10-09
applied improvements from #499 

epeee:
- public api cleanup
 - added javadoc
 - added todo to remove certain classes once we migrate to java8
 - apply GitBranchPlugin in ReleasePlugin and add task dependency to identifyGitBranch task

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
@@ -128,13 +128,6 @@ public class UpdateReleaseNotesTask extends DefaultTask {
     }
 
     /**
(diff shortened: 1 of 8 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/gradle/notes/UpdateReleaseNotesTask.java
```diff
@@ -128,13 +128,6 @@ public String getGitHubRepository() {
     }
 
     /**
-     * @return a link to the generated release notes file hosted on Github.
-     */
-    public String getReleaseNotesUrl(String branch) {
```
szczepiq: Thank you for reducing the footprint of the public API!

szczepiq:
Merging. Thank you for tidying up after previous PR.

Sorry for late review. I was really busy this week.

Issue #513, opened 2017-10-15
Automated commit message should have CI build ID


mockitoguy:
### Problem

While working on #512 I couldn't work out why the release build was failing. I think it would be really valuable to include CI build ID in the commit message (or even better: the link to CI job).

### Suggested solution

Current:

```
0.9.115 release (previous 0.9.114) + release notes updated [ci skip]
```

Desired:

```
0.9.115 release (previous 0.9.114) + release notes updated. CI job: https://travis-ci.org/mockito/shipkit/builds/288214072 [ci skip]
```

### Implementation

Let's add some code to TravisPlugin that update the commit message postfix. I will put together an experimental branch with some code for someone to get started on this ticket

ArturSkowronski:
@mockitoguy @wwilk I will try to expose that information

ArturSkowronski:
@mockito/developers I'm back and today I'll try to start this task :)

ArturSkowronski:
@mockitoguy as I understand, not all builds will have tags like [ci skip]. Do we want to recognize if [ci skip] is at the end of the commit message and if is put CI job before [ci skip] tag or we are good with just putting "CI job: https://travis-ci.org/mockito/shipkit/builds/288214072" info as a postfix of commit message?

PR #514, opened 2017-10-15
Initial code for improving automated commit message

mockitoguy:
- TravisPlugin updates the commit message postfix and can include build url.
- See issue #513 for details

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/TravisPlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/TravisPlugin.java
@@ -5,11 +5,15 @@ import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.logging.Logger;
 import org.gradle.api.logging.Logging;
+import org.shipkit.gradle.configuration.ShipkitConfiguration;
+import org.shipkit.gradle.git.GitCommitTask;
 import org.shipkit.gradle.git.IdentifyGitBranchTask;
 import org.shipkit.gradle.release.ReleaseNeededTask;
 import org.shipkit.internal.gradle.configuration.BasicValidator;
 import org.shipkit.internal.gradle.configuration.LazyConfiguration;
+import org.shipkit.internal.gradle.configuration.ShipkitConfigurationPlugin;
 import org.shipkit.internal.gradle.git.GitBranchPlugin;
+import org.shipkit.internal.gradle.git.GitPlugin;
 import org.shipkit.internal.gradle.git.GitSetupPlugin;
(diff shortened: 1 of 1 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/TravisPlugin.java
```diff
@@ -74,5 +79,12 @@ public void execute(ReleaseNeededTask t) {
                 t.setPullRequest(isPullRequest);
             }
         });
+
+        GitCommitTask gitCommit = (GitCommitTask) project.getTasks().getByName(GitPlugin.GIT_COMMIT_TASK);
+        //can we construct the url from some Travis env variable?
+        //this needs to be safe for local invocation where there are no travis env variables.
+        String travisJobUrl = "https://travis-ci.org/mockito/shipkit/builds/288214072";
+        String postfix = ". CI job: " + travisJobUrl + " " + conf.getGit().getCommitMessagePostfix();
```
szpak: The first commit will be too long. It would be better to have extra information available in the second (and the following) lines.

szpak: Probably also with the information about the release notes generation.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/TravisPlugin.java
```diff
@@ -74,5 +79,12 @@ public void execute(ReleaseNeededTask t) {
                 t.setPullRequest(isPullRequest);
             }
         });
+
+        GitCommitTask gitCommit = (GitCommitTask) project.getTasks().getByName(GitPlugin.GIT_COMMIT_TASK);
+        //can we construct the url from some Travis env variable?
+        //this needs to be safe for local invocation where there are no travis env variables.
+        String travisJobUrl = "https://travis-ci.org/mockito/shipkit/builds/288214072";
```
szpak: I don't get it. Should it be project specific? PowerMock or Mockito should rather have some other Travis links.

I haven't check it, but couldn't [`TRAVIS_BUILD_ID`](https://docs.travis-ci.com/user/environment-variables/) be used here?

mockitoguy: >I haven't check it, but couldn't TRAVIS_BUILD_ID be used here?

Yes, that's the idea! We won't be hardcoding any url here!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/TravisPlugin.java
mockitoguy: Good idea!

PR #517, opened 2017-10-18
Setting contributors should not break "updateReleaseNotes" task

### LINKED ISSUE #516: Setting contributors explicitly breaks task `updateReleaseNotes`
If the contributors are set explicitly in the ShipKit configuration, the `updateReleaseNotes` task fails with this exception:

    A problem was found with the configuration of task ':updateReleaseNotes'.
    > File '.../build/shipkit/all-contributors.json' specified for property 'contributorsDataFile' does not exist.

See https://github.com/myniva/gradle-s3-build-cache/commit/73789368015a91c003c242967889102f012795d7 and https://travis-ci.org/myniva/gradle-s3-build-cache/builds/289614934 for example.

wwilk:
The problem was caused by the fact that if "shipkit.team.contributors" property is set, fetchContributors task is skipped so "all-contributors.json" is not created. In this case this file is not needed so can be made @Optional but unfortunately Gradle will still throw an exception that it doesn't exist (see https://github.com/gradle/gradle/issues/2919). Therefore I'm setting it to null in this case to be ignored by Gradle validation.

Additionally I applied "base" plugin in Shipkit root project, because there is no clean for it currently, and some files (like 'all-contributors.json') are created in root project's build dir.

Fixes #516

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -13,6 +13,7 @@ buildscript {
 apply from: "$rootDir/gradle/java6-compatibility.gradle"
 apply from: "$rootDir/gradle/precommit.gradle"
 
+apply plugin: 'base'
 apply plugin: 'org.shipkit.gradle-plugin'
 
 allprojects {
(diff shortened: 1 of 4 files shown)


build.gradle
```diff
@@ -13,6 +13,7 @@ buildscript {
 apply from: "$rootDir/gradle/java6-compatibility.gradle"
 apply from: "$rootDir/gradle/precommit.gradle"
 
+apply plugin: 'base'
```
mockitoguy: This is a good idea.

Later, we probably need to apply 'base' inside our org.shipkit.xxx plugins. Otherwise our users might get surprised by not working 'clean'.

PR #521, opened 2017-10-21
Draft of creating metadata file with all dependencies

wwilk:
Due to unreliable pom generation we need a universal way of comparing poms. As discussed with @mockitoguy on Slack, the best way to do is to actually not use poms at all, but instead create a metadata file inside sources jar (we compare the jar anyway) which would contain all dependencies.

Here is the first approach. No tests or documentation, I will create it later, wanted to discuss the implementation first. 

Notes:
- It seems that we have to use resolved dependencies, otherwise eg adding an exclude may not change the content of the file
- gradleApi(), groovyLocal() - they are not included in resolved dependencies, can it be a problem? 
- we have to filter out sibling dependencies the same way we did it in PomFilter, I added a TODO to not forget about it
- I wonder if we can rely on the order of the dependencies, maybe we should them before writing to file? 
- "default" configuration should be enough, it includes runtime and compile. 
- it seems that dependencies added by "files ('some.jar')" will not be visible in resolved dependencies because they are self-resolved. Probably using "resolvedConfiguration.files" and removing not-self-resolved dependencies would be enough to find those files. 

I thought about using the same way that gradle native generating pom task uses, but I would have to rely on SoftwareComponentInternal which, as the name suggests, is not part of the public API.

@mockitoguy waiting for your thoughts!


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +50,37 @@ public void execute(CopySpec copy) {
             }
         });
 
+
+        final File dependenciesFile = new File(project.getBuildDir(), "shipkit-dependencies");
```
mockitoguy: Let's add extension to the file for clarity. I suggest different name, too. "runtime-dependencies", "dependency-information", "dependency-info". I like "dependency-info" most. When you change the name, let's update the task name so that it is consistent with new naming. Consistency is king :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +50,37 @@ public void execute(CopySpec copy) {
...
+        Task copyDepsToFile = TaskMaker.task(project, "createDependenciesFile", CreateDependenciesFile.class, new Action<CreateDependenciesFile>() {
+            @Override
+            public void execute(CreateDependenciesFile task) {
+                task.setDescription("Creates file with resolved runtime dependencies.");
+                task.setOutputFile(dependenciesFile);
+                task.setConfiguration(project.getConfigurations().getByName("default"));
+            }
+        });
+
+        final CopySpec dependencyFile = project.copySpec(new Action<CopySpec>() {
```
mockitoguy: CopySpec feels like an overkill here, why not just adding "from" to the "sourcesJar" task?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +50,37 @@ public void execute(CopySpec copy) {
...
+            public void execute(CopySpec copy) {
+                copy.from(dependenciesFile);
+            }
+        });
+
+
         ((Jar) project.getTasks().getByName("jar")).with(license);
 
         final Jar sourcesJar = project.getTasks().create(SOURCES_JAR_TASK, Jar.class, new Action<Jar>() {
```
mockitoguy: I'm wondering if we shouldn't add the file to the main jar, too. What do you think?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependenciesFile.java
```diff
@@ -0,0 +1,61 @@
...
+import org.gradle.api.tasks.OutputFile;
+import org.gradle.api.tasks.TaskAction;
+import org.shipkit.internal.gradle.util.StringUtil;
+import org.shipkit.internal.notes.util.IOUtil;
+
+import java.io.File;
+
+public class CreateDependenciesFile extends DefaultTask {
+
+    @Input
```
mockitoguy: "InputFiles" annotation is the way of declaring FileCollection / Configuration input in Gradle. "Input" will not work correctly (unless Gradle folks have changed things :).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +50,37 @@ public void execute(CopySpec copy) {
...
             public void execute(Jar jar) {
                 jar.from(JavaPluginUtil.getMainSourceSet(project).getAllSource());
                 jar.setClassifier("sources");
-                jar.with(license);
+                jar.with(license, dependencyFile);
             }
         });
 
+        sourcesJar.dependsOn(copyDepsToFile);
```
mockitoguy: Nice!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependenciesFile.java
```diff
@@ -0,0 +1,61 @@
...
+    private Configuration configuration;
+    @OutputFile
+    private File outputFile;
+
+    @TaskAction public void createFile() {
```
mockitoguy: Might be cool to add information about the content at the top of the file. Perhaps a line that starts with '#' indicating a user comment. Something like:

```
# This file was generated by Shipkit Gradle plugin. See http://..."
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +50,37 @@ public void execute(CopySpec copy) {
...
 
+
+        final File dependenciesFile = new File(project.getBuildDir(), "shipkit-dependencies");
+
+        Task copyDepsToFile = TaskMaker.task(project, "createDependenciesFile", CreateDependenciesFile.class, new Action<CreateDependenciesFile>() {
+            @Override
+            public void execute(CreateDependenciesFile task) {
+                task.setDescription("Creates file with resolved runtime dependencies.");
+                task.setOutputFile(dependenciesFile);
+                task.setConfiguration(project.getConfigurations().getByName("default"));
```
mockitoguy: You might want to use "runtime" configuration. It is more visible and well known. "default" configuration is somewhat internal Gradle configuration.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
wwilk: It seems that we don't need in the main jar at the moment, and we can always add it alter if necessary. Do you agree?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependenciesFile.java
wwilk: Thanks, I didn't know that Configuration is a FileCollection!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
wwilk: What kind of extension did you have in mind? At the moment file doesn't have a common format, just one dependency per line with dependency properties delimeted by ":". It would be easy to make it "csv", a little more effort would be required to make it "json". Do you think it's worth? Maybe it would be more readable for debugging purposes

wwilk: OK JSON will be simple and much more readable, so in my opinion it would be the best choice.

wwilk:
Suggestion applied and additionally included self-resolving deps (files, gradleApi, localGroovy), and excluded sibling projects. I tested it by running **sourcesJar** in shipkit-example.

Configuration for api:
```
testCompile 'junit:junit:4.12'
compile 'com.google.guava:guava:21.0'
```

Configuration for impl:
```
compile project(":api")
compile 'com.google.guava:guava:20.0'
testCompile 'junit:junit:4.13'
compile files("shipkit-0.9.118.jar")
compile gradleApi()
compile localGroovy()
```

Output file:
```
{
	"description":"This file was generated by Shipkit Gradle plugin. See http:\/\/shipkit.org\n\n",
	"dependencies":[
		{
			"files":[
				"shipkit-0.9.118.jar"
			],
			"selfResolved":"true"
		},
		{
			"files":[
				"gradle-api-3.4.jar",
				"groovy-all-2.4.7.jar",
				"gradle-installation-beacon-3.4.jar"
			],
			"selfResolved":"true"
		},
		{
			"files":[
				"groovy-all-2.4.7.jar"
			],
			"selfResolved":"true"
		},
		{
			"artifactType":"jar",
			"dependencyGroup":"com.google.guava",
			"dependencyName":"guava",
			"artifactName":"guava",
			"artifactClassifier":null,
			"artifactExtension":"jar",
			"dependencyVersion":"21.0"
		}
	]
}
```

mockitoguy:
Nice progress!!!

Feedback about the JSON format:
 - I suggest to remove 'self-resolved' dependencies completely. If we bump version of Gradle but we still produce the same binary and the same runtime dependencies, we should skip the release.
 - I suggest to include the file name even for repo dependencies

I haven't yet looked at the code but I will very soon!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -39,23 +42,45 @@
...
+        });
+
         ((Jar) project.getTasks().getByName("jar")).with(license);
 
         final Jar sourcesJar = project.getTasks().create(SOURCES_JAR_TASK, Jar.class, new Action<Jar>() {
             public void execute(Jar jar) {
                 jar.from(JavaPluginUtil.getMainSourceSet(project).getAllSource());
                 jar.setClassifier("sources");
                 jar.with(license);
+                jar.metaInf(new Closure(this) {
```
mockitoguy: Do we need to use Closure magic here? What if we do jar.getMetaInf().from(...)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -39,23 +42,45 @@
...
 
+        final File dependenciesFile = new File(project.getBuildDir(), "dependency-info.json");
+
+        Task createDependencyInfoFileTask = TaskMaker.task(project, "createDependencyInfoFile", CreateDependencyInfoFileTask.class, new Action<CreateDependencyInfoFileTask>() {
+            @Override
+            public void execute(CreateDependencyInfoFileTask task) {
+                task.setDescription("Creates file with resolved runtime dependencies.");
+                task.setOutputFile(dependenciesFile);
+                task.setConfiguration(project.getConfigurations().getByName("runtime"));
+                task.setProjectGroup(project.getGroup().toString());
```
mockitoguy: Is it safe to access project's group outside of the 'deferredConfiguration'? Typically we put it to deferredConfiguration to honor user setting of the group.

I've looked at our builds and I think that we need deferredConfiguration for group for single-project builds. Imagine this sequence of plugins:

```
apply plugin: 'org.shipkit.java'
apply plugin: 'java' // <- adds createDependencyInfoFileTask

group = 'foo' // <- group configured after createDependencyInfoFileTask was added
```

 This is needed for single module projects that apply 'org.shipkit.java' and 'java' plugin in root project.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -0,0 +1,95 @@
...
+\t\t\t"dependencyVersion":null
+\t\t}
+\t]
+}"""
+    }
+
+    private Set<ResolvedArtifact> resolvedArtifacts() {
+        def artifacts = new LinkedHashSet<ResolvedArtifact>()
+
+        def artifact = Mock(ResolvedArtifact)
```
mockitoguy: I suggest to use block style of mocking to save on characters :)

```Groovy
def artifact = Mock(ResolvedArtifact) {
  classifier >> "..."
}
```

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -0,0 +1,95 @@
...
+import org.gradle.api.artifacts.ResolvedModuleVersion
+import testutil.PluginSpecification
+
+class CreateDependencyInfoFileTest extends PluginSpecification {
+
+    CreateDependencyInfoFile createDependencyInfoFile = new CreateDependencyInfoFile()
+
+    def "should create correct output file"() {
+        given:
+        def configuration = Mock(Configuration)
```
mockitoguy: Alternatively, you can add jcenter and resolve the dependencies normally. It would make the test easier to read and probably to maintain, too.

```
project.repositories.jCenter()
def conf = project.configurations.create("conf")
project.dependencies {
  conf "junit:junit:4.10"
  conf "org.mockito:mockito-core:2.10.0"
}
```

Using real dependencies will also allow you to assert on the transitive dependencies (which I leave up to you to decide if we want to support them - I'm ok with that).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
```diff
@@ -0,0 +1,70 @@
...
+
+    private static final String DESCRIPTION = "This file was generated by Shipkit Gradle plugin. See http://shipkit.org\n\n";
+
+    public void createDependencyInfoFile(CreateDependencyInfoFileTask task) {
+        JsonObject result = createJsonObject();
+        result.put("description", DESCRIPTION);
+
+        JsonArray dependencies = createJsonArray();
+
+        for (ResolvedArtifact artifact : task.getConfiguration().getResolvedConfiguration().getResolvedArtifacts()) {
```
mockitoguy: In fact, we only need the 1st level dependencies. I think that the code currently gets you full transitive graph. pom file only contains 1st level dependencies (direct).

However, maybe it's not that bad to include full transitive graph. It might be an interesting piece of information to keep in the jar.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +48,38 @@ public void execute(CopySpec copy) {
...
+        });
+
         ((Jar) project.getTasks().getByName("jar")).with(license);
 
         final Jar sourcesJar = project.getTasks().create(SOURCES_JAR_TASK, Jar.class, new Action<Jar>() {
             public void execute(Jar jar) {
                 jar.from(JavaPluginUtil.getMainSourceSet(project).getAllSource());
                 jar.setClassifier("sources");
                 jar.with(license);
+                jar.getMetaInf().from(dependenciesFile);
```
mockitoguy: Nitpick: You can also inline dependenciesFile variable and use createDependencyInfoFileTask.getOutputFile() here. This helps getting the code simpler - less variables to remember about.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/JavaLibraryPlugin.java
```diff
@@ -46,16 +48,38 @@ public void execute(CopySpec copy) {
             }
         });
 
+        final File dependenciesFile = new File(project.getBuildDir(), "dependency-info.json");
+
+        Task createDependencyInfoFileTask = TaskMaker.task(project, "createDependencyInfoFile", CreateDependencyInfoFileTask.class, new Action<CreateDependencyInfoFileTask>() {
```
mockitoguy: Nitpick: I suggest shorter variable names like "dependencyTask" or "dependencyInfoTask". The code is easier to read (less characters).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTask.java
```diff
@@ -0,0 +1,97 @@
...
+/**
+ * Creates a file {@link #getOutputFile()} with information about all resolved dependencies of the project.
+ * Dependencies are taken from passed {@link #getConfiguration()}.
+ * Results include:
+ * - resolved dependencies
+ * - files
+ * - gradleApi()
+ * - localGroovy()
+ *
+ * They don't include the sibling projects of the same version. Eg. if project A depends on sibling project B
```
mockitoguy: When explaining behavior it is worth describing not only what the code does but also why. In this example: why we are not including sibling dependencies?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
```diff
@@ -0,0 +1,76 @@
...
+        if (!dependency.getArtifacts().isEmpty()) {
+            for (DependencyArtifact artifact : dependency.getArtifacts()) {
+                dependencies.add(getDependencyForSingleArtifact(dependency, artifact));
+            }
+        } else {
+            dependencies.add(getDependencyForSingleArtifact(dependency, null));
+        }
+    }
+
+    private boolean isSiblingDependency(CreateDependencyInfoFileTask task, Dependency dependency) {
```
mockitoguy: Why are we filtering out sibling dependencies? To streamline comparison?

I don't think we should be excluding dependencies on a submodule. What if we add another submodule dependency? This would be a change in pom.xml that should trigger the release.

I suggest to set version to null in that case and document why we are doing it.

```JSON
{
  "version": null,
  "submodule": true
}
```

Also, I suggest to use "submodule" or "subproject" name rather than "sibling". Sibling is a new term whereas "submodule" is something pretty well understood.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
```diff
@@ -0,0 +1,76 @@
...
+
+    private static final String DESCRIPTION = "This file was generated by Shipkit Gradle plugin. See http://shipkit.org\n\n";
+
+    public void createDependencyInfoFile(CreateDependencyInfoFileTask task) {
+        JsonObject result = createJsonObject();
+        result.put("description", DESCRIPTION);
+
+        JsonArray dependencies = createJsonArray();
+
+        for (Dependency dependency: task.getConfiguration().getAllDependencies()) {
```
mockitoguy: I think it is a good idea to work with declared dependencies rather than resolved configuration. I have tested it. I configured dependency resolution rules that force different version. In result, the effective resolved dependency graph had different versions than declared versions. The pom file was still based on declared versions.

Nice catch!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+            runtime child.project(":sibling")
+            runtime "projGroup:sibling:1.1.1"
+            runtime "junit:junit:4.10"
+            runtime "org.mockito:mockito-core:2.10.0:sources"
+            runtime "org.mockito:mockito-core:2.10.0"
+        }
+
+        def runtimeConf = child.configurations.getByName("runtime")
+
+        def createDependencyInfoFileTask = project.tasks.create("createTask", CreateDependencyInfoFileTask)
```
mockitoguy: Shorthand notation (less code!):

```Gradle
project.tasks.create("createTask", CreateDependencyInfoFileTask) {
  outputFile = output
  configuration = runtimeConf
}
```

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+        createDependencyInfoFileTask.configuration = runtimeConf
+        createDependencyInfoFileTask.projectGroup = "projGroup"
+        createDependencyInfoFileTask.currentProjectVersion = "1.2.3"
+
+        when:
+        createDependencyInfoFile.createDependencyInfoFile(createDependencyInfoFileTask)
+
+        then:
+        output.text == """{
+\t"description":"This file was generated by Shipkit Gradle plugin. See http:\\/\\/shipkit.org\\n\\n",
```
mockitoguy: Line breaks "\\n" at the end of description? Intentional?

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -0,0 +1,74 @@
...
+\t"dependencies":[
+\t\t{
+\t\t\t"dependencyGroup":"projGroup",
```
mockitoguy: How about we dedupe the metadata and make it a bit more structured?

```JSON
{
    "description":"This file was...",
    "dependencies":[
        {
            "group": "foo",
            "name": "bar",
            "submodule": false,
            "version": "1.0",
            "artifact": {}
        },
        {
            "group": "foo",
            "name": "bar",
            "submodule": true,
            "version": null,
            "artifact": {}
        },        
        {
            "group": "foo",
            "name": "bar",
            "submodule": false,
            "version": "1.0",
            "artifact": {
                "type": "jar",
                "name": "mockito-core",
                "classifier": "sources",
                "extension": "jar"
            }
        }
    ]
}
```

I was thinking whether it is better to do ```"version": null``` or to skip the version from metadata. Same for "artifact". I'm leaning towards keeping the keys with empty/null values. This way it is more explicit. The users when they see it they can build code around it, rather find out in surprising way when code breaks.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
wwilk: When comparing poms there was a problem that version of submodules was always bumped so pom was never equal. But we can handle at the comparePublications level, this information should be kept here. Good point!

wwilk: I will just remove this comparison, it will simplify the task (no need for group and version as properties) and we can still figure out if it's a submodule or not in comparePublications.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
wwilk: No, thanks for catching this!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
wwilk: OK, I agree, let's keep the null values.

Issue #478, opened 2017-10-04
Bintray/Gradle URLs


wwilk:
# Problem
Generating and customization of URLs, in this case for artifact repository (Bintray or Gradle plugin at the moment), but the similar problem would apply to other areas (eg. GitHub vs other VCS's). 

# Use cases

1.  Release notes - we need an URL of a release in the repository for the badge, eg "https://plugins.gradle.org/plugin/org.shipkit.java/0.9.95" or "https://bintray.com/powermock/maven/powermock/powermock-1.7.3" 

There are some issues with generating it automatically:

- version can be customized for Bintray, eg. powermock uses "powermock-1.7.3" while project version is 1.7.3, in this case version is **tag** + **projectVersion**, but it doesn't have to be. It can be customized in **bintray.pkg.version.name**

- "org.shipkit.java" part of Gradle plugins URL cannot really be automated, especially if there are multiple Gradle plugins in the project

- other parts of Bintray URL can be customized as well (userOrg, repo, packageName), eg in Mockito name = centralRelease? "mockito" : "mockito-development"

- at the moment we allow generating release notes for multiple versions. So what happens if you have a customized version name, we would need some kind of version resolver too.  

2. URL of arbitrary version artifact for comparing publications. This may be easier to automate but in some cases (like with Mockito different packageName for different versions) it can be hard.

# Possible solutions

1. OO approach - using a factory returning a correct type of UrlResolver depending on applied plugins (eg. if BintrayPlugin applied -> BintrayResolver)

Really nice solution for Shipkit developers, easily extendible and testable. The problem is with any customizations, user would have to write their own implementations of url resolvers. It is not that hard but also doesn't look so good in Gradle configuration files. 

2. Pattern with placeholders, eg "https://bintray.com/{userOrg}/{repository}/{package}/{version}". Easily customizable by the user, but we would have to maintain a lot of different placeholders (eg.  {bintrayVersionName}). So it doesn't scale.

3. Closure returning the correct URL, which you could configure like that:
shipkit.releaseNotes.repositoryUrl = {
      https://bintray.com/${bintray.pkg.userOrg}/repo/package/${bintray.pkg.version.name}
}

Easy to customize but harder to test and maintain by us. But what happens when generating release notes for different versions? Version would have to be an argument of closure which makes it more complicated.

[...]

PR #522, opened 2017-10-22
First step of fixing the problem with generating Gradle/Bintray URLs

wwilk:
First step of #478. It solves the problem for Powermock, so that they could use "powermock-{VERSION}" in the release notes. When it's merged I will raise a PR in powermock to fix this issue.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -254,8 +254,8 @@ public class ShipkitConfiguration {
         }
 
         /**
-         * Set the Publication Repository where we look for your published binary.
-         * Version will be concatenated to it.
+         * Set the Publication Repository where we look for your published binary. Version will be concatenated to it.
+         * It is currently used to configure repository Badge URL when generating release notes.
          * E.g.
          * <pre>
          *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/"
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -255,10 +255,13 @@ public class ShipkitConfiguration {
 
         /**
          * Set the Publication Repository where we look for your published binary.
+         * Can contain placeholder {VERSION} which will be replaced with a currently built version.
          * E.g.
          * <pre>
-         *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
+         *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/{VERSION}"
          * </pre>
+         * For version = "1.2.3" will result in:
+         * "https://plugins.gradle.org/plugin/org.shipkit.java/1.2.3"
          * This will be used for adding and linking the repository in the release notes.
          *
          * @see #getPublicationRepository()
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -255,10 +255,10 @@ public class ShipkitConfiguration {
 
         /**
          * Set the Publication Repository where we look for your published binary.
-         * Can contain placeholder {VERSION} which will be replaced with a currently built version.
+         * Version will be concatenated to it.
          * E.g.
          * <pre>
-         *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/{VERSION}"
+         *   releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/"
          * </pre>
          * For version = "1.2.3" will result in:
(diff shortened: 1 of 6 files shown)


mockitoguy:
The powermock use case is already handled. We can do something like that:

```
shipkit {
  releaseNotes.publicationRepository = "https://bintray.com/powermock/maven/powermock/powermock-$version"
}
```

Why do we need the "{VERSION}" placeholder?

wwilk:
Placeholder instead of concatenation should be easier to use and more self-explanatory. If you see "http://https://bintray.com/powermock/maven/powermock/powermock-" in the code, you don't really know that version will be concatenated to it. It's quite clear when you see {VERSION} placeholder. I thought that maybe in the future we would be able to add more placeholders. 

I'm not so sure about it, but I don't like current concatenation
```
"publicationRepository` + "/" + version"
```
because if you want to use it, you have to check if "/" is added in the Shipkit code or not. In case of placeholder you don't really care about this.

What do you think @mockitoguy ?

mockitoguy:
My gut feel is that it is an unnecessary complexity.

1. Currently the user can do this:
```
shipkit {
  releaseNotes.publicationRepository = "https://bintray.com/powermock/maven/powermock/powermock-$version"
}
```

2. This PR makes it so that the user can do that:
```
shipkit {
  releaseNotes.publicationRepository = "https://bintray.com/powermock/maven/powermock/powermock-{VERSION}"
}
```

The ROI is minimal, if any. Both look clean to me.

Approach 1) is based on well known GString variable replacement + well known project property "version". Approach 2) introduces a brand new style of token replacement that needs to be implemented, documented and maintained. Users might even try to use "{VERSION}" keyword in other Shipkit properties...

That's my honest feedback :) I'm curious what do other contributors think?

epeee:
I'd vote for Approach 1 (since it is well known and we do not have extra effort).

wwilk:
OK let's keep concatenation, but I think it would be better to avoid adding a version with slash "/". Otherwise it's not really possible to have a version like "powermock-1.2.3", it would be "powermock-/1.2.3". So I would like to still change this line:
https://github.com/mockito/shipkit/blob/514ac7a43af0e3513a83f7518221086b203d4b45/subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/format/DetailedFormatter.java#L82
to "publicationRepository + version". And also encapsulating this part in a separate class. 
OK with you guys? @mockitoguy @epeee ?
Btw because of that line, user can't really do sth like "powermock-$version". It would result in "powermock-1.2.3/1.2.3"

mockitoguy:
Oh, I didn't see that it wouldn't work!

Let's simplify this. Let's remove version concat completely and have the formatter just use the "publicationRepository".

This way, the following will work:

```
//powermock
shipkit {
  releaseNotes.publicationRepository = "https://bintray.com/powermock/maven/powermock/powermock-$version"
}

//shipkit
shipkit {
  releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/$version"
}
```

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -255,10 +255,13 @@ public void setIgnoreCommitsContaining(Collection<String> commitMessageParts) {
 
         /**
          * Set the Publication Repository where we look for your published binary.
```
mockitoguy: Let's write what do we use it for:
 - release notes generation
 - getting binaries for comparing with previous version

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/artifact/DefaultArtifactUrlResolverFactory.java
```diff
@@ -6,6 +6,8 @@
 
 public class DefaultArtifactUrlResolverFactory {
 
+    private static final String VERSION_PLACEHOLDER = "{VERSION}";
```
mockitoguy: Unused constant

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/DownloadPreviousPublications.java
```diff
@@ -25,7 +25,9 @@ private void downloadRemoteFile(String remoteUrl, File localFile) {
             IOUtil.downloadToFile(remoteUrl, localFile);
         } catch (Exception e) {
             LOG.lifecycle("  Unable to download, ignoring. Run with '-d' for stack trace. Url: {}", remoteUrl);
-            LOG.debug("Unable to download, ignoring", e);
+            LOG.debug("Unable to download, ignoring." +
```
mockitoguy: This should probably be a "lifecycle" message, right?

Suggestion: 
 - lifecycle message with general info + a hint how to resolve the issue
 - debug message with stack trace

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/format/DetailedFormatter.java
```diff
@@ -46,7 +47,7 @@ public String formatReleaseNotes(Collection<ReleaseNotesData> data) {
         for (ReleaseNotesData d : data) {
             sb.append(header(d.getVersion(), d.getDate(), emphasizeVersion));
             String vcsCommitsLink = MessageFormat.format(vcsCommitsLinkTemplate, d.getPreviousVersionVcsTag(), d.getVcsTag());
-            sb.append(releaseSummary(d.getVersion(), d.getDate(), d.getContributions(), contributors, vcsCommitsLink, publicationRepository));
```
mockitoguy: Nice! Thank you for removing dead parameter.

PR #525, opened 2017-10-27
Integration test for InitPlugin

wwilk:
I went through TODOs in code on a plane today and I figured this test is important, it assures that you can apply Shipkit plugin in empty project and run initShipkit on it.

### DIFF
--- a/subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/init/InitPluginIntegTest.groovy
+++ b/subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/init/InitPluginIntegTest.groovy
@@ -1,0 +1,30 @@
+package org.shipkit.internal.gradle.init
+
+import org.gradle.testkit.runner.BuildResult
+import testutil.GradleSpecification
+
+class InitPluginIntegTest extends GradleSpecification {
+
+        def "runs initShipkit task in a project without any Shipkit configuration"() {
+            given:
+            projectDir.newFolder("gradle")
+
+            buildFile << """
+            apply plugin: "org.shipkit.java"
+        """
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/init/InitPluginIntegTest.groovy
```diff
@@ -0,0 +1,30 @@
+package org.shipkit.internal.gradle.init
+
+import org.gradle.testkit.runner.BuildResult
+import testutil.GradleSpecification
+
+class InitPluginIntegTest extends GradleSpecification {
+
+        def "runs initShipkit task in a project without any Shipkit configuration"() {
+            given:
+            projectDir.newFolder("gradle")
```
mockitoguy: I don't think this is needed. If "gradle" dir does not exist, initialization should work.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/init/InitPluginIntegTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+
+            buildFile << """
+            apply plugin: "org.shipkit.java"
+        """
+
+            expect:
+            BuildResult result = pass("initShipkit", "-s")
+
+            result.tasks.collect { it.path }.join("\n") == """:identifyGitOrigin
```
mockitoguy: I would not assert on the tasks because it is enough to verify that the files were generated. Less code to read and maintain. However, I will leave it up to you :)

wwilk:

PR #538, opened 2017-10-29
Comparing publications by dependency-info.json instead of pom.xml

wwilk:
We can now use dependency-info.json for comparing publications. The file started to be generated in Shipkit 0.9.122.

Tested in shipkit and shipkit-example by running "./gradlew assertReleaseNeeded".


mockitoguy:
Can you write an overview of the change. I have a couple of questions :)

1. Why do we need to write any code to parse the dependency-info.json file? If we ensure that the dependency-info.json contents are written in a consistent way we can leverage an existing comparison / diff logic. I'd hope that we can reduce the complexity as much as possible and avoid writing a parser / comparator for the file.
2. Do we need to introduce more zip parsing logic? I thought that we already have all the code in place for that. In the PR it looks like new zip logic was introduced. Again, it seems to me that extra complexity can be avoided. We can also use Gradle's zip features to avoid maintaining our own code.
3. dependency-info.json looks untidy with an empty artifact and is inconsistent with Gradle model. AFAIR, Gradle returns 'null' when there is no artifact. Let's change the metadata format to something like this:

```
{
	"artifact": null,
	"name":"diffutils",
	"version":"1.3.0",
	"group":"com.googlecode.java-diff-utils"
}	
```
4. To simplify comparison, perhaps we want to avoid JSON and have a flat list / yml. This way, we don't ever parse the file, we diff it textually as every other file in the sources jar. Examples:

- yaml:
```YAML
description: xxx		
dependencies:
  # standard dependencies:
  - com.googlecode.java-diff-utils:diffutils:1.3.0
  - com.gradle.publish:plugin-publish-plugin:0.9.7
  # internal dependencies:
  - com.gradle.publish:plugin-publish-plugin
  # dependency with artifact
  - dependency:
    module: foo:bar:1.0
    artifacts: 
        #name:classifier:type:extension notation for artifacts
        - foo:sources:jar:jar        
```

- just text:
```
#description xxx

#dependencies
com.googlecode.java-diff-utils:diffutils:1.3.0
com.gradle.publish:plugin-publish-plugin:0.9.7

#internal dependency, no version
foo:bar

#dependency with artifact
foo:bar:1.0
  #name:classifier:type:extension notation for artifacts
  foo:sources:jar:jar
```

BTW, the code is greatly simplified when we don't have to parse the POM. BIG THANKS for working on this!!!

wwilk:
OK, I changed the format of dependency-info and created a separate PR for that #543, to keep it simpler. When it's merged I will update this one.

wwilk:
Dependency-info.md now used. 

Example output in shipkit when added one dependency and changed version of another: 

```
Differences between files:
  --- /Users/wilk/Projects/shipkit/subprojects/shipkit/build/previous-release-artifacts/shipkit-0.9.130-sources.jar
  +++ /Users/wilk/Projects/shipkit/subprojects/shipkit/build/libs/shipkit-0.9.131-sources.jar

    Added files:
    ++ org/shipkit/internal/comparison/DependencyInfoComparator.java
    ++ org/shipkit/internal/gradle/util/ZipUtil.java

    Removed files:
    -- org/shipkit/internal/comparison/PomComparator.java

    Modified files:
    +- org/shipkit/gradle/java/ComparePublicationsTask.java
    +- org/shipkit/gradle/java/DownloadPreviousPublicationsTask.java
    +- org/shipkit/internal/comparison/ZipComparator.java
    +- org/shipkit/internal/comparison/diff/Diff.java
    +- org/shipkit/internal/comparison/diff/FileDiffGenerator.java
    +- org/shipkit/internal/gradle/java/ComparePublicationsPlugin.java
    +- org/shipkit/internal/gradle/java/tasks/ComparePublications.java
    +- org/shipkit/internal/gradle/java/tasks/DownloadPreviousPublications.java


  Differences between files:
  --- /Users/wilk/Projects/shipkit/subprojects/shipkit/build/previous-release-artifacts/shipkit-0.9.130-sources.jar/META-INF/dependency-info.md
  +++ /Users/wilk/Projects/shipkit/subprojects/shipkit/build/libs/shipkit-0.9.131-sources.jar/META-INF/dependency-info.md

  Here you can see the changes in declared dependencies between versions.

    @@ -7,0 +7,1 @@
    + - com.googlecode.json-simple:json-simple:1.1.1^M
    @@ -8,1 +9,1 @@
    - - com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.3
    + - com.jfrog.bintray.gradle:gradle-bintray-plugin:1.7.1
```

I didn't change anything about comparing sources jars, but still here for you to see the whole picture.

About the zip complexity - I extracted typical zip operations into ZipUtil class.
It's a little harder because I would like to check in comparePublications if dependency-info.md is present in previous sources jar, to keep the backwards compatibility. For this I have to open the zip file. 

I have to open it in two other places - ZipComparator and DependencyInfoComparator, because I need to extract dependency-info.md from jar. I don't like passing around open zip file, so I prefer to open and parse it in a few different places, even if it has a slight performance impact.

mockitoguy:
>About the zip complexity - I extracted typical zip operations into ZipUtil class.
It's a little harder because I would like to check in comparePublications if dependency-info.md is present in previous sources jar, to keep the backwards compatibility. For this I have to open the zip file.

hmmm, I suspect that we can simplify the implementation. For example, during zip comparison you can notify some listener that some file is different in zips and here're the input streams. The listener implementation could do content comparison for clean diff.

In general, I'd shy away from creating "special" implementations for diffing a particular file from the zip.

I'm tempted to skip dependency-info comparison. We don't do it for other files in the jar, why do we bother for this one? However, since we already have the code, let's keep it because it can potentially improve developer experience in some scenarios.

BTW. the output looks amazing!!! Thank you!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/DependencyInfoComparator.java
```diff
@@ -0,0 +1,53 @@
...
+import org.gradle.api.logging.Logging;
+import org.shipkit.internal.comparison.diff.Diff;
+import org.shipkit.internal.comparison.diff.FileDiffGenerator;
+import org.shipkit.internal.gradle.java.tasks.ComparePublications;
+
+import java.io.File;
+
+import static org.shipkit.internal.util.ArgumentValidation.notNull;
+
+public class DependencyInfoComparator {
```
mockitoguy: I gave some feedback about this guy in the PR comments. Perhaps we can simplify the implementation / make it more generic?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/PomComparator.java
```diff
@@ -1,58 +0,0 @@
...
-import org.gradle.api.logging.Logging;
-import org.shipkit.internal.comparison.diff.Diff;
-import org.shipkit.internal.comparison.diff.FileDiffGenerator;
-import org.shipkit.internal.notes.util.IOUtil;
-
-import java.io.File;
-
-import static org.shipkit.internal.util.ArgumentValidation.notNull;
-
-public class PomComparator {
```
mockitoguy: I'm very happy to see this code go :)

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/java/DownloadPreviousPublicationsTask.java
```diff
@@ -14,10 +14,8 @@
  */
 public class DownloadPreviousPublicationsTask extends DefaultTask {
```
mockitoguy: Can you update the Javadoc of this task? Please try to be as comprehensive as possible!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/ComparePublicationsPlugin.java
```diff
@@ -49,6 +49,7 @@
```
mockitoguy: I reviewed the existing Javadoc for this plugin and it is splendid :)

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/java/ComparePublicationsTask.java
```diff
@@ -14,7 +14,7 @@
 import java.io.File;
 
 /**
- * Compares sources jars and pom files produced by the build with analogical artifacts
+ * Compares sources jars and {@link ComparePublications#DEPENDENCY_INFO_FILEPATH} files produced by the build with analogical artifacts
```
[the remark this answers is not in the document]
wwilk: Done

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/java/DownloadPreviousPublicationsTask.java
wwilk: Done

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/DependencyInfoComparator.java
wwilk: I made it StringComparator, and moved all the complexity from this class (also from ComparePublications) regarding formatting the output to a new class ComparePublicationsResultFormatter.

wwilk:
I created a new class ComparePublicationsResultFormatter and moved there all the complexity regarding the above comparison output. Also instead of DependencyInfoComparator we have a generic StringComparator

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/ComparePublications.java
```diff
@@ -2,65 +2,58 @@
...
         File currentVersionSourcesJarFile = task.getSourcesJar().getArchivePath();
 
-        LOG.lifecycle("{} - about to compare publications, for versions {} and {}",
-                task.getPath(), task.getPreviousVersion(), task.getCurrentVersion());
+        LOG.lifecycle("{} - about to compare publications",
+                task.getPath());
+
+        if (!ZipUtil.fileContainsEntry(task.getPreviousSourcesJar(), DEPENDENCY_INFO_FILEPATH)) {
+            LOG.lifecycle("{} - previous {} file not found, nothing to compare, skipping", task.getPath(), DEPENDENCY_INFO_FILEPATH);
+            return;
```
mockitoguy: I don't think we should bail out here. What if someone disables dependency-info.json generation? We still want to compare the sources jar. Thoughts?

mockitoguy: I don't think we should bail out here. What if someone disables dependency-info.json generation? We still want to compare the sources jar. Thoughts?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/comparison/ZipComparator.java
```diff
@@ -58,15 +68,15 @@ Diff compareZips(File previousFile, File currentFile) {
             }
             try {
                 current.remove(name);
-                if (!streamsEqual(file1.getInputStream(file1.getEntry(name)), file2.getInputStream(file2
-                        .getEntry(name)))) {
+                if (!streamsEqual(previousFile.getInputStream(previousFile.getEntry(name)),
+                                  currentFile.getInputStream(currentFile.getEntry(name)))) {
                     changedFiles.add(name);
```
mockitoguy: This is a good place to notify some listener that a file has changed between source jars. The listener receives info: file name and 2 contents. The listener can check if the file is interesting (e.g. 'dependency-info.json') and start building a nice diff.

Next time we can do it this way :) I think it would simplify the algorithm and avoid redundant work. This way, we could avoid the "previous.remove(ComparePublications.DEPENDENCY_INFO_FILEPATH);" code.

wwilk: I added TODO in code addressed at me. Also for unit tests for ComparePublications class.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/ComparePublications.java
wwilk: Good point! For now I will print that there were no differences (only for dep-info comparison) if this file is not present.

PR #543, opened 2017-10-30
Dependency-info format changed from json to txt

wwilk:
I completely agree that we shouldn't parse dependency-info when comparing publications. The only problem is with these submodules dependencies. 

I thought at the beginning to exclude them from dependency-info but you pointed out correctly that we are losing important information.  Then I thought of removing it at the point of comparing publications, as proposed in #538 , but I had to parse JSON and additionally there was a problem with JSON ordering (since json-simple internally uses hash maps which don't guarantee order). 

So I like the idea of using a different structure - let's go with a simple and concise txt format, similar to the one in which dependencies are declared.

About submodules I really like the idea of not saving the version if the submodule is in the same version as the project being built. This way there won't be a difference when comparing publications. Let's still keep the version if it's different though. 

I also moved "createDependencyInfoFileTask" config from JavaLibrary to ComparePublicationsPlugin.

I tested it by running ./gradlew clean and then ./gradlew sourcesJar in shipkit-example. 

Result for impl:
```
#description
This file was generated by Shipkit Gradle plugin. See http://shipkit.org.

#dependencies
#group : name : version @@@ extension : classifier : artifactName : type
org.mockito.shipkit-example : api
com.google.guava : guava : 20.0
```


wwilk:
As discussed on Slack with @mockitoguy format changed to markdown with artifacts nested deeper than dependencies.

Here is an example output file from unit test (md so it's formatted below):

# Description
This file was generated by Shipkit Gradle plugin. It contains all declared dependencies of the project. See http://shipkit.org.
The format of dependencies is: group:name:version
Each of the dependencies may contain artifacts (more nested) formatted like: artifactName:classifier:type:extension

# Dependencies
 - projGroup:submodule
 - projGroup:submodule2:1.0.0
 - junit:junit:4.10
 - org.mockito:mockito-core:2.10.0
   - mockito:jar:sources:jar
   - mockito:jarClass:jar:jarExt

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
```diff
@@ -1,74 +1,76 @@
...
-    private static final String DESCRIPTION = "This file was generated by Shipkit Gradle plugin. See http://shipkit.org.";
+    public static final String SEPARATOR = ":";
+    public static final String NEWLINE = "\r\n";
+    public static final String DEPENDENCY_INDENT = NEWLINE + " - ";
+    public static final String ARTIFACT_INDENT = NEWLINE + "   - ";
 
-    public void createDependencyInfoFile(CreateDependencyInfoFileTask task) {
-        JsonObject result = createJsonObject();
-        result.put("description", DESCRIPTION);
+    private static final String DESCRIPTION = "This file was generated by Shipkit Gradle plugin. " +
```
mockitoguy: Let's make the textual description as short as possible. It will be hard to change later on. It's better to link to our docs instead (I just added http://dependency-info.shipkit.org subdomain to point to our docs ;):

```
# Description
Generated by Shipkit - http://dependency-info.shipkit.org
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
```diff
@@ -1,74 +1,76 @@
...
 
-        result.put("dependencies", dependencies);
+        result += DEPENDENCY_INDENT + StringUtil.join(dependencies, DEPENDENCY_INDENT);
 
-        IOUtil.writeFile(task.getOutputFile(), Jsoner.prettyPrint(result.toJson()));
+        IOUtil.writeFile(task.getOutputFile(), result.toString());
     }
 
-    private void addResolvedDependency(JsonArray dependencies, ModuleDependency dependency) {
+    private String getResolvedDependency(CreateDependencyInfoFileTask task, ModuleDependency dependency) {
```
mockitoguy: I suggest to rename this method because we no longer work on resolved dependencies.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTask.java
```diff
@@ -52,4 +58,31 @@ public void setOutputFile(File outputFile) {
         this.outputFile = outputFile;
     }
 
+    /**
+     * See {@link Project#getGroup()}
+     */
+    public String getProjectGroup() {
```
mockitoguy: Can you go through all the public methods and the Javadoc on the class type and ensure it has the right level of detail? For example: how does the task use the 'group',  'version', and 'configuration' properties? The class javadoc should explain the details like the distinction between internal module dependencies and external dependencies and how we use version.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFileTest.groovy
```diff
@@ -12,17 +12,30 @@ class CreateDependencyInfoFileTest extends PluginSpecification {
...
-            runtime "org.mockito:mockito-core:2.10.0:sources"
-            runtime "org.mockito:mockito-core:2.10.0"
+            runtime (group: 'org.mockito', name: 'mockito-core', version: '2.10.0') {
+                artifact {
+                    name = 'mockito'
+                    extension = 'jar'
+                    type = 'sources'
+                    classifier = 'jar'
+                }
+                artifact {
```
mockitoguy: Nice! Thank you for adding coverage for that!

Can you spice up the test data and change/add artifact that does not declare all fields? (e.g. has nulls)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/java/tasks/CreateDependencyInfoFile.java
wwilk: Cool, thanks for the subdomain, we can redirect to docs for this file later :)

PR #547, opened 2017-10-31
Fix #426 - Only cache jar, pom and xml files in .gradle/caches

### LINKED ISSUE #426: Template travis.yml cache always changed
The travis.yml shipkit sets up results in a cache which is always changed, slowing down builds.

Travis recommend [removing some files](https://docs.travis-ci.com/user/languages/java/#Caching), but this seems to be out of date too as it still resulted in changed on every build with gradle 4.2.

Copying the [cache section from this project](https://github.com/mockito/shipkit/blob/master/.travis.yml#L25) still didn't work.

Also removing SNAPSHOT dependencies and metadata-*/*.bin seems to work.

```yml
before_cache:
# SNAPSHOT dependency related files were being updated for every build even with no actual change to the dependency
  - find $HOME/.gradle/caches/modules-2 -name \*-SNAPSHOT -type d -exec rm -rf {} +
# these are changed every build and can't be cached
  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
  - rm -f  $HOME/.gradle/caches/modules-2/metadata-*/*.bin
  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
  - rm -fr $HOME/.gradle/caches/*/scripts/
  - rm -fr $HOME/.gradle/caches/*/scripts-remapped/
  - rm -fr $HOME/.gradle/caches/*/fileHashes/
  - rm -fr $HOME/.gradle/caches/transforms-1/transforms-1.lock
cache:
  directories:
    - $HOME/.gradle/caches/
    - $HOME/.gradle/wrapper/
```

The cache will also get bigger if the gradle wrapper version is changed. It'd be good to either automatically clear out older versions in that case somehow, or to add documentation recommending clearing the travis cache after changing the gradle version.

nallar:
Fixes #426.

Same change working for one of my projects - https://travis-ci.org/MinimallyCorrect/DefaultsPlugin/builds/295473198

Testing builds of this project by triggering PR build multiple times and checking that cache isn't changed.

----

[Test build 1 - cache changes due to removed .lock files, expected](https://travis-ci.org/mockito/shipkit/builds/295478474)  
[Test build 2 - has cache changes, should have no cache changes](https://travis-ci.org/mockito/shipkit/builds/295483481) ❌  
[Test build 3 - cache changes due to removing gradle-worker.jar and ivy.xml](https://travis-ci.org/mockito/shipkit/builds/295489863)  
[Test build 4 - has no cache changes, should have no cache changes](https://travis-ci.org/mockito/shipkit/builds/295493206) ✔️  
[Test build 5 - cache changes due to removing lots of shipkit-VERSION.jar from modules-2](https://travis-ci.org/mockito/shipkit/builds/295498115)  
[Test build 6 - has no cache changes, should have no cache changes](https://travis-ci.org/mockito/shipkit/builds/295501139)✔️

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,13 +23,19 @@ script:
 
 # Remove often changing files to prevent cache re-upload on no changes in dependencies
 before_cache:
-  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
-  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
-  - rm -fr $HOME/.gradle/caches/*/scripts/
-  - rm -fr $HOME/.gradle/caches/*/scripts-remapped/
-  - rm -fr $HOME/.gradle/caches/*/fileHashes/
-  - rm -fr $HOME/.gradle/caches/transforms-1/transforms-1.lock
+# pom, xml and jar files are known not to change
+# assume non-whitelisted extensions are changing
+# (this will include .lock, .bin, .metadata and potential future additions to gradle's caches folder)
+  - find $HOME/.gradle/caches -type f -not \( -iname \*.pom -o -iname \*.jar -o -iname \*.xml \) -delete
+# jar files which change
+  - rm -fr $HOME/.gradle/caches/*/workerMain/gradle-worker.jar
+# xml files which change
+  - find $HOME/.gradle/caches/modules-2 -name ivy.xml -delete
+# specific to shipkit build
   - rm -fr $HOME/.gradle/caches/jars-*/*/shipkit-*.jar
+  - find $HOME/.gradle/caches/modules-2 -name shipkit-\*.jar -delete
+# remove left over empty directories
+  - find $HOME/.gradle/caches -type d -empty -delete
 cache:
   directories:
     - $HOME/.gradle/caches/
--- a/subprojects/shipkit/src/main/resources/template.travis.yml
+++ b/subprojects/shipkit/src/main/resources/template.travis.yml
@@ -7,6 +7,19 @@ cache:
     - $HOME/.gradle/caches/
     - $HOME/.gradle/wrapper/
 
+# Remove often changing files to prevent cache re-upload on no changes in dependencies
+before_cache:
+# pom, xml and jar files are known not to change
+# assume non-whitelisted extensions are changing
+# (this will include .lock, .bin, .metadata and potential future additions to gradle's caches folder)
+  - find $HOME/.gradle/caches -type f -not \( -iname \*.pom -o -iname \*.jar -o -iname \*.xml \) -delete
+# jar files which change
+  - rm -fr $HOME/.gradle/caches/*/workerMain/gradle-worker.jar
+# xml files which change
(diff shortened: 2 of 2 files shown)


.travis.yml
```diff
@@ -23,13 +23,19 @@ script:
...
+# assume non-whitelisted extensions are changing
+# (this will include .lock, .bin, .metadata and potential future additions to gradle's caches folder)
+  - find $HOME/.gradle/caches -type f -not \( -iname \*.pom -o -iname \*.jar -o -iname \*.xml \) -delete
+# jar files which change
+  - rm -fr $HOME/.gradle/caches/*/workerMain/gradle-worker.jar
+# xml files which change
+  - find $HOME/.gradle/caches/modules-2 -name ivy.xml -delete
+# specific to shipkit build
   - rm -fr $HOME/.gradle/caches/jars-*/*/shipkit-*.jar
+  - find $HOME/.gradle/caches/modules-2 -name shipkit-\*.jar -delete
```
nallar: Should catch `/home/travis/.gradle/caches/modules-2/files-2.1/org.shipkit/shipkit/XXXX/c6fdcc7067e6182eb8c36fb52359a0abf82c0a6b/shipkit-XXXX.jar`

Otherwise every release changes the cache when testing downstream projects

mockitoguy:
Thank you for working on this!

I'm hesitant to merge it because it adds a lot of complexity to script and the ROI is unknown.

I've looked at recent builds on master:

9:00
5:53
7:32
11:50
7:00
8:50
7:35

Travis CI build for this PR:

7:40
6:40

What are the build times when we turn off the build completely? I'm not too worried about failed downloads from JCenter. If we hit them frequently after turning off the cache we can always bring back the cache :)

Can you show the numbers if we don't use the cache? (or just cache Gradle distros at ~/.gradle/wrapper/*)

Looking forward to hearing from you. Let's close it down. Either by removing the cache or by accepting added complexity and living with it :)

nallar:
The "store build cache" action which this avoids takes ~20s for my project. As the total build time is ~120 seconds that's nearly a 20% improvement.

Shipkit's builds take much longer but the cache is also bigger. For example, the first shipkit build with a cache change there records `85.23s` uploading the cache.

mockitoguy:
Fresh datapoints:

- On my local machine, full build with cold Gradle cache is 1m 38s
- Recent build on CI took 6m 48sec. In the log I can see that cache related operations take
33s, 33s, 137s, total: 3m 28s
- I opened a PR #668 to remove cache, let's how long that build will take

Travis cache seems to be providing marginal value. Yet it is an additional complexity that we need to maintain and review periodically. The format of Gradle cache is internal and can change. When it changes, in order to keep our settings effective, we should review and update them. Who is going to remember about it and have time for this?

Performance wise, I don't see enough value from the Travis Cache. If it helps with stability of builds, we can bring the settings back.

@nallar! I appreciate your effort digging into this problem. At this time, we want to avoid additional complexity. I'll close this PR without merging. We would love more contributions from you!!!

PR #552, opened 2017-11-03
Shipkit version downgraded to the last working one and CloneGitRepositoryTask fixed

wwilk:
CloneGitRepositoryTask failed because we have two separate processes that use it (testDownstream and performRelease) in Travis script. I added a check if directory is empty, because it doesn't make sense to clone it again when running "upgradeDownstream". I also improved documentation of the task, so that it's clear how it behaves.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java
@@ -18,8 +18,11 @@ import java.util.List;
 import static java.lang.String.valueOf;
 
 /**
- * This task clone git project from repository to target dir.
- * It support clone from remote server and from local filesystem.
+ * This task clones git project from {@link #getRepositoryUrl()} to {@link #getTargetDir()}.
+ * It supports clone from remote server and from local filesystem.
+ * The task execution is skipped if {@link #getTargetDir()} exists and is not empty.
+ * If you want to always execute it, use clean task before calling this one.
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java
```diff
@@ -37,7 +40,12 @@
 
     @TaskAction
     public void cloneRepository() {
+        if (!isTargetEmpty()) {
+            LOG.lifecycle("{} -  Target dir {} already exists and is not empty. Skipping execution of the task.");
```
mockitoguy: You already have found out that 'return' is missing here :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java
```diff
@@ -120,4 +128,8 @@ public int getDepth() {
     public void setDepth(int depth) {
         this.depth = depth;
     }
+
+    private boolean isTargetEmpty() {
+        return !targetDir.exists() || targetDir.list().length == 0;
```
mockitoguy: I'm curious if this is our long term strategy for clone task. Let's flesh out all the use cases and identify what should be the desired behavior of the clone task. For example:

- for given external repo, there should be 1 clone repo task for this repo
- if the outputs were tinkered with, we should clear the contents and clone the repo again
- if the outputs are the same, the task is UP-TO-DATE

Perhaps the problem with clone tasks that you have encountered is a sign that we should push harder on integrating test-downstream + upgrade-downstream plugin.

PR #555, opened 2017-11-03
Shipkit version downgraded to the last working one + CloneGitReposito…

wwilk:
…ryTask fixed


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/CloneGitRepositoryTask.java
```diff
@@ -42,6 +42,7 @@
     public void cloneRepository() {
         if (!isTargetEmpty()) {
             LOG.lifecycle("{} -  Target dir {} already exists and is not empty. Skipping execution of the task.");
+            return;
```
mockitoguy: This should be easily unit testable.

PR #561, opened 2017-11-05
fix user used for shipkit version upgrade

### LINKED ISSUE #500: [upgrade-downstream] automated commit uses Travis CI user
### Problem

Automated commit triggered by upgrade-downstream plugin uses Travis CI user. In comparison, the automated commit for release notes is cleanly attributed to shipkit-org. See the screen shot from [commits page](https://github.com/mockito/shipkit/commits/master).

screen shot 2017-10-08 at 1 30 33 pm

### Solution

Let's make our automated commits consistent.

epeee:
fixes #500 

basically, I extracted all the git config related task to a separate plugin and made use of it in GitSetupPlugin and UpgradeDependencyPlugin.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitConfigPlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitConfigPlugin.java
@@ -1,0 +1,37 @@
+package org.shipkit.internal.gradle.git;
+
+import org.gradle.api.Action;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.shipkit.gradle.configuration.ShipkitConfiguration;
+import org.shipkit.gradle.exec.ShipkitExecTask;
+import org.shipkit.internal.gradle.configuration.ShipkitConfigurationPlugin;
+import org.shipkit.internal.gradle.exec.ExecCommandFactory;
+import org.shipkit.internal.gradle.util.TaskMaker;
+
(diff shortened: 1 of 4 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitConfigPlugin.java
```diff
@@ -0,0 +1,59 @@
...
+ *         'setGitUserName' - sets generic user name so that CI server can commit code as neatly described robot,
+ *         uses value from {@link org.shipkit.gradle.configuration.ShipkitConfiguration.Git#getUser()}
+ *     </li>
+ *     <li>
+ *         'setGitUserEmail' - sets generic user email so that CI server can commit code as neatly described robot,
+ *         uses value from {@link org.shipkit.gradle.configuration.ShipkitConfiguration.Git#getEmail()}
+ *     </li>
+ * </ul>
+ */
+public class GitConfigPlugin implements Plugin<Project> {
```
wwilk: It is not related to the change but it would be nice to add unit tests for this plugin, to check if it can be applied and if it creates correct tasks. Or at least TODO in code if you don't have time for it at the moment.

epeee: Yes, we should do so. I just added tests to check whether GitConfigPlugin and GitSetupPlugin can be applied.

PR #568, opened 2017-11-14
minor cleanup

epeee:
- get rid of unused imports
 - make use of IOUtil.close 
 - get rid of output variable

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
@@ -52,13 +52,28 @@ public class IOUtil {
      * @param closeable the target, may be null
      */
     public static void close(Closeable closeable) {
+        close(closeable, false);
+    }
+
+    private static void close(Closeable closeable, boolean quietly) {
         if (closeable != null) {
             try {
                 closeable.close();
             } catch (IOException e) {
-                throw new RuntimeException("Problems closing closeable", e);
+                if (!quietly) {
+                    throw new RuntimeException("Problems closing closeable", e);
+                }
             }
         }
+    }
+
+    /**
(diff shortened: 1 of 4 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
```diff
@@ -80,31 +80,26 @@ public static void createDirectory(File file) {
...
+            FileOutputStream fos = null;
+            try {
+                fos = new FileOutputStream(file);
+                byte[] buf = new byte[1024];
+                int n;
+                while ((n = input.read(buf)) != -1) {
+                    fos.write(buf, 0, n);
+                }
+            } finally {
+                close(fos);
```
mockitoguy: If close() fails with IOException, we are losing the original exception.

Perhaps we need 2 methods: "close()" and "closeQuietly()".

I know that this problem was present before... since you are changing the code now, let's improve it :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
```diff
@@ -80,31 +80,26 @@ public static void createDirectory(File file) {
...
-                if (input != null) {
-                    input.close();
-                }
-                if (output != null) {
-                    output.close();
-                }
-            } catch (IOException e) {
-                throw new RuntimeException(e);
-            }
+            close(input);
```
mockitoguy: This can also swallow the original exception.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/util/IOUtil.java
epeee: +1

PR #571, opened 2017-11-18
Make ReleaseNeeded task configurable to force release 

### LINKED ISSUE #354: Make ReleaseNeeded task configurable to force release
### Problem

In Mockito, we have special handling of "[ci maven-central-release]" keyword in the commit message. This keyword instructs Shipkit to push to Central instead of just Bintray. For more information why we do it, see [the section of README.md](https://github.com/mockito/mockito/blob/release/2.x/README.md#how-to-release-new-version) of Mockito project. Problem is that just using "[ci maven-central-release]" keyword is sometimes insufficient to make the release because "assertReleaseNeeded" task may decide that the release is not necessary. To solve this use case, we need to do a whitespace change in the Javadoc as described in #353.

### Suggested solution

I'd like the "assertReleaseNeeded" task to expose a property that I can use in Mockito project to force the release. Basically in Mockito's shipkit.gradle, [in this line](https://github.com/mockito/mockito/blob/release/2.x/gradle/shipkit.gradle#L32), I would like to configure this:

```Gradle
boolean centralRelease = shouldReleaseToCentral(project)
afterEvaluate {
  //using afterEvaluate because 'assertReleaseNeeded' task is added after 'shipkit.gradle' file is evaluated by Gradle
  assertReleaseNeeded.forceRelease = centralRelease
}
```

This way, when "[ci maven-central-release]" keyword is present, the release will go out. This is the most intuitive behavior of release automation.

### References

Ticket #353 is related but not the same. #353 adds new keyword that can be used in commit message to force the release. This ticket adds a property on the task for advanced users that want fine grain control over "release needed" logic.

Let us know if you want to work on this or if you have any questions!

epeee:
fixes #354

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
@@ -87,6 +87,13 @@ public class ReleaseNeededTask extends DefaultTask {
 
     /**
      * If the comparison of publications should be skipped.
+     * <br>
+     * This can be useful if you would like to trigger a release (based on the given condition) even if
+     * publications are identical.
+     * <br>
+     * E.g. mockito project uses this to make sure that releases to mavenCentral are actually triggered:
+     * <br>
+     * <code>assertReleaseNeeded.skipComparePublications = shouldReleaseToCentral(project)</code>
      */
     public boolean isSkipComparePublications() {
         return skipComparePublications;
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
@@ -25,6 +25,7 @@ public class ReleaseNeededTask extends DefaultTask {
     private String releasableBranchRegex;
     private String commitMessage;
     private boolean pullRequest;
+    private boolean skipComparePublications;
     private boolean explosive;
     private List<File> comparisonResults = new LinkedList<File>();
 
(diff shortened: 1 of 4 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleaseNeededPlugin.java
```diff
@@ -19,6 +19,7 @@
  *      <li>SKIP_RELEASE env variable: no release if this environment variable is available</li>
  *      <li>[ci skip-release] commit message: no release if this commit message is available</li>
  *      <li>PR build: no release if the build job is a pull request build</li>
+ *      <li>skipComparePublications task property (ReleaseNeededTask): release if this property is true (although the previous and the current release might be identical)</li>
```
wwilk: It's not entirely true, it doesn't always release (it may still skip release for different reasons).

mockitoguy: Can you add Javadoc link to ReleaseNeededTask.setSkipComparePublications() method?

This way, it's easy to navigate the Javadoc!

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
```diff
@@ -84,6 +85,20 @@ public void setPullRequest(boolean pullRequest) {
         this.pullRequest = pullRequest;
     }
 
+    /**
+     * If the comparison of publications should be skipped.
```
mockitoguy: Can you write when this property is useful? (the use case)

It really helps understand the feature, if the user can read and understand the reason why he might use it!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleaseNeededPlugin.java
mockitoguy: I just read @wwilk, feedback to this commit message. Indeed we might want to reword it: "if the property is true, then comparing publications is skipped and the release can be triggered even if publications are identical. It is useful for..."

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/release/ReleaseNeededTask.java
epeee: yes, indeed. I will add a proper use case.

Issue #375, opened 2017-08-31
Better exception message when user forgets to configure GH_WRITE_TOKEN


mockitoguy:
### Problem

When the user forgets to configure GH_WRITE_TOKEN in Travis CI, the failure on Travis is not very clear:

```
[./gradlew] :gitPush
[./gradlew]   'git push' uses GitHub write token
[./gradlew]   Executing:
[./gradlew]     git push https://github.com/mockito/shipkit-bootstrap.git v0.0.3 master
[./gradlew] :gitPush FAILED
[./gradlew] 
[./gradlew] FAILURE: Build failed with an exception.
[./gradlew] 
[./gradlew] * What went wrong:
[./gradlew] Execution failed for task ':gitPush'.
[./gradlew] 
[./gradlew] > Execution of command failed (exit code 128):
[./gradlew]     git push https://github.com/mockito/shipkit-bootstrap.git v0.0.3 master
[./gradlew]     Captured command output:
[./gradlew]   remote: Invalid username or password.
[./gradlew]   fatal: Authentication failed for 'https://github.com/mockito/shipkit-bootstrap.git/'
```

I bumped into this issue when testing bootstrapping scenarios ([build failure](https://travis-ci.org/mockito/shipkit-bootstrap/builds/268985752)).

### Suggested solution

Instead of letting 'gitPush' fail with generic error, we should write a hint that the problem might be caused by missing or invalid setting of GH_WRITE_TOKEN. To get started, see the implementation of [git push task](https://github.com/mockito/shipkit/blob/master/src/main/groovy/org/shipkit/internal/gradle/git/tasks/GitPush.java#L32). Instead of just letting it fail, we should test return value and emit a warning or different exception.

If you want to take on this ticket, write a comment so that the team knows :) Let us know if you have any questions!

ArturSkowronski:
Hey @mockitoguy , I'd like to take this task :) I hope it won't take me so much time as previous one.

After a talk with @wwilk we found following things to do in this task: 
- Update docs about authorization with GH_WRITE_TOKEN and possible fallback to default git configuration
- Catch exception and check if it's message follow patter containing "Authentication failed"
- In this case, check if GH_WRITE_TOKEN is set and if not, provide hint that this can be possible issue. Otherwise, suggest it can be invalid.

mockitoguy:
Thank you @ArturSkowronski for contributing!

ArturSkowronski:
[...]

PR #576, opened 2017-11-26
[#375] Better exception message when user forgets to configure GH_WRITE_TOKEN

ArturSkowronski:
Throwing specific exception messages as requested in #375.


subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/CannotPushToGithubExceptionTest.groovy
```diff
@@ -0,0 +1,47 @@
...
+class CannotPushToGithubExceptionTest extends Specification {
+
+    def "should throw proper message for lack of GH_WRITE_TOKEN"() {
+        given:
+        GitPushTask gitPushTask = Mock(GitPushTask)
+        Exception originalException = new RuntimeException("Exception message")
+        when:
+        def result = CannotPushToGithubException.create(originalException, gitPushTask)
+        then:
+        result.message == CannotPushToGithubException.GH_WRITE_TOKEN_NOT_SET_MSG
```
mockitoguy: Let's test if we are correctly propagating exception 'cause'. This way we have full coverage for the behavior. Applies to other tests, too.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/tasks/GitPush.java
```diff
@@ -28,9 +31,16 @@
...
     public void gitPush(GitPushTask task) {
         TokenAvailabilityMessage.logMessage("git push", task.getSecretValue());
-
-        new DefaultProcessRunner(task.getProject().getProjectDir())
+        try {
+            new DefaultProcessRunner(task.getProject().getProjectDir())
                 .setSecretValue(task.getSecretValue())
                 .run(GitPush.gitPushArgs(task.getUrl(), task.getTargets(), task.isDryRun()));
+        } catch (GradleException e) {
+            if (CannotPushToGithubException.matchException(e)) {
```
mockitoguy: Let's add some tests around that so that we have good coverage for all "interesting" complexity, such as the conditionals. There are a couple of ways, the simplest could be adding a method that handles the exception for a Runnable:

```java
friendlyException(
            new Runnable() {
                public void run() {
                    new DefaultProcessRunner(task.getProject().getProjectDir())
                        .setSecretValue(task.getSecretValue())
                        .run(GitPush.gitPushArgs(task.getUrl(), task.getTargets(), task.isDryRun()));
                }
            });
```

New "friendlyException" method (or however we want to call it :) is easy to unit test.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/CannotPushToGithubExceptionTest.groovy
```diff
@@ -0,0 +1,47 @@
+package org.shipkit.internal.gradle.util
+
+import org.gradle.api.GradleException
+import org.shipkit.gradle.git.GitPushTask
+import spock.lang.Specification
+import spock.lang.Unroll
+
+class CannotPushToGithubExceptionTest extends Specification {
```
mockitoguy: Nice & clean unit tests. Thank you!

ArturSkowronski:
@mockitoguy Sure, I will address them today :)

ArturSkowronski:
@mockitoguy I decided to change way in which exceptions are handled. It is now more "enterprise-ish" but for IMHO more clear and more testable. Any feedback welcomed!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/handler/ProcessExceptionHandler.java
```diff
@@ -0,0 +1,27 @@
+package org.shipkit.internal.gradle.util.handler;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.gradle.api.GradleException;
+
+public class ProcessExceptionHandler {
```
mockitoguy: Can we make this simpler?

```java
public class Exceptions {
    public static void handling(Runnable code, Action<Throwable> handler) {
       //...
    }
}
```

Summary:
 - just one class with one static method
 - only one exception handler is needed
 - the handler reuses Action type that comes from Gradle API
 - the handler has just one method, no need for 2 separate "match" and "create"

This way, the code get simpler and the test code should get way simpler! Code review is the last chance to catch complexity before it goes to production!!!

Besides that, I really like the direction of those changes!

mockitoguy:
>@mockitoguy I decided to change way in which exceptions are handled. It is now more "enterprise-ish" but for IMHO more clear and more testable. Any feedback welcomed!

@ArturSkowronski - I like it! I think we can make it simpler and hence less code, much less test code. Then we merge ;)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/handler/ProcessExceptionHandler.java
ArturSkowronski: @mockitoguy Sure, fighting with complexity is a noble cause :) I implemented feedback, I hope you will like it!

ArturSkowronski:
@mockitoguy I have two more question 
- should I update some kind of documentation? 
- e2e tests - should I create some scenario covering lack of access to repo and proper messages? This seems a bit heavy IMHO but I can try to come up with some solution

wwilk:
About e2e - it might seem like an overkill to create an e2e for just this particular error, but on the other hand this is quite important, because not setting GH_WRITE_TOKEN is one of the first mistakes that user can make when using Shipkit. A pleasant experience should be guaranteed :)

We have a few of tests that use Gradle TestKit, they all end with IntegTest.groovy,, InitPluginIntegTest. 

You should be able to mock lack of GH_WRITE_TOKEN by setting ShipkitConfiguration.setWriteAuthToken to empty string, although I haven't done much with this for a while so I might be wrong. Let me know if you have any problems with it.

mockitoguy:
I merged this PR. @ArturSkowronski, I made a couple of commits on top of your PR, mainly to drive simplicity to the test code. Check it out :)

mockitoguy:
THANK YOU for working on this!

PR #578, opened 2017-11-28
update release needed doc

epeee:
The javadoc in release-needed-plugin.md did not contain the latest changes.
Furthermore, I thought it might be a good idea to get rid of duplication in javadoc and create a link instead of duplicating the doc to multiple java files.


mockitoguy:
Really nice change! Absolutely merging!!! Thank you also for reducing duplication.

Issue #460, opened 2017-10-01
Reduce number of open PR requests created automatically


szpak:
# Description

Currently Shipkit creates a separate PR in a downstream project for every released version. With high pace of development in a situation they are not merged/closed regularly there could be a number of open PRs which reduces visibility and they have to be closed manually. 

# Proposed solution

 - Simple version. Close old PR when a new one is created.
 - More sophisticated version. For not closed/merged PRs, update previously created branch (probably with `--force-with-lease`, but simple change could be acceptable as well), update a PR name and (optionally) a description (a new comment would be enough).

szczepiq:
Neat idea! Do you want to take a stab at it?

Updating PR has slightly better UX but it may be more problematic to implement/add complexity. +1 to closing PR.

szpak:
I consider that. It's quite separated topic, so maybe I will have a chance to practice Kotlin a little bit :).

wwilk:
Closing the previous PR would be quite simple - basically two GH API calls, one for all pull requests, and then another to close some of them (usually one). It would be also good to clean up the mess (delete the branch). 

The second solution - first check if there are open PRs. If not then we don't need to do anything. Otherwise we find the latest PR (GH API call), checkout the proper branch (easy - naming convention), do a version update as we do now, and then update PR instead of creating a new one.

As @szczepiq said I think the second solution would have better UX (less CRs to get around) but since these PRs may live longer, the risk of conflict with other changes is bigger. Of course this kind of conflicts can be resolved quickly and directly on GitHub but it puts this job on user.

So I think the first one is good enough.

@szpak About Kotlin - it seems that we don't have any decision yet if we are going in this direction https://github.com/mockito/shipkit/issues/453
What is your opinion about it?

nallar:
> As @szczepiq said I think the second solution would have better UX (less CRs to get around) but since these PRs may live longer, the risk of conflict with other changes is bigger. Of course this kind of conflicts can be resolved quickly and directly on GitHub but it puts this job on user.

It should be possible to do the second solution, but reset to the master branch before doing the update and force push with lease to the PR branch. This would avoid building up repeated update commits and avoid merge conflicts.

szpak:
@wwilk If the first solution can be delivery much faster it would be probably good to have it up and running. Then, that issues could be closed and the another enhancement could be created with "please contribute!" label waiting for someone interested in that feature having enough time to play with it.

szczepiq:
[...]

PR #582, opened 2017-12-05
Reducing number of version upgrade pull requests 

wwilk:
… by updating the open pull request instead of creating a new one #460 

I tested it by running "performVersionUpgrade" on shipkit-example:
- first with one open pull request -> pull request updated with new commit changing the version
- no open pull request -> new pull request created

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
@@ -1,0 +1,40 @@
+package org.shipkit.internal.gradle.versionupgrade;
+
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.json.simple.DeserializationException;
+import org.json.simple.JsonArray;
+import org.json.simple.JsonObject;
+import org.json.simple.Jsoner;
+import org.shipkit.internal.util.GitHubApi;
+
+import java.io.IOException;
+
+class FindOpenPullRequest {
+
+    private static final Logger LOG = Logging.getLogger(FindOpenPullRequest.class);
+
+    public String findOpenPullRequest(FindOpenPullRequestTask task) throws IOException, DeserializationException {
+        return findOpenPullRequest(task, new GitHubApi(task.getGitHubApiUrl(), task.getAuthToken()));
+    }
+
+    public String findOpenPullRequest(FindOpenPullRequestTask task, GitHubApi gitHubApi) throws IOException, DeserializationException {
+        String response = gitHubApi.get("/repos/" + task.getUpstreamRepositoryName() + "/pulls?state=open");
+
+        JsonArray pullRequests = Jsoner.deserialize(response, new JsonArray());
+
+        for (Object pullRequest : pullRequests) {
+            JsonObject head = (JsonObject) ((JsonObject) pullRequest).get("head");
+            String branchName = head.getString("ref");
+            if (branchName.matches(task.getVersionBranchRegex())) {
+                LOG.lifecycle("  Found an open pull request with version upgrade on branch {}", branchName);
+                return head.getString("ref");
+            }
+        }
+
+        LOG.lifecycle("  No open pull request with version upgrade found.");
+
+        return null;
+    }
+
+}
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
@@ -15,24 +15,25 @@ class FindOpenPullRequest {
     private static final Logger LOG = Logging.getLogger(FindOpenPullRequest.class);
 
     public String findOpenPullRequest(FindOpenPullRequestTask task) throws IOException, DeserializationException {
-        return findOpenPullRequest(task, new GitHubApi(task.getGitHubApiUrl(), task.getAuthToken()));
+        return findOpenPullRequest(task.getUpstreamRepositoryName(), task.getVersionBranchRegex(),
+            new GitHubApi(task.getGitHubApiUrl(), task.getAuthToken()));
     }
 
-    public String findOpenPullRequest(FindOpenPullRequestTask task, GitHubApi gitHubApi) throws IOException, DeserializationException {
-        String response = gitHubApi.get("/repos/" + task.getUpstreamRepositoryName() + "/pulls?state=open");
+    public String findOpenPullRequest(String upstreamRepositoryName, String versionBranchRegex, GitHubApi gitHubApi) throws IOException, DeserializationException {
+        String response = gitHubApi.get("/repos/" + upstreamRepositoryName + "/pulls?state=open");
 
         JsonArray pullRequests = Jsoner.deserialize(response, new JsonArray());
 
         for (Object pullRequest : pullRequests) {
             JsonObject head = (JsonObject) ((JsonObject) pullRequest).get("head");
             String branchName = head.getString("ref");
-            if (branchName.matches(task.getVersionBranchRegex())) {
+            if (branchName.matches(versionBranchRegex)) {
                 LOG.lifecycle("  Found an open pull request with version upgrade on branch {}", branchName);
                 return head.getString("ref");
             }
         }
 
-        LOG.lifecycle("  No open pull request with version upgrade found.");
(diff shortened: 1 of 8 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * See {@link ShipkitConfiguration.GitHub#getRepository()}
+     */
+    public void setUpstreamRepositoryName(String upstreamRepositoryName) {
+        this.upstreamRepositoryName = upstreamRepositoryName;
+    }
+
+    /**
+     * Regex matching version upgrade branch for any version.
```
mockitoguy: The description describes well what it is. Can you also information why why need it and an example value?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -0,0 +1,108 @@
...
+import org.gradle.api.tasks.TaskAction;
+import org.json.simple.DeserializationException;
+import org.shipkit.gradle.configuration.ShipkitConfiguration;
+
+import java.io.IOException;
+
+/**
+ * Looks for an open pull request with a version upgrade by:
+ * - querying GitHubAPI of {@link #upstreamRepositoryName} for all open pull requests
+ * - checks if any HEAD branch of resulting pull requests matches {@link #versionBranchRegex}
```
mockitoguy: Can you add an example branch name/regex that is used in standard Shipkit workflow.

This is potentially very useful Gradle task. I'm looking forward to elevating it to a public API!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -0,0 +1,108 @@
...
+
+    /**
+     * Returns branch of the current open pull request with version upgrade or null if it doesn't exist.
+     */
+    public String getOpenPullRequestBranch() {
+        return openPullRequestBranch;
+    }
+
+    /**
+     * Call if you want {@param #openPullRequestBranchCallback} to be executed after this task is finished.
```
mockitoguy: Can you also write why we would want to use it? It is very useful to understand the use case (e.g. the 'why') as much as what it is.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -0,0 +1,108 @@
...
+     */
```
mockitoguy: Nitpick: concise names are easier to read. Example:

```java
public void provideOpenPullRequestBranch(Task t, final Action<String> openPullRequestBranchCallback) {

public void provideBranchTo(Task t, final Action<String> branchAction) {
```

I'm wondering if the task should not be called "FindPullRequestBranchTask" because what we really care about is the branch.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -231,6 +271,13 @@ public void execute(Task task) {
         });
     }
 
+    static String getCurrentVersionBranchName(String dependencyName, String version, String openPullRequestBranch) {
+        if (openPullRequestBranch != null) {
+            return openPullRequestBranch;
+        }
+        return getVersionBranchName(dependencyName, version);
```
mockitoguy: There is a caveat: it is very useful to have the version in the pull request title. However, when we update the PR, the title is no longer accurate. I don't want to block this PR but let's think how we can address it.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
```diff
@@ -0,0 +1,40 @@
...
+        for (Object pullRequest : pullRequests) {
+            JsonObject head = (JsonObject) ((JsonObject) pullRequest).get("head");
+            String branchName = head.getString("ref");
+            if (branchName.matches(task.getVersionBranchRegex())) {
+                LOG.lifecycle("  Found an open pull request with version upgrade on branch {}", branchName);
+                return head.getString("ref");
+            }
+        }
+
+        LOG.lifecycle("  No open pull request with version upgrade found.");
```
mockitoguy: This message may be misleading to the user. Is it a problem that the pull request was not found? If the build fails and the user investigates the log he might think that it has something to do with that. I suggest to write something like (please tweak it :):

```
New pull request will be opened because we didn't find an existing PR to reuse
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -212,15 +241,26 @@ public void execute(String originRepoName) {
...
+                            upgradeDependencyExtension.getNewVersion(), openPullRequestBranch));
+                    }
+                });
+
+                task.onlyIf(wasOpenPullRequestNotFound(findOpenPullRequestTask));
                 task.onlyIf(wasBuildFileUpdatedSpec(replaceVersionTask));
             }
         });
 
+        //TODO: WW add validation for the case when 'dependency' property is not provided
```
mockitoguy: Good idea!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubApi.java
```diff
@@ -50,6 +50,24 @@ public String post(String relativeUrl, String body) throws IOException {
         }
     }
 
+    public String get(String relativeUrl) throws IOException {
```
mockitoguy: At some point let's figure out how to reliably test methods like those.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTest.groovy
```diff
@@ -0,0 +1,50 @@
...
+import spock.lang.Specification
+
+class FindOpenPullRequestTest extends Specification {
+
+    FindOpenPullRequest findOpenPullRequest = new FindOpenPullRequest()
+
+    def "should return null if response is empty"() {
+        given:
+        def gitHubApi = Mock(GitHubApi)
+        def findOpenPullRequestTask = Mock(FindOpenPullRequestTask)
```
mockitoguy: Hint: to simplify tests, you can pass repo and regex to the overloaded "findOpenPullRequest" method, instead of passing the task instance.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTest.groovy
```diff
@@ -0,0 +1,50 @@
+package org.shipkit.internal.gradle.versionupgrade
+
+import org.shipkit.internal.util.GitHubApi
+import spock.lang.Specification
+
+class FindOpenPullRequestTest extends Specification {
```
mockitoguy: Fantastic test coverage. Thank you!!!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
wwilk: I will change the naming of the method, but I would like to keep the name of the task. I can imagine a situation where we need to also use pull request title because we want to update it (the caveat you are mentioning in this review)

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTest.groovy
wwilk: Good idea, thanks!

wwilk:
Suggestions applied and merging.

wwilk:
I will address the caveat you mentioned in a issue aggregating all things that can be improved about version upgrade.

Issue #584, opened 2017-12-09
'published to' link in release notes is broken


epeee:
e.g. see https://github.com/mockito/mockito/blob/release/2.x/doc/release-notes/official.md
the generated 'published to' link is https://bintray.com/mockito/maven/mockito2.13.0 -> it should be https://bintray.com/mockito/maven/mockito/2.13.0
The link is also broken in shipkit release notes: https://plugins.gradle.org/plugin/org.shipkit.java0.9.148 -> should be https://plugins.gradle.org/plugin/org.shipkit.java/0.9.148

epeee:
it looks like this one is broken since #522

epeee:
fixed, see the linked pull requests.

PR #585, opened 2017-12-09
fix 'published to' links in shipkit release notes

epeee:
addresses #584 for shipkit project by applying the config described in javadoc:

https://github.com/mockito/shipkit/blob/ac02a3ad51e3d9defcfd480090ec24c3b61161c4/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java#L256-L268

### DIFF
--- a/gradle/shipkit.gradle
+++ b/gradle/shipkit.gradle
@@ -5,7 +5,7 @@ shipkit {
     team.developers = ['mockitoguy:Szczepan Faber', 'mstachniuk:Marcin Stachniuk',
                        'wwilk:Wojtek Wilk', 'epeee:Erhard Pointl', 'NagRock:Maciej Kuster']
 
-    releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
+    releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/"
 }
 
 apply plugin: 'org.shipkit.upgrade-dependency'


epeee:
did some testing locally:

----------

#### 0.9.149
 - 2017-12-09 - [3 commits](https://github.com/mockito/shipkit/compare/v0.9.148...v0.9.149) by epeee (1), [shipkit-org](https://github.com/shipkit-org) (1), [Szczepan Faber](http://github.com/mockitoguy) (1) - published to [Bintray](https://plugins.gradle.org/plugin/org.shipkit.java/0.9.149)
 - Version of shipkit upgraded to 0.9.148 [(#581)](https://github.com/mockito/shipkit/pull/581)

----------

0.9.149 is not available yet, but the link generated (https://plugins.gradle.org/plugin/org.shipkit.java/0.9.149) looks good.

gradle/shipkit.gradle
```diff
@@ -5,7 +5,7 @@ shipkit {
     team.developers = ['mockitoguy:Szczepan Faber', 'mstachniuk:Marcin Stachniuk',
                        'wwilk:Wojtek Wilk', 'epeee:Erhard Pointl', 'NagRock:Maciej Kuster']
 
-    releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java"
+    releaseNotes.publicationRepository = "https://plugins.gradle.org/plugin/org.shipkit.java/"
```
wwilk: What if we make the getter add a slash at the end if it doesn't end with slash yet? This way user doesn't have to care about it. Maybe we can encapsulate this logic in a utility method and use it for all getters that return URLs and are part of the public API?
@mockitoguy What do you think?

epeee: @wwilk This would break the powermock use case, isn't it?
Currently, we are using
`releaseNotes.publicationRepository = "https://bintray.com/powermock/maven/powermock/powermock-"`
in powermock in order to get links like https://bintray.com/powermock/maven/powermock/powermock-1.7.3

wwilk: True, good catch! But anyway, we should stick to one convention for all default values that we provide. So maybe all of them should have slash at the end?

Issue #590, opened 2017-12-10
Upgrade downstream improvements


wwilk:
Here are a few problems with the current design of version upgrade:

1. When we update open pull request, we don't update the title of it. Two solutions I can think of:
- update the title too, it would require another task and GitHub API call
- use title without version, eg "Shipkit upgraded". Simpler but it would make it harder to investigate history of closed pull requests, I guess. 

2. Merge pull request automatically when all checks passed. 

3. Add validation to UpgradeDependencyPlugin for "dependency" property. All tasks that depend on this property should fail fast and with a meaningful message if it's not provided.

PR #599, opened 2017-12-20
[Shipkit-590] Automatic Merge if all checks passed

ArturSkowronski:
Task for automatic merging of Pull Request after all checks pass successfully -
 second task from https://github.com/mockito/shipkit/issues/590

I'd like to receive your opinion @mockitoguy @wwilk  - what would be the best way to test it? Are there any nice frameworks that would allow creating automatic tasks for that? Doing it manually need a lot of ceremony with setting up checks for repo.

Like we discussed with @wwilk, I assume same GitHub write token will allow you to merge - I don't differentiate possibility that somebody can have write access but not merge access


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+                    gitHubApi.post("/repos/" + task.getUpstreamRepositoryName() + "/merges", body);
+                } else {
+                    int waitTime = 10000 * timeouts;
+                    Thread.sleep(waitTime);
+                    timeouts++;
+                    LOG.lifecycle("Pull Request checks still in pending state. Waiting %d seconds...", waitTime / 1000);
+                }
+            }
+        } catch (Exception e) {
+            throw new GradleException(e.getMessage());
```
wwilk: It would be good to add a message explaining that this exception happened during trying to merge the version upgrade pull request.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+
+        int timeouts = 0;
+
+        try {
+            while (isPending) {
+                if (timeouts > 20) {
+                    throw new RuntimeException("To many retries. Merge aborted");
+                }
+
+                String statusesResponse = gitHubApi.get("/repos/" + task.getUpstreamRepositoryName() + "/statuses/" + sha);
```
wwilk: Let's extract getting status information into its own class. It will make this class cleaner and remove one of its responsibilities.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequestTask.java
```diff
@@ -0,0 +1,126 @@
...
+    /**
+     * See {@link org.shipkit.gradle.configuration.ShipkitConfiguration.GitHub#getWriteAuthToken()}
+     */
+    public void setAuthToken(String authToken) {
+        this.authToken = authToken;
+    }
+
+
+    public UpgradeDependencyExtension getVersionUpgrade() {
+        return versionUpgrade;
```
wwilk: Since we only use baseBranch property, we should have it as a property of the task instead of UpgradeDependencyExtension. It makes API of this task easier to use. If you only want to change baseBranch you don't need to modify/create the whole versionUpgrade object. I see that I used it in CreatePullRequest and ReplaceVersion tasks, I will change it there too.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+        JsonObject commit = (JsonObject) branch.get("commit");
+        return commit.getString("sha");
+    }
+
+    private boolean stateResolver(String state, String description) {
+        if (state.equals("success")) {
+            return false;
+        }
+        if (state.equals("error")) {
+            throw new RuntimeException("Status of check '" + description + "':error. Merge aborted");
```
wwilk: This message could be sth like this:
"Pull request #pullRequestNumber cannot be merged. #description. You can see the details here : #targetUrl". This way we can merge error and failure ifs into one. I don't know what's the difference between these statuses, I didn't find an explanation in API docs. We should also use GradleException.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+                    throw new RuntimeException("To many retries. Merge aborted");
+                }
+
+                String statusesResponse = gitHubApi.get("/repos/" + task.getUpstreamRepositoryName() + "/statuses/" + sha);
+                JsonArray statuses = Jsoner.deserialize(statusesResponse, new JsonArray());
+
+                for (Object status : statuses) {
+                    String state = ((JsonObject) status).getString("state");
+                    String description = ((JsonObject) status).getString("description");
+                    isPending = stateResolver(state, description);
```
wwilk: What if the first check is pending while the second one passed? We should wait for all the checks to pass before merging, so it should be a logical conjunction instead of just an assignment.

wwilk: It seems that if we use a different endpoint for combined status (/status instead of /statuses), we don't need to iterate over a collection. Also it seems that even if all checks pass we get some additional pending ones for some reason. See these ones (one of commits from shipkit):
https://api.github.com/repos/mockito/shipkit/commits/68f9450/statuses
https://api.github.com/repos/mockito/shipkit/commits/68f9450/status

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+            "  \"head\": \"" + headBranch + "\"," +
+            "  \"base\": \"" + task.getVersionUpgrade().getBaseBranch() + "\"" +
+            "}";
+
+        int timeouts = 0;
+
+        try {
+            while (isPending) {
+                if (timeouts > 20) {
+                    throw new RuntimeException("To many retries. Merge aborted");
```
wwilk: Nitpicky: "too" instead of "to". Also probably GradleException would be better and mentioning a pull request number here.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,94 @@
...
+            LOG.lifecycle(" Skipping pull request merging due to dryRun = true");
+            return;
+        }
+
+        String headBranch = BranchUtils.getHeadBranch(task.getForkRepositoryName(), task.getVersionBranch());
+
+        IncubatingWarning.warn("merge pull requests");
+        LOG.lifecycle("Waiting for status of a pull request in repository '{}' between base = '{}' and head = '{}'.", task.getUpstreamRepositoryName(), task.getVersionUpgrade().getBaseBranch(), headBranch);
+
+        String sha = retrievePullRequestSha(task, gitHubApi, headBranch);
```
wwilk: Can we get this SHA (along with pull request number) from one of the tasks accessing the pull request (CreatePullRequest and FindOpenPullRequest)? You can use FindOpenPullRequest#provideBranchTo method and sth similar in CreatePullRequest. This way we have one API call less.

We would have to check if it works with both creating a new pull request and with updating the existing one. I'm not sure what happens in Gradle if you depend on a task that is skipped (CreatePullRequest is skipped if we already have an open pull request).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
ArturSkowronski: Thanks for catching typo! This exception will be rethrowed to Gradle exception. I will mention number of pull request.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
ArturSkowronski: @wwilk To first comment - we will. If any of check will fail, exception will be thrown - loop will stop, GradleException will be rethrown. In case of any Status check in pending state, method will return false so while loop will go once again.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
ArturSkowronski: It's something worth checking with this potential skipped depdendency, I will do that. I will also try to retrieve sha from previous calls.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
ArturSkowronski: All Runtime exceptions will be rethrowed as Gradle Exception.  I will update the message.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
wwilk: Yeah but in case you have a statuses like that:
{ status: 'Pending'}
{ status: 'Success'}

You start with isPending = true.
After the first status you set it to true.
After the second one to false, because it's a success.
So in the end isPending is false, although there is still one check pending.
Or am I missing sth?

But anyway if we just check combined status we don't need to worry about that :)

ArturSkowronski: Yeah, your point is true and I don't want why I fixed myself on this solution. It's a stupid miss from me :) I'll change it to the combined endpoint

ArturSkowronski:
@wwilk I moved Status Check logic to external class and used bulk endpoint for statuses. 

What's left:
* I need to reuse Pull Request data from previous tasks
* I need to investigate if statuses endpoint show result properly
* few tests need to be added to new logic

ETA: I hope tomorrow :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
ArturSkowronski: @wwilk it seems that skipping task don't skip dependant tasks: https://discuss.gradle.org/t/onlyif-affect-only-task-but-not-tasks-that-depends-on/14887 . But I'm looking for confirmation in official docs

ArturSkowronski:
@wwilk please check my changes. I see I need to introduce one more in MergePullRequestTask.java about baseBranch, but otherwise code seems fine from my perspective

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -254,6 +256,37 @@ public void execute(String openPullRequestBranch) {
...
+
+                gitOriginPlugin.provideOriginRepo(task, new Action<String>() {
+                    @Override
+                    public void execute(String originRepoName) {
+                        task.setForkRepositoryName(originRepoName);
+                        task.setUpstreamRepositoryName(conf.getGitHub().getRepository());
+                    }
+                });
+
+                findOpenPullRequestTask.provideOpenPullRequest(task, new Action<OpenPullRequest>() {
```
wwilk: What about the case when there is no open pull request? I think we have to have two different modes (doesn't have to be implemented now) - one when we update the PR and one when we automatically merge. It doesn't make sense to use both.

So we have to add providePullRequest method to createPullRequest task too. This way the pull request to merge will be provided either by findOpenPullRequest task or createPullRequest (when there is no open one).

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTest.groovy
```diff
@@ -28,9 +29,9 @@ class FindOpenPullRequestTest extends Specification {
     def "should return head->ref if it matches versionBranchRegex"() {
         given:
         def gitHubApi = Mock(GitHubApi)
-        gitHubApi.get("/repos/repo/pulls?state=open") >> "[{\"head\" : {\"ref\" : \"shipkit-1\"}} ]"
+        gitHubApi.get("/repos/repo/pulls?state=open") >> "[{\"url\": \"url-1\", \"head\" : {\"ref\" : \"shipkit-1\", \"sha\" : \"sha-1\"}} ]"
 
         expect:
-        "shipkit-1" == findOpenPullRequest.findOpenPullRequest("repo", "shipkit-[0-9]*", gitHubApi)
+        new OpenPullRequest(ref: "shipkit-1", sha: "sha-1", "url": "url-1") == findOpenPullRequest.findOpenPullRequest("repo", "shipkit-[0-9]*", gitHubApi)
```
wwilk: Thanks for updating the test!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/BranchUtils.groovy
```diff
@@ -0,0 +1,12 @@
+package org.shipkit.internal.gradle.util
+
+class BranchUtils {
+
+    public static String getHeadBranch(String forkRepositoryName, String headBranch) {
```
wwilk: Thanks for extracting it into an utility class!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/GitHubStatusCheckTest.groovy
```diff
@@ -0,0 +1,38 @@
...
+        task.getUpstreamRepositoryName() >> "upstreamRepo"
+        1 * gitHubApi.get("/repos/upstreamRepo/commits/sha/status") >> "{\"state\": \"pending\"}"
+        1 * gitHubApi.get("/repos/upstreamRepo/commits/sha/status") >> "{\"state\": \"success\"}"
+        when:
+        def result = gitHubStatusCheck.checkStatusWithTimeout()
+        then:
+        result == true
+    }
+
+    def "should return true if has error status"() {
```
wwilk: Test name lies, it actually throws an exception. Also there is no test for when checkStatusWithTimeout returns false.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/OpenPullRequest.java
```diff
@@ -0,0 +1,62 @@
...
+    public void setRef(String ref) {
+        this.ref = ref;
+    }
+
+    public String getUrl() {
+        return url;
+    }
+
+    @Override
+    public boolean equals(Object o) {
```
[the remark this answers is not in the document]
ArturSkowronski: I introduced it for test sake to easily compare objects. I will do assertion on properties and remove those

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -0,0 +1,83 @@
...
+    private MergePullRequestTask task;
+    private GitHubApi gitHubApi;
+
+    public GitHubStatusCheck(MergePullRequestTask task, GitHubApi gitHubApi) {
+        this.task = task;
+        this.gitHubApi = gitHubApi;
+    }
+
+    public boolean checkStatusWithTimeout() throws IOException, InterruptedException {
+        int timeouts = 0;
```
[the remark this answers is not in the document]
ArturSkowronski: Sure, probably it's better name

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -0,0 +1,83 @@
...
+        while (timeouts < 20) {
+            if (statusCheck(task, gitHubApi)) {
+                return true;
+            } else {
+                int waitTime = 10000 * timeouts;
```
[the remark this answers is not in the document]
ArturSkowronski: @wwilk I'm a bit relucant with that myself but for different case - this solution is not configurable. I didn't wanted to make it more complex than necessary, but I wonder if there won't be case when all checks will be very long. That's why I made waitTime growning in every iteration - still similar solution can be achieved by allowing more retries OR making it configurable (for now it is hardcoded on 20). Please let me know what you think.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/GitHubStatusCheckTest.groovy
ArturSkowronski: Due to retries, such a test would be very long and don't bring much value. I'll change it to set up amount of retries in constructor.

wwilk: Maybe it would be good to put Thread.sleep into a separate class, and mock it?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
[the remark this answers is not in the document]
wwilk: It will be certainly configurable in the long run, for now it can be a fixed value. We will test it on Shipkit and example project for a while before making it public. We can add a TODO in code. 

I think we can make a little more retries, since 20 * 10 s may be not enough for all shipkit checks to pass. It seems like it's 5-7 minutes at the time. We can set retries = 100 for now. What's the advantage of incremental delay between retries in this case?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -254,6 +256,37 @@ public void execute(String openPullRequestBranch) {
             }
         });
 
+        TaskMaker.task(project, MERGE_PULL_REQUEST, MergePullRequestTask.class, new Action<MergePullRequestTask>() {
+            @Override
+            public void execute(final MergePullRequestTask task) {
+                task.setDescription("Merge pull request when all checks will be passed");
+                task.mustRunAfter(CREATE_PULL_REQUEST);
```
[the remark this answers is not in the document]
ArturSkowronski: @wwilk I'm not sure. What if check will fail for some reason? I assume there is possibility to fix something without opening new pull request, just over old one. 

Maybe we should base it on testing if pull request is open and run this method only if PullRequest is really open? We should 

    task.onlyIf(isPullRequestOpen());
    task.onlyIf(wasBuildFileUpdatedSpec(replaceVersionTask));

where isPullRequestOpen contains logic about PullRequest accessibility?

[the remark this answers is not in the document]
wwilk: You mean when check fails and it's not auto-merged? I guess it should happen much less often than the check succeeding so in this case we can create another PR. But anyway maybe it would be better to base .onlyIf on the existence of pull request to be merged. So if either FindOpenPullRequest returns one, or CreatePullRequest creates one, we merge. Otherwise we don't.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
ArturSkowronski: @wwilk I pushed my proposition of solution. Please take a look

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/BranchUtils.groovy
```diff
@@ -0,0 +1,34 @@
+package org.shipkit.internal.gradle.util
+
+import org.gradle.api.logging.Logger
+import org.gradle.api.logging.Logging
+import org.json.simple.JsonObject
+import org.shipkit.internal.gradle.git.OpenPullRequest
+
+class BranchUtils {
```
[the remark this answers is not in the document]
ArturSkowronski: @mockitoguy Nope, I'll migrate it to Java. In truth, I'm myself a bit reluctant in using Groovy for production code (we had a lot of bumps with its interoperability with Java in one of the products) so I fully agree with you.

ArturSkowronski:
@mockitoguy I updated docs and addressed feedback.

@wwilk I did a bit of change in behavior - I decided not to merge if no status checks defined. I don't fail task but log information that PR cannot be merged without any checks. I think it possibly could backfire user. If you think differently, It's really easy to change to also merge those without any statuses. Still, I extended GitHubStatusCheck to pass that information to the MergePullRequest Task.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/OpenPullRequest.java
```diff
@@ -0,0 +1,31 @@
+package org.shipkit.internal.gradle.git;
+
+public class OpenPullRequest {
+    private String sha;
+    private String ref;
+    private String url;
+
+    public String getSha() {
```
mockitoguy: Can you add Javadoc on the class and all the public methods? (examples of values in the getters are super useful). Given my understanding of the code, this class will be public at some point when we promote pull request features from incubating state.

When you add Javadoc, this code will be even better :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/BranchUtils.java
```diff
@@ -0,0 +1,35 @@
...
+
+    public static String getHeadBranch(String forkRepositoryName, String headBranch) {
+        return getUserOfForkRepo(forkRepositoryName) + ":" + headBranch;
+    }
+
+    private static String getUserOfForkRepo(String forkRepositoryName) {
+        return forkRepositoryName.split("/")[0];
+    }
+
+    public static OpenPullRequest getOpenPullRequest(JsonObject pullRequest, String filter) {
```
mockitoguy: I think we can make it simpler. It's not intuitive that "BranchUtils" class holds logic that transforms JSON to PullRequest object. Also, null handling requires me to navigate various methods to understand how it works.

Let's drop this method. We can have a separate class/factory method that is responsible for translating JSON to PullRequest. The client can call "matches" method on ```pullRequest.getRef()``` if branch filtering is needed.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/CreatePullRequest.java
```diff
@@ -4,46 +4,43 @@
...
-
-    private String getHeadBranch(String forkRepositoryName, String headBranch) {
-        return getUserOfForkRepo(forkRepositoryName) + ":" + headBranch;
-    }
-
-    private String getUserOfForkRepo(String forkRepositoryName) {
-        return forkRepositoryName.split("/")[0];
+        String response = gitHubApi.post("/repos/" + task.getUpstreamRepositoryName() + "/pulls", body);
+        JsonObject pullRequest = Jsoner.deserialize(response, new JsonObject());
+        return BranchUtils.getOpenPullRequest(pullRequest, null);
```
mockitoguy: Suggestion: ```return toPullRequest(pullRequest)``` (based on my other comment)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequest.java
```diff
@@ -14,28 +16,25 @@
...
 
         JsonArray pullRequests = Jsoner.deserialize(response, new JsonArray());
 
         for (Object pullRequest : pullRequests) {
-            JsonObject head = (JsonObject) ((JsonObject) pullRequest).get("head");
-            String branchName = head.getString("ref");
-            if (branchName.matches(versionBranchRegex)) {
-                LOG.lifecycle("  Found an open pull request with version upgrade on branch {}", branchName);
-                return head.getString("ref");
+            OpenPullRequest openPullRequest = BranchUtils.getOpenPullRequest((JsonObject) pullRequest, versionBranchRegex);
```
mockitoguy: suggestion (based on my previous feedback):

```java
PullRequest pr = toPullRequest(pullRequest);
if (versionBranchRegex.matches(pr.getRef()) { ...
```

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -100,26 +100,32 @@ public void setVersionBranchRegex(String versionBranchRegex) {
     /**
      * Returns branch of the current open pull request with version upgrade or null if it doesn't exist.
      */
-    public String getOpenPullRequestBranch() {
-        return openPullRequestBranch;
+    //TODO Refactor this method to replace provideBranchTo method
```
mockitoguy: Let's wrap up this TODO :)

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/FindOpenPullRequestTask.java
```diff
@@ -100,26 +100,32 @@ public void setVersionBranchRegex(String versionBranchRegex) {
...
+     * the correct time when the task should call {@link #getOpenPullRequest()}, because this value is only available
      * after {@link FindOpenPullRequestTask} is executed.
      *
      * Using this method guarantees that:
      * - the value will be already available when the callback {@param #branchAction} is executed
      * - the task {@param #dependant} is executed after {@link FindOpenPullRequestTask}
      */
-    public void provideBranchTo(Task dependant, final Action<String> branchAction) {
+
+    public void provideOpenPullRequest(Task dependant, final Action<OpenPullRequest> branchAction) {
```
mockitoguy: This is nice!!! Suggestion: "branchAction" -> "action".

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,58 @@
...
+        String url = task.getPullRequestUrl();
+        String body = "{" +
+            "  \"head\": \"" + headBranch + "\"," +
+            "  \"base\": \"" + task.getBaseBranch() + "\"" +
+            "}";
+
+        try {
+            PullRequestStatusCheck checkStatus = gitHubStatusCheck.checkStatusWithRetries();
+
+            if (checkStatus == PullRequestStatusCheck.STATUS_TIMEOUT) {
```
mockitoguy: What about the "unsuccessful" status? E.g. when one of the checks fails?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/MergePullRequest.java
```diff
@@ -0,0 +1,58 @@
...
+import org.gradle.api.GradleException;
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
+import org.shipkit.internal.gradle.git.PullRequestStatusCheck;
+import org.shipkit.internal.gradle.util.BranchUtils;
+import org.shipkit.internal.util.GitHubApi;
+import org.shipkit.internal.util.GitHubStatusCheck;
+import org.shipkit.internal.util.IncubatingWarning;
+
+class MergePullRequest {
```
mockitoguy: I like this new feature a lot!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -209,46 +219,83 @@ public void execute(String originRepo) {
...
+                createPullRequestTask.provideCreatedPullRequest(task, new Action<OpenPullRequest>() {
+                    @Override
+                    public void execute(OpenPullRequest openPullRequest) {
+                        if (openPullRequest != null) {
+                            setPullRequestDataToTask(openPullRequest, task);
+                        }
+                    }
+                });
+
+                task.onlyIf(isPullRequestOpen(task));
```
mockitoguy: I suggest that we simplify this complexity. We can drop all 3 'onlyIf' clauses from here. Instead we add and 'else' section to line 294. Am I correct? The less complexity the better!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/BranchUtils.java
ArturSkowronski: Good Point!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
ArturSkowronski: @mockitoguy In the previous version, there was the usage of  PullRequest in two cases: when PR was freshly created and if it was already opened in past but the task was rerun. After discussion in some previous comments with @wwilk we decided, for now, to just try merging directly after creating a pull request. In the majority of cases a failure of the checks will force manual intervention of committers and as this is the last step in the process (at least for now), they can do merging by themselves. 

That's why this three onlyIfs were added (but, in truth, we can rid off ```task.onlyIf(isPullRequestOpen(task))``` one, as this situation should never happen in a standard case).

mockitoguy:
@ArturSkowronski, you applied the fixes and we are ready to merge, right? I'd love to see this feature running! @wwilk, do you have any comments/feedback?

ArturSkowronski:
Hey, @mockitoguy, yep, I think it's ready for merge (I addressed every feedback item other than this TODO stuff, as we discussed previously with @wwilk that we'll address it in future as this is place which is basically not touched by logic of PR, TODO was added to not forget about this item as it was spotted)

I was out for the weekend that's why this reply is so delayed :)

mockitoguy:
Let's ship this baby :)

ArturSkowronski:
Splendid!

PR #600, opened 2017-12-20
Ignore test of default behaviour if env variable exists

jhoglin:
How about this solution to the problem with #594 @wwilk ?

Its not perfect to ignore test but as far as my abilities go right now its to much work to refactor so I can mock the `EnvVariables` class, which is the best way to go I believe. 

About the existence of `GH_WRITE_TOKEN` is that present on all branches or just master or some such?

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -108,6 +108,8 @@ public class ShipkitConfiguration {
 
     public class GitHub {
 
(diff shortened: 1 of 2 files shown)


wwilk:
Thanks you for working on this! As you said yourself it's not a perfect solution and @epeee  suggested a more generic one, that can be used in more places, so we would probably go with it. 

If you want to contribute by fixing any other issue I would be happy to help you with it! :)

wwilk:
I'm not closing this one until https://github.com/mockito/shipkit/pull/601 is merged

jhoglin:
Cool. Perfect that @epeee fixad that refactoring. Now it will be as it should be.

wwilk:
Merged #601 so closing

PR #601, opened 2017-12-21
Add possibility to mock EnvVariables

epeee:
currently, tests are failing on travis due to "GH_WRITE_TOKEN" env variable.
let's mock `getNonEmptyEnv("GH_WRITE_TOKEN")` in order to make the test more robust.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -35,7 +35,7 @@ public class ShipkitConfiguration {
         this(new EnvVariables());
     }
 
-    public ShipkitConfiguration(EnvVariables envVariables) {
+    ShipkitConfiguration(EnvVariables envVariables) {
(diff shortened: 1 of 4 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -31,7 +32,11 @@
     private String previousReleaseVersion;
 
     public ShipkitConfiguration() {
-        this(new ShipkitConfigurationStore());
+        this(new EnvVariables());
+    }
+
+    public ShipkitConfiguration(EnvVariables envVariables) {
```
epeee: Using internal EnvVariable in public api is not good. I will rework this one later on today.

wwilk: What if you make this constructor package-private and create ShipkitConfigurationFactory in the same package, that would be located in test code, not production one? This way it can be used across all tests and it is not a part of public API.

wwilk:
Thanks for working on this, it's been an issue for a while!

PR #605, opened 2017-12-24
get rid of hardcoded github url

### LINKED ISSUE #595: Hardcoded url to github
There are hardcoded url to github.com. The setting `conf.getGitHub().getUrl()` should be used instead. 

https://github.com/mockito/shipkit/blob/514ac7a43af0e3513a83f7518221086b203d4b45/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotes.java#L53

https://github.com/mockito/shipkit/blob/514ac7a43af0e3513a83f7518221086b203d4b45/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotes.java#L61

did take a quick look and didn't find any test that runs this code section so I didn't fix it in the current PR #594

epeee:
... and use conf.getGitHub().getUrl() instead.

fixes #595

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotes.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotes.java
@@ -22,6 +22,7 @@ import java.io.File;
 import java.util.Collection;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Set;
 
 public class UpdateReleaseNotes {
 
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotes.java
```diff
@@ -42,25 +43,33 @@ void updateReleaseNotes(boolean previewMode, File releaseNotesFile, String newCo
...
+                                                    String githubUrl) {
         Map<String, Contributor> out = new HashMap<String, Contributor>();
-        for (String contributor : contributorsFromConfiguration) {
+        out.putAll(transform(contributorsFromConfiguration, githubUrl));
+        out.putAll(transform(contributorsFromGitHub.getAllContributors()));
+        out.putAll(transform(developers, githubUrl));
+        return out;
+    }
+
+    private static Map<String, Contributor> transform(Collection<String> contributors, String githubUrl) {
```
wwilk: Since the two transform methods in this class have different behaviour, let's name them accordingly. It would be easier to read this code without looking for the type of variables to see which method is used. Maybe sth like "transformToDefaultContributorsByName" and "groupByName"?

epeee: I already have plans to do some further refactoring in this area and most likely this part of the code will change again.
We could rename the methods but I think it does not add a lot of value.
@wwilk, are you ok with merging it as it is and I will come up with a larger refactoring during the next days?

wwilk: Yeah I guess removing hardcoded value is more important at the moment. Looking forward to this refactoring!

PR #610, opened 2017-12-26
Improved authentication for GH API requests

wwilk:
CI Build fails due to:

Execution failed for task ':findOpenPullRequest'.
[upgradeDownstream] [performVersionUpgrade] > java.io.IOException: GET https://api.github.com/repos/mockito/shipkit/pulls?state=open?access_token=[SECRET] failed, response code = 403, response body:
[upgradeDownstream] [performVersionUpgrade]   {"message":"API rate limit exceeded for 52.54.31.11. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)","documentation_url":"https://developer.github.com/v3/#rate-limiting"}

As you can notice we use 
pulls?state=open?access_token=[SECRET],
 so we concatenate strings in a wrong way, and because of that authentication is not picked up by GitHub. I used Authorization header instead.

Tested it by running ./gradlew findOpenPullRequest and checking if in response headers I get correct limits, see https://developer.github.com/v3/#rate-limiting for details.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubApi.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubApi.java
@@ -1,5 +1,7 @@
 package org.shipkit.internal.util;
 
+import org.gradle.api.logging.Logger;
+import org.gradle.api.logging.Logging;
 import org.shipkit.internal.notes.util.IOUtil;
 
 import javax.net.ssl.HttpsURLConnection;
@@ -13,6 +15,8 @@ import java.net.URL;
  */
 public class GitHubApi {
 
+    private static final Logger LOG = Logging.getLogger(GitHubApi.class);
+
     private final String gitHubApiUrl;
     private final String authToken;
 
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubApi.java
```diff
@@ -40,35 +45,31 @@ public String post(String relativeUrl, String body) throws IOException {
             }
         }
 
-        if (conn.getResponseCode() < HttpURLConnection.HTTP_BAD_REQUEST) {
```
epeee: thx for getting rid of this duplication!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/GitHubApiTest.groovy
```diff
@@ -26,7 +25,6 @@ class GitHubApiTest extends Specification {
 
         then:
         def ex = thrown(Exception)
-        ex.message.startsWith(
-            "GET https://api.github.com/repos/shipkit-example/pulls?access_token=[SECRET] failed")
+        !ex.message.contains("accessToken");
```
epeee: a minor one, the test name does no longer fit 100%, isn't it?

Issue #57, opened 2017-04-02
Ignore defined users in release notes and contributors list in pom.xml


mstachniuk:
Sometimes it's needed to ignore some users in release notes and in contributors list in pom.xml. It can be e.g. a technical account like continuous-delivery-drone in Mockito project. The scope of this task is to add a possibility to ignore defined users in release notes and contributors list in pom.xml.

szczepiq:
Seems useful!

I suggest to keep low priority on this on for now. We definitely should automatically exclude the generic use that the release tools use. I haven't got a use case yet to support other 'excluded' contributors, but yeah, maybe it is useful long term :)

micd:
Simply and effective solution is to add new properties file which will include the list of ignored users and checking them when generating release notes and pom contributors.

Second one is that we can add argument with ignored users when calling gradle plugin from command line. It may be painful for end user if there's a lot of them. On the other hand you can set which user or commit should be ignored for every run.  

We can also combine both ideas.

@mstachniuk , @mockitoguy any other ideas?

wwilk:
We already have one properties files, no need to create a new one. Having command line argument is not very convenient. It shouldn't be something that changes with every run. More like a blacklist. 

It would be best to add it to ShipkitConfiguration.

micd:
I've done some research. 

As @wwilk proposed it would be best to add new ShipkitConfiguration field for ignored contributors
I have two ideas for using it when generating pom contributors and release notes. 

We can extend `IgnoredCommit` predicate to ignore commits with ignored authors. For pom contributors we can extend GitHubContributorsProvider.mergeContributors to ignore contributors set in configuration. This solution is quite nice, because we have only one point of filtering and we do not care about those commits/contributors in further processing.

Another option is extend places where pom and release notes are generated to ignore commits and contributors set in configuration, such as: `DetailedFormatter` and `ContributorsSerializer`(?).

What do you think?

wwilk:
Awesome investigation!

Yeah it seems that we can ignore every commit that either contains an ignored part or was created by an ignored contributor. You are right, the sooner we remove ignored commits/contributors, the better.

Thanks for working on this. When you create a pull request please describe how you tested it.

PR #612, opened 2017-12-28
Ignoring contributors

micd:
Solution for issue #57
Added new configuration field and extended plugins to ignore contributors (or contributions done by contributors) defined in config.

Test I - release notes
1. install new shipkit version
2. call `./gradlew updateReleaseNotes `- release notes should contain all contributors
3. add `team.ignoredContributors = ['shipkit-org']` to **shipkit.gradle**
4. call `./gradlew updateReleaseNotes` - release notes should not contain _shipkit-org_ and its commits should not be included in the summary

Test II - contributors
1. install new shipkit version
2.  call `./gradlew fetchContributors` - `all-contributors.json` should contain all contributors
3. add `team.ignoredContributors = ['shipkit-org']` to **shipkit.gradle**
4.  call `./gradlew fetchContributors` - `all-contributors.json` should not cointain _shipkit-org_ user


subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -409,19 +423,20 @@ public void setContributors(Collection<String> contributors) {
...
-     * This is a good default because it helps us identify missing configuration early.
-     * However sometimes we want to check if the user has configured a property (like for GitHub write token) without failing.
-     *
-     * @return lenient copy of this configuration instance
-     */
-    public ShipkitConfiguration getLenient() {
-        return new ShipkitConfiguration(store.getLenient());
+        /**
+         * Contributors to be ignored in release notes and generated pom file.
+         * It should be VCS login
```
wwilk: It would be good to show an example of VCS login, let's make this comment compatible with the one for getContributors and also mention what happens if user is included in both contributors and ignoredContributors.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/contributors/IgnoredContributor.java
```diff
@@ -0,0 +1,35 @@
...
+    static IgnoredContributor of(Collection<String> ignoredContributors) {
+        return new IgnoredContributor(ignoredContributors);
+    }
+
+    private IgnoredContributor(Collection<String> ignoredContributors) {
+        this.ignoredContributors = ignoredContributors;
+    }
+
+    @Override
+    public boolean isTrue(Contributor contributor) {
```
wwilk: Let's make this class a little more generic by accepting a string here, and explaining what format exactly this contributor string should be. This way we can reuse this class in IgnoredCommit instead of duplicating the for loop.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/contributors/github/Contributors.java
```diff
@@ -15,9 +17,11 @@
      * @param apiUrl address of GitHub api endpoint, for example: "https://api.github.com"
      * @param repository name of GitHub repository, for example: "mockito/mockito"
      * @param readOnlyAuthToken the GitHub auth token
+     * @param ignoredContributors contributors to be ignored - VCS logins
```
wwilk: Please include an example of acceptable VCS login.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/generator/ReleaseNotesGenerators.java
```diff
@@ -15,11 +15,11 @@
...
     /**
-     * @param workDir the working directory where 'git' operations will be executed
-     * @param gitHubApiUrl GitHub API endpoint address, for example: https://api.github.com
-     * @param gitHubRepository GitHub gitHubRepository in format USER|COMPANY/REPO_NAME, for example: mockito/mockito
+     * @param workDir           the working directory where 'git' operations will be executed
+     * @param gitHubApiUrl      GitHub API endpoint address, for example: https://api.github.com
+     * @param gitHubRepository  GitHub gitHubRepository in format USER|COMPANY/REPO_NAME, for example: mockito/mockito
      * @param readOnlyAuthToken read only auth token used to communicate with GitHub
-     * @param ignoredCommit responsible decide if commits should not be included in release notes
+     * @param ignoredCommit     responsible decide if commits should not be included in release notes
```
wwilk: I know that you didn't add this content, but if you have a second you can change it to "responsible for deciding".

subprojects/shipkit/src/test/groovy/org/shipkit/internal/notes/contributors/IgnoredContributorTest.groovy
```diff
@@ -0,0 +1,30 @@
...
+import spock.lang.Specification
+
+class IgnoredContributorTest extends Specification {
+
+    def "should ignore contributor from ignored list"() {
+        def ignoredContributor = IgnoredContributor.of(["ignoredContributor"])
+        def contributor = new DefaultProjectContributor("name", "ignoredContributor", "profileUrl", 1)
+
+        expect:
+        ignoredContributor.isTrue(contributor) == true
```
wwilk: You can just use `ignoredContributor.isTrue` or `!ignoredContributor.isTrue` instead of comparing to boolean literals.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/contributors/IgnoredContributor.java
micd: It's better to reuse the code, but I'm not sure if `IgnoredContributor` should be parametrized with String. We would move some logic for ignoring contributor outside the class, because it's caller, who  would have to know which part of contributor should be checked. It also closes the class from extension of ignoring rules.

I would propose extending class `IgnoredContributor` with method `#isTrue(String contributorLogin)` and provide `IgnoredContributor` explicitly (by exact class) to `IgnoredCommit`

micd:
Thank you for quick review @wwilk.
I would like to discuss one change that you have requested.

Could you address it, please?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/contributors/IgnoredContributor.java
wwilk: I'm not sure if I follow. 
Let me sum it up:
- method IgnoredContributor.isTrue would be overloaded, another instance of it would take String as argument instead of Contributor object
- IgnoredCommit uses IgnoredContributor instead of duplicating logic

Let me know if I got it right :)

> We would move some logic for ignoring contributor outside the class, because it's caller, who would have to know which part of contributor should be checked.

You are right, using an object like Contributor generally makes it easier to use it - we have to only document Contributor class (what each field means exactly), and stick to this convention when we create this object and when we use it. In this case though we would have to create DefaultContributor in IgnoredCommit, which can be done but it makes it more complicated and harder to maintain (if Contributor grows, you have to remember about creating it with all necessary fields). This idea with overloading method sounds good! We can support both String (with a clear explanation what format it is in) and Contributor.

> It also closes the class from extension of ignoring rules.

I don't think it does that. It's not public API, it's easy to extend a method with additional argument in a scope of the single project. If we have more of them, we can stick to using Contributor itself and pass the problem of creating this object to the caller.

micd: Yes, you got it right!

>  If we have more of them, we can stick to using Contributor itself and pass the problem of creating this object to the caller.

I agree. I will overload method for now and if it would need extension in the future then someone will have to think about using `Contributor` only.

Thank you!

micd:
I have changed `IgnoredContributor` and definition of configuration a bit, because it turned out that login is not provided in `Commit` object.

Double check in `IgnoredContributor#isTrue` is mostly for users like 'shipkit-org', which has different commit's author name and contributor name

subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/contributors/github/Contributors.java
```diff
@@ -15,9 +17,11 @@
      * @param apiUrl address of GitHub api endpoint, for example: "https://api.github.com"
      * @param repository name of GitHub repository, for example: "mockito/mockito"
      * @param readOnlyAuthToken the GitHub auth token
+     * @param ignoredContributors contributors to be ignored - VCS logins (e.g. 'shipkit-org')
```
wwilk: It's not consistent with documentation in ShipkitConfiguration.Git.getIgnoredContributors. Let's just use a reference to this method here.

wwilk:
Thanks! Just two more things :)

micd:
Thank you for collanoration!

PR #615, opened 2018-01-06
Improving UpgradeDependencyPlugin tasks public API

wwilk:
.. by removing UpgradeDependencyExtension from CreatePullRequestTask and ReplaceVersionTask


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -169,7 +169,10 @@ public void execute(String openPullRequestBranch) {
             public void execute(final ReplaceVersionTask task) {
                 task.setDescription("Replaces dependency version in build file.");
                 task.mustRunAfter(CHECKOUT_VERSION_BRANCH);
-                task.setVersionUpgrade(upgradeDependencyExtension);
+                task.setDependencyGroup(upgradeDependencyExtension.getDependencyGroup());
```
mockitoguy: Do we need "deferred configuration" here?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -229,9 +232,9 @@ public void execute(final CreatePullRequestTask task) {
                 task.setGitHubApiUrl(conf.getGitHub().getApiUrl());
                 task.setDryRun(conf.isDryRun());
                 task.setAuthToken(conf.getLenient().getGitHub().getWriteAuthToken());
-                task.setVersionUpgrade(upgradeDependencyExtension);
-                task.setPullRequestTitle(getPullRequestTitle(task));
-                task.setPullRequestDescription(getPullRequestDescription(task));
+                task.setBaseBranch(upgradeDependencyExtension.getBaseBranch());
```
mockitoguy: deferred configuration?

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPluginTest.groovy
```diff
@@ -111,11 +111,16 @@ class UpgradeDependencyPluginTest extends PluginSpecification {
         project.extensions.dependency = "org.shipkit:shipkit:0.1.2"
 
         when:
-        UpgradeDependencyExtension versionUpgrade = project.plugins.apply(UpgradeDependencyPlugin).upgradeDependencyExtension
+        project.plugins.apply(UpgradeDependencyPlugin)
         ReplaceVersionTask task = project.tasks.replaceVersion
 
         then:
-        task.versionUpgrade == versionUpgrade
+        with(task) {
```
mockitoguy: Can you verify if spock automatically adds "assert" in the statements below? Spock by default does not add implicit "assert" for code in blocks.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
wwilk: I don't think so, dependencyGroup is set basing on project property that is passed with command line invocation. I can't think of another place where it would make sense to set it. What do you think?

wwilk: Maybe it shouldn't be a part of UpgradeDependencyExtension, would be better to create an separate object for fields that are extracted from "dependency" property.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
wwilk: Yes good catch, in this case it's needed

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
mockitoguy: My goal was to consider the configuration order when making this change. If you believe that we don't need to use deferred configuration, I trust you :)

upgradeDependencyExtension will be a public object and users can configure it in the build file. Hence, using deferred configuration is desired even if our default use case is "project property". Otherwise the user can set the values on the extension but they are not used. Thoughts?

mockitoguy:
Hey @wwilk, let me know when you're ready with the changes (slack). GitHub UI does not make it easy to know that there are new revisions on the PR. I like this change - it makes the API much cleaner and conventional.

wwilk:
I put getBaseBranch into deferredConfiguration and added a TODO to extract fields that depend on "dependency" property into its own object.

Issue #614, opened 2018-01-06
Validation for tasks created in upgradeDependencyPlugin


wwilk:
## Problem
A few of the tasks created in UpgradeDependencyPlugin depend on "dependency" property that is passed as argument when running the task, eg:
`performVersionUpgrade -Pdependency="org.shipkit:shipkit:1.2.3"`
Unfortunately if you run one of them without this property there is no meaningful error message, it can fail on NPE or some other error during execution phase.

You can find the documentation for upgrade-dependency [here](https://github.com/mockito/shipkit/blob/master/docs/gradle-plugins/upgrade-dependency-plugin.md). 

## Proposed solution
We should add a simple validation checking if the property is null or empty. We have to do this validation at the beginning of execution of task (maybe task.doFirst() would be the best place), we don't want to fail at configuration phase because task might not be actually used.

The message to the user should be meaningful, and actionable (telling the user that they should add the property to task execution).

The tasks affected by this problem are all that depend on values of following fields of [UpgradeDependencyExtension](https://github.com/mockito/shipkit/blob/master/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyExtension.java) class: newVersion, dependencyName, dependencyGroup. These fields are extracted from "dependency" property.

wwilk:
@micd Want to work on this one?

micd:
Yes. I will take a deeper look when I have time.

micd:
Some validation is currently done in `DependencyNewVersionParser#isValid`, but there's a null check before method invocation. 

I would propose to create `Precondition<String>` which will validate dependency property and place it in one of those places:
- `UpgradeDependencyPlugin#apply` just after extracting the property from project. Then after checking  `Precondition#isTrue` we may throw a `GradleException` or do it even within `Precondition#isTrue`.
- `DependencyNewVersionParser#fillVersionUpgradeExtension` we can move whole validation to precondition and throw `GradleException` when precondition returns false. 

Those are two "entry points" of dependency property and we can validate it at the very beginning. 
I like second solution more.  
 
What do you think @wwilk ?

wwilk:
If you set it in `UpgradeDependencyPlugin#apply` validation will fail Gradle configuration phase even if task is never executed. You can try implementing that and then running `./gradlew tasks`. It will fail because of lacking "dependency" property, although **tasks** task doesn't use it.

[...]

PR #620, opened 2018-01-30
Dependency parameter validation for tasks created in upgradeDependencyPlugin

micd:
As described and discussed within #614. I have added lazy configuration to the tasks that depends on ``dependency`` parameter, which will validate if dependency exists.

**Tests**
After running:
- ``gradlew createPullRequest``
- ``gradlew pushVersionUpgrade``
- ``gradlew replaceVersion``
- ``gradlew findOpenPullRequest``
- ``gradlew commitVersionUpgrade``
- ``gradlew performVersionUpgrad``

you should get ``GradleException`` with ``"Dependency property not set. You need to add 'dependency' parameter in order to run this task. E.g.: -Pdependency="org.shipkit:shipkit:1.2.3"``message.

When ``-Pdependency`` parameter set there should not be such exception message.


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -22,44 +23,44 @@
 /**
  * BEWARE! This plugin is in incubating state, so its API may change in the future!
  * The plugin applies following plugins:
- *
+ * <p>
```
wwilk: What are these formatting changes for? It doesn't seems like an improvement, especially removing spaces/tabs from the beginnings of some lines

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -137,18 +138,19 @@ public void execute(String originRepo) {
...
-            }
-        });
+                    task.setGitHubApiUrl(conf.getGitHub().getApiUrl());
+                    task.setAuthToken(conf.getLenient().getGitHub().getReadOnlyAuthToken());
+                    task.setUpstreamRepositoryName(conf.getGitHub().getRepository());
+                    task.setVersionBranchRegex(getVersionBranchName(
+                        upgradeDependencyExtension.getDependencyName(), ReplaceVersionTask.VERSION_REGEX));
+                }
+            });
+        addLazyDependencyVaidation(dependency, findOpenPullRequestTask);
```
wwilk: Typo - vaidation

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -300,7 +307,6 @@ public void execute(PullRequest pullRequest) {
             }
         });
 
-        //TODO: WW add validation for the case when 'dependency' property is not provided
```
wwilk: Thanks for removing the TODO!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
```diff
@@ -369,4 +375,15 @@ public UpgradeDependencyExtension getUpgradeDependencyExtension() {
         return upgradeDependencyExtension;
     }
 
+    private void addLazyDependencyVaidation(final String dependency, Task task) {
+        LazyConfiguration.lazyConfiguration(task, new Runnable() {
+            @Override
+            public void run() {
+                if (dependency == null) {
```
wwilk: Since **dependency** will not change its value (it will be null at the moment when addLazyDependencyValidation is called), we can always throw inside lazyConfiguration, but initialize it only when **dependency** is null.

wwilk:
Please also describe briefly how this code was tested. It has to be unit tested too. The scenario of such test would be:
- set project property (it should be possible by accessing project.getProperties() but I'm not sure)
- apply plugin
- LazyConfiguration#forceConfiguration
- expect exception
The opposite should be tested too (when project property is not null - no exception).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/versionupgrade/UpgradeDependencyPlugin.java
micd: Autoformatting unfortunately. 
I will rollback to the previous formatting. 
Sorry for that.

mockitoguy:
Hey @micd, thank you for submission and sorry for lack of response. New Year's break caught us pretty busy :)

Can you rebase with master so that it merge cleanly?

Is it ok to review commit-by-commit? This PR contains many formatting changes so it would be useful to review commit-by-commit. This is only productive if your commits form a coherent history (e.g. no commits that revert previous commits, etc.).

mockitoguy:
I'll take a stab at it tomorrow.

mockitoguy:
Due to conflicts, I've opened a separate PR with @micd commits. Please review: #627 

Thank for this contribution! We should be able to merge soon. Great error handling is very important for our library!

PR #623, opened 2018-02-05
gradle build scans and stop building with '-i'

### LINKED ISSUE #622: Cleaner CI logs: enroll to Gradle Enterprise and stop building with '-i'
Let's clean up the CI logs. How about we enroll to Gradle Enterprise and avoid building with '-i' parameter (it's --info level logging).

Info level logging is noisy and rarely needed. For example, the checkstyle task logs the entire XSL transformation at info level.

epeee:
fixes #622 

I thought it might be a good idea to use `publishAlways()` only if we are in CI. One can trigger a build scan from non CI environments manually if needed.
Furthermore, we can switch back to `publishOnFailure()` later on, but for now let's have a look at some build scans.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -14,9 +14,7 @@ buildscript {
 apply plugin: 'com.gradle.build-scan'
 
 buildScan {
-    if (System.getenv('CI')) {
-        publishAlways()
-    }
+    publishAlways()
     licenseAgreementUrl = 'https://gradle.com/terms-of-service'
     licenseAgree = 'yes'
 }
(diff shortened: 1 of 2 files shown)


build.gradle
```diff
@@ -6,10 +6,21 @@ buildscript {
...
         classpath 'org.shipkit:shipkit:1.0.7'
+        classpath 'com.gradle:build-scan-plugin:1.11'
         classpath 'ru.vyarus:gradle-animalsniffer-plugin:1.3.0'
     }
 }
 
+apply plugin: 'com.gradle.build-scan'
+
+buildScan {
+    if (System.getenv('CI')) {
```
mockitoguy: I'd say we always publish, even from local. If you are out of network, build scans degrade gracefully and scan is simply not sent. We can always turn it back to CI-only later, after we learn how it works for us.

Issue #629, opened 2018-02-14
mergePullRequest is failing due to 'No value has been specified for property 'versionBranch''


epeee:
It looks like automerge did not work:

> [upgradeDownstream] [performVersionUpgrade] :mergePullRequest FAILED
> [upgradeDownstream] [performVersionUpgrade] 
> [upgradeDownstream] [performVersionUpgrade] 
> [upgradeDownstream] [performVersionUpgrade] BUILD FAILED
> [upgradeDownstream] [performVersionUpgrade] 
> [upgradeDownstream] [performVersionUpgrade] Total time: 26.173 secs
> [upgradeDownstream] [performVersionUpgrade] FAILURE: Build failed with an exception.
> [upgradeDownstream] [performVersionUpgrade] 
> [upgradeDownstream] [performVersionUpgrade] * What went wrong:
> [upgradeDownstream] [performVersionUpgrade] A problem was found with the configuration of task ':mergePullRequest'.
> [upgradeDownstream] [performVersionUpgrade] > No value has been specified for property 'versionBranch'.

see [travis ci log](https://travis-ci.org/mockito/shipkit/builds/340491561) for details.

sadly, we have no buildscan for this build.

epeee:
This problem is solved now.
Auto merging still does not work as expected -> I will create a separate ticket to fix the remaining problem.

PR #630, opened 2018-02-14
fix auto merging

epeee:
this one should allow us to merge pr's automatically again (see #629).
It looks like we somehow lost this assignment. Without it `task.setVersionBranch` is never called.


mockitoguy:
Thank you! It would be great to have integration tests for this functionality at some point.

PR #631, opened 2018-02-15
fix some warnings

epeee:
.. which I stumbled over while working on #630.
let's merge it after #630 and the upcoming shipkit upgrade pr.


mockitoguy:
Good idea!

PR #637, opened 2018-02-17
Migrate to Java8 and Gradle4 - new major version

### LINKED ISSUE #624: Upgrade to Java8 and Gradle4
How about we upgrade to Java8 and Gradle4?

We can produce Shipkit 2.0 that would have only one incompatible change: Java8 + Gradle4. For uses on lower versions of those, they can use Shipkit 1.0 (compatible feature-wise, public API wise). For users that want latest improvements, they need to be on Java8 + Gradle4. This is the best of both worlds - we still have a working version for old Java and Gradle but we are not blocked using latest features of Java (very needed, see the bug #621)

Thoughts?

epeee:
fixes #624 

@mockito/shipkit-developers let's sync about merging this one/discuss which ticket we should work on before merging this one.


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,7 +1,6 @@
-#Thu Sep 14 09:54:35 CEST 2017
+#Fri Feb 16 18:55:01 CET 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.14.1-all.zip
-distributionSha256Sum=88a910cdf2e03ebbb5fe90f7ecf534fc9ac22e12112dc9a2fee810c598a76091
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.0.2-bin.zip
```
szpak: Why not 4.5.1?

szpak: `distributionSha256Sum` is missing

subprojects/shipkit/src/test/groovy/testutil/GradleVersionsDeterminer.groovy
```diff
@@ -24,7 +24,7 @@ trait GradleVersionsDeterminer {
             case null:
                 return [currentGradleVersion()]
             case QUICK_GRADLE_VERSIONS_VALUE:
-                return [currentGradleVersion(), "4.1", "3.5.1", "2.14.1"].unique()
+                return [currentGradleVersion(), "4.2.1", "4.0.2"].unique()
```
szpak: Did you mean `4.5.1` instead of `4.2.1`?

version.properties
```diff
@@ -1,6 +1,6 @@
 #Version of the produced binaries. This file is intended to be checked-in.
 #It will be automatically bumped by release automation.
-version=1.0.13
+version=2.0.0
```
szpak: I wonder (rather more as a consideration than a real suggestion) if those are all of the backward incompatible changes that are planned by the way of bumping the major version number?

szpak: Ok, reading the body of this PR I realized that it is taken into consideration :).

gradle/wrapper/gradle-wrapper.properties
mockitoguy: >Why not 4.5.1?

Good question! We don't want to force clients to latest Gradle, we want to allow them to stay at 4.2 or 4.3.

subprojects/shipkit/src/test/groovy/testutil/GradleVersionsDeterminer.groovy
epeee: thx for pointing that out!
It looks like we do have a problem with gradle 4.5.1. I will create a separate ticket for this one.

epeee:
@mockitoguy no, I just wanted to ask if we are ready for merging this one or if we have to finish work on some other tasks.
but it looks like we are ready  :smiley:

gradle/wrapper/gradle-wrapper.properties
szpak: That's why I created regression tests with older Gradle version(s) long time ago. Thanks to that the project can be built with the latest version while (to the extent covered by tests :) ) it is still compatible with older versions.

Issue #634, opened 2018-02-16
mergePullRequest is not merging the pr due to 'No checks defined in pull request'


epeee:
After solving #629 we are still not able to automatically merge pr's. 

> [upgradeDownstream] [performVersionUpgrade] :mergePullRequest
> [upgradeDownstream] [performVersionUpgrade]   [INCUBATING] merge pull requests is incubating and may change in any version.
> [upgradeDownstream] [performVersionUpgrade] Waiting for status of a pull request in repository 'mockito/shipkit' between base = 'master' and head = 'mockito:upgrade-shipkit-to-1.0.12'.
> [upgradeDownstream] [performVersionUpgrade] No checks defined in pull request https://api.github.com/repos/mockito/shipkit/pulls/633! Merge aborted.

see [travis build log](https://travis-ci.org/mockito/shipkit/builds/342197410#L1116) for details.

mockitoguy:
Thank you for reporting! Do you want to take a stab at it?

epeee:
yep, I will have a look at this one.

epeee:
Doing one retry does the trick, let's close this one.
There is still one problem remaining until we are able to automatically merge pr's. I will create a new issue.

PR #638, opened 2018-02-18
refactor GitHubStatusCheck

epeee:
this one is related to #634

there are two possibilities why the automatic merge now does not work

- the first check is too fast and 'statuses' are not set yet:
   I have added a retry in order to perform the check for a 2nd time. I'm not entirely happy with this solution but it looks like we need something like that.

 - the pullrequest sha is not the one we are supposed to use:
  I have added a log message to be able to check whether we are using the correct sha

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
@@ -18,32 +18,33 @@ public class GitHubStatusCheck {
     private MergePullRequestTask task;
     private GitHubApi gitHubApi;
     private final int amountOfRetries;
+    private final long defaultInitialTimeout;
 
-    public GitHubStatusCheck(MergePullRequestTask task, GitHubApi gitHubApi, int amountOfRetries) {
+    public GitHubStatusCheck(MergePullRequestTask task, GitHubApi gitHubApi, int amountOfRetries, long defaultInitialTimeout) {
         this.task = task;
         this.gitHubApi = gitHubApi;
         this.amountOfRetries = amountOfRetries;
+        this.defaultInitialTimeout = defaultInitialTimeout;
     }
 
     public GitHubStatusCheck(MergePullRequestTask task, GitHubApi gitHubApi) {
         this.task = task;
         this.gitHubApi = gitHubApi;
         this.amountOfRetries = 20;
+        this.defaultInitialTimeout = TimeUnit.SECONDS.toMillis(10);
     }
 
     public PullRequestStatus checkStatusWithRetries() throws IOException, InterruptedException {
         int timeouts = 0;
         while (timeouts < amountOfRetries) {
             JsonObject status = getStatusCheck(task, gitHubApi);
-            if (status.getCollection("statuses") == null || status.getCollection("statuses").size() == 0) {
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -17,43 +18,48 @@
...
         int timeouts = 0;
         while (timeouts < amountOfRetries) {
             JsonObject status = getStatusCheck(task, gitHubApi);
-            if (status.getCollection("statuses") == null || status.getCollection("statuses").size() == 0) {
+            if (timeouts > 0 && isNullOrEmpty(status, "statuses")) {
                 return PullRequestStatus.NO_CHECK_DEFINED;
-            }
-
-            if (allStatusesPassed(status)) {
+            } else if (!isNullOrEmpty(status, "statuses") && allStatusesPassed(status)) {
```
mockitoguy: We're missing a scenario here: what if the check fails? If it fails, we should return failure rather than attempt to retry with timeout. I know that this code was like that before :) We can fix it in the next round.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -17,43 +18,48 @@
...
     }
 
+    private boolean isNullOrEmpty(JsonObject status, String key) {
+        return status.getCollection(key) == null || status.getCollection(key).size() == 0;
+    }
+
     private JsonObject getStatusCheck(MergePullRequestTask task, GitHubApi gitHubApi) throws IOException {
-        String statusesResponse = gitHubApi.get("/repos/" + task.getUpstreamRepositoryName() + "/commits/" + task.getPullRequestSha() + "/status");
+        String relativeUrl = "/repos/" + task.getUpstreamRepositoryName() + "/commits/" + task.getPullRequestSha() + "/status";
+        LOG.lifecycle("Using {} for status check", relativeUrl);
```
mockitoguy: It would be nice to emit how long we have been waiting so far (minor suggestion).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -17,43 +18,48 @@
...
         this.amountOfRetries = 20;
+        this.defaultInitialTimeout = TimeUnit.SECONDS.toMillis(10);
     }
 
     public PullRequestStatus checkStatusWithRetries() throws IOException, InterruptedException {
         int timeouts = 0;
         while (timeouts < amountOfRetries) {
             JsonObject status = getStatusCheck(task, gitHubApi);
-            if (status.getCollection("statuses") == null || status.getCollection("statuses").size() == 0) {
+            if (timeouts > 0 && isNullOrEmpty(status, "statuses")) {
```
mockitoguy: Nice change! It would be good to add code comment explaining why we are coding the algorithm this way (minor suggestion).

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
epeee: yep, we definitely should do so, thx!

PR #639, opened 2018-02-19
add gradle version to test name

epeee:
in order to make it easier to analyze integration test problems (e.g. one test is just failing while using a certain gradle version).

test names will look like 

> all tasks in dry run (using gradle version 4.0.2)
> all tasks in dry run (using gradle version 4.2.1)

instead of

> all tasks in dry run[0]
> all tasks in dry run[1]


mockitoguy:
Thank you! Merging.

PR #642, opened 2018-02-20
update checkstyle to 8.8

epeee:
let's use latest version of checkstyle. we can do so now since we are using java8 now.

What about using `[ci skip-release]` while merging this one?
It is just a minor change and doesn't have to trigger a release.


mockitoguy:
>What about using [ci skip-release] while merging this one?
It is just a minor change and doesn't have to trigger a release.

If the binary changes, we should release it. Otherwise we have unproven - unuseful - unreleased code in the main branch (work-in-progress == cost). Let's ship it!

mockitoguy:
Thank you!

PR #648, opened 2018-02-25
make use of codenarc v1.1

epeee:


subprojects/shipkit/src/test/groovy/org/shipkit/gradle/ShipkitJavaIntegTest.groovy
```diff
@@ -92,6 +91,6 @@ class ShipkitJavaIntegTest extends GradleSpecification {
 :performRelease"""
 
         where:
-            gradleVersionToTest << determineGradleVersionsToTest()
+        gradleVersionToTest << determineGradleVersionsToTest()
```
szpak: Maybe i would be better to adjust the rule to permit/force indentation in Spock tests? :)

epeee: yes, we could do so. On the other hand we do have consistent indentation now.

mockitoguy: Can we try to configure codenarc so that it is possible to have one liner "when:/then:"? 

```Java
when: conf.team.contributors = ["ala:"]
then: thrown(TeamParser.InvalidInput.class)
```

If above it's not possible, I'd say it's better to have consistent whitespace even though it is more restrictive. If can however make use of Spock more convenient, that would be better. @epeee, do you want to take a quick look at it?

mockitoguy:
I left one comment for consideration, otherwise, looks great!

subprojects/shipkit/src/test/groovy/org/shipkit/gradle/ShipkitJavaIntegTest.groovy
epeee: yes, I will have a look at it.

epeee: I have disabled the `IndentationRule` for now and reverted the spock label related changes. 
It looks like codenarc master already has some fixes in that area. Let's pick them up once they do a new release and give it another try.

PR #652, opened 2018-02-28
Use put instead of post for merge pr rest request

epeee:
It looks like I somehow forgot to change the request method while migrating to the new pr merge rest api.
We have to use put instead of post, see https://developer.github.com/v3/pulls/#merge-a-pull-request-merge-button for details.

Let's use put instead of post now.


mockitoguy:
At some point we should think about some kind of an integration test that actually tests the entire pipeline.

PR #660, opened 2018-03-07
Enable gradle build cache

epeee:
since we recently migrated to gradle 4, let's also enable the gradle build cache now.


mockitoguy:
Absolutely!

PR #661, opened 2018-03-07
Migrate codebase towards java8

epeee:
Let's do some housekeeping and migrate the codebase towards java8 features. 
This pr mainly contains changes which are actually java7 improvements.


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/ZipUtil.java
```diff
@@ -21,18 +21,10 @@ public static ZipFile openZipFile(File file) {
     }
 
     public static boolean fileContainsEntry(File file, String entry) {
-        ZipFile zip = null;
-        try {
-            zip = openZipFile(file);
+        try (ZipFile zip = openZipFile(file)) {
```
mockitoguy: This is super nice!

PR #662, opened 2018-03-09
org.shipkit.gradle-plugin automatically avoids releases when publications are identical

### LINKED ISSUE #487: Include compare publications plugin in org.shipkit.gradle-plugin
I suggest we implement @epeee's suggestion from: https://github.com/mockito/shipkit/pull/479#discussion_r142850584

@wwilk, thoughts?

epeee:
... and also make compare-publications plugin a default for ShipkitGradlePlugin users.

fixes #487


mockitoguy:
Nice change - absolutely useful! Have you run a dry run release with this change? Let's make sure it's not going to break stuff :)

epeee:
thx for the reminder about doing a dry run.
I just did so locally and it looks good. I run two tests, once still having the 'org.shipkit.compare-publications' plugin applied in shipkit.gradle and once without. Comparing publications does work on both cases.

mockitoguy:
Great! Merging!

PR #669, opened 2018-03-13
Gradle build script cleanup

epeee:
No longer apply 'compare-publications' plugin in gradle file because it is now applied by default via 'org.shipkit.gradle-plugin'.


mockitoguy:
Nice!

Issue #673, opened 2018-03-21
Automate releases of linkedin/play-parseq project


mockitoguy:
Automate releases of https://github.com/linkedin/play-parseq project 

Ticket in target project: https://github.com/linkedin/play-parseq/issues/9

Goals:
 - automated releases to Maven Central
 - release notes

Complexity:
 - it is an SBT project so Gradle automation will be wrapping SBT

I'm really looking forward to empowering more projects to full release automation!!!

PR #674, opened 2018-03-22
Exposed "org.shipkit.base" plugin

mockitoguy:
- started using  "org.shipkit.bintray" instead of "com.jfrog.bintray" because this way we can honor user's settings (usability, related to Gradle's order of evaluation)
 - java8 lambda in few classes I worked on
 - updated docs, started using 'org.shipkit.base' plugin for non standard java projects that we want to support (like SBT plugin)

I suggest to review per commit.

Changes are related to #673

### DIFF
--- a/docs/features/publishing-binaries.md
+++ b/docs/features/publishing-binaries.md
@@ -66,7 +66,7 @@ shipkit {
 }
 
 allprojects {
-    plugins.withId('com.jfrog.bintray') {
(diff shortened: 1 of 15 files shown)


mockitoguy:
I don't see why the PR is failing and Travis CI does not show any builds (seems that Travis is not very stable... time for Circle CI? :)

I'm feeling lucky and I'm merging. Worst case scenario, the build will fail :)

Issue #675, opened 2018-03-22
Annoying 'Gradle Test Executor' finished with non-zero exit value 137


mockitoguy:
Hey team!

Can someone take a look at the issue that intermittently hits us on CI? I'm seeing this issue a couple of time per week and I'm forced to re-run builds.

```
> Process 'Gradle Test Executor 2' finished with non-zero exit value 137
```

Example full build log: https://travis-ci.org/mockito/shipkit/builds/356200130

Suggestions:
 - SO and google show hits for this problem
 - perhaps we need to fiddle with java/memory settings for test execution?

epeee:
yep, most likely a memory problem. I will have a look at it.

mockitoguy:
Great! Thank you!

mockitoguy:
Failed twice for this reason between yesterday and today :( @epeee, rescue us :)

polson:
How did you guys fix this?  I have the same issue

ziakhalid:
This is actually a memory issue, Generally, Docker container has a memory limit of 4G so you need to take care that your java heap doesn't cross that limit, there are few solutions to it I am taking reference of android

1. Add this to gradle.properties (change size according to your need)
    `org.gradle.jvmargs=-Xmx10248m -XX:MaxPermSize=256m`

2. Add this to your build.gradle

`android.testOptions.unitTests.all { maxHeapSize = "1024m" }`

or 

```
android {
    testOptions {
        unitTests {
            // Any other configurations
            all {
                maxHeapSize = "1024m"
            }
        }
    }
```

If you are still running into OOM issues you can also limit the max workers for gradle:` ./gradlew test --max-workers 4`

Hope this helps.

PR #678, opened 2018-03-23
remove --daemon from travis.yml

epeee:
let's remove --daemon from travis.yml. We added it back then because travis did default to very high xmx values and adding the --daemon did solve our memory problems previously.
this pr is related to #675 although it might need another pr to do some more tuning.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,4 +19,4 @@ install:
  - true
 
 script:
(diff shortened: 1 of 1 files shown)


mockitoguy:
I think it is a good idea because it eliminates complexity. I doubt it will fix the unit test issue :)

PR #649, opened 2018-02-27
fix UpgradeDependencyPlugin for gradle 4.5.1

epeee:
fixes #641 
and add gradle 4.5.1 to set of gradle versions used for integration testing.

PR #681, opened 2018-03-26
Fix memory problems in CI

epeee:
Issue: #675

Travis is using `_JAVA_OPTIONS="-Xmx2048m -Xms512m"` by default. 
Let's adjust this values to our needs.

The build is failing now because mockito project recently updated to Gradle 4.6 and we are heading into something similar to https://github.com/mockito/shipkit/pull/649 now (see [gist](https://gist.github.com/shipkit-org/1a4425d5698b990ca6978256d14d5778) for details).
Let's have a look at this one in a separate ticket.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,4 +22,4 @@ before_script:
   - _JAVA_OPTIONS=
 
 script:
-  - ./gradlew build testDownstream -PcheckJava8Compatibility -s && ./gradlew ciPerformRelease -s
+  - ./gradlew build testDownstream -PcheckJava8Compatibility -Xmx1024m -Xms128m -s && ./gradlew ciPerformRelease -s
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,5 +18,8 @@ branches:
 install:
  - true
 
+before_script:
+  - _JAVA_OPTIONS="-Xmx1024m -Xms128m"
```
mockitoguy: If this is needed for test, why don't we configure it in the *.gradle file?

```
test.maxHeapSize = ...
```

This way, we reduce the secret sauce we keep in CI configuration.

mockitoguy:
Thank you for working on this!!! It is super annoying when Travis fails for us :/

epeee:
@mockitoguy thx for your feedback!
I somehow have to reduce Xmx and Xms of the gradle process because the default Xmx and Xms values in the container based environment we are running on travis are pretty high. As a result we therefore would  hit the memory limit without restricting Xms and Xms. 
I'm now using 'org.gradle'jvmargs' in gradle.properties in order to do so.

PR #685, opened 2018-03-29
Fix a bunch of javadoc warnings

epeee:
I stumbled over this one while working on the CI memory issue.


epeee:
It looks like we are getting this warnings because we are importing `ShipkitConfiguration` and using `ShipkitConfiguration.Team#getIgnoredContributors()`.
If we are just using `Team#getIgnoredContributors()` and changing the import to `Team` the warning also goes away.
I changed it to fully qualified since we already have it like that in quite some other places.

Issue #437, opened 2017-09-27
Add "contributorTestRelease" task to simplify testing


mockitoguy:
### Problem

- Contributors cannot use "testRelease" task because it requires credentials
- Downstream testing task cannot use "testRelease" for the same reason

Instead they need to use complicate sequence of tasks:

```
./gradlew releaseNeeded performRelease releaseCleanUp -PdryRun -x gitPush -x bintrayUpload
```

### Solution

Let's create "contributorTestRelease" task that wraps other tasks similarly how "testRelease" does it.

- [ ] - add "contributorTestRelease" task.
- [ ] - update CONTRIBUTING.md
- [ ] - see if we can update "test-downstream" plugin so that we can leverage the new task

### Implementation

Check out how "testRelease" task is configured in the "ReleasePlugin"

PR #689, opened 2018-04-01
Added 'contributorTestRelease' task

mockitoguy:
New task streamlines testing of Shipkit for contributors.

Addresses most of #437

Tested with 'shipkit-example' - build log: https://gist.github.com/mockitoguy/4d602ab652527d9d8af5088a5fce7d00

I suggest to review commit-by-commit!

Also tested with 'shipkit' project itself - build log: https://gist.github.com/mockitoguy/ff823427f836d33ceb468feb5ecb2702

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleasePlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleasePlugin.java
@@ -1,26 +1,22 @@
 package org.shipkit.internal.gradle.release;
 
-import org.gradle.api.Action;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
-import org.gradle.api.Task;
 import org.shipkit.gradle.exec.ShipkitExecTask;
 import org.shipkit.gradle.git.IdentifyGitBranchTask;
 import org.shipkit.gradle.notes.UpdateReleaseNotesTask;
 import org.shipkit.internal.gradle.git.GitBranchPlugin;
+import org.shipkit.internal.gradle.git.GitPlugin;
 import org.shipkit.internal.gradle.notes.ReleaseNotesPlugin;
 import org.shipkit.internal.gradle.notes.tasks.UpdateReleaseNotes;
-import org.shipkit.internal.gradle.version.VersioningPlugin;
-import org.shipkit.internal.gradle.git.GitPlugin;
 import org.shipkit.internal.gradle.util.TaskMaker;
 import org.shipkit.internal.gradle.util.TaskSuccessfulMessage;
-
-import java.util.function.Supplier;
+import org.shipkit.internal.gradle.version.VersioningPlugin;
 
 import static java.util.Arrays.asList;
+import static org.shipkit.internal.gradle.exec.ExecCommandFactory.execCommand;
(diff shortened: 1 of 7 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleasePlugin.java
```diff
@@ -55,49 +53,49 @@ public void apply(final Project project) {
...
+            t.setDescription("Similar to 'testRelease' but excludes tasks that require Git/Bintray permissions. " +
+                "Useful for contributors who don't have the permissions.");
 
-                //using finalizedBy so that all clean up tasks run, even if one of them fails
-                t.finalizedBy(GitPlugin.PERFORM_GIT_COMMIT_CLEANUP_TASK);
-                t.finalizedBy(GitPlugin.TAG_CLEANUP_TASK);
-            }
+            t.getExecCommands().add(execCommand("Performing release in dry run, with cleanup",
+                asList("./gradlew", RELEASE_NEEDED, PERFORM_RELEASE_TASK, RELEASE_CLEAN_UP_TASK, "-PdryRun",
+                    "-x", GitPlugin.GIT_PUSH_TASK, "-x", ShipkitBintrayPlugin.BINTRAY_UPLOAD_TASK)));
```
epeee: Due to `-x bintrayUpload` this one will fail if we are running `contributorTestRelease` on a gradle plugin project (because we are using `publishPlugins` task in this case instead of `bintrayUpload`).
What about also covering this case?

mockitoguy: Great catch, thank you!!! How did you find out about this problem? I will try to figure out a solution to it.

mockitoguy:
@epeee, thank you review and for catching this problem! I updated the code and made additional testing to ensure it works correctly.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/ReleasePlugin.java
epeee: I had a similar problem recently. Therefore I was a little bit curious about it.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/bintray/BintrayReleasePluginTest.groovy
```diff
@@ -38,4 +40,14 @@ class BintrayReleasePluginTest extends PluginSpecification {
...
     }
+
+    def "configures contributor test"() {
+        when:
+        project.plugins.apply(BintrayReleasePlugin)
+        project.plugins.apply(ShipkitBintrayPlugin)
+
+        then:
+        ShipkitExecTask contrib = project.tasks.getByName(ReleasePlugin.CONTRIBUTOR_TEST_RELEASE_TASK)
+        contrib.execCommands*.commandLine.toString() == "[[./gradlew, releaseNeeded, performRelease, releaseCleanUp, -PdryRun, -x, gitPush, -x, bintrayUpload]]"
```
epeee: Thank you for adding test cases!

PR #690, opened 2018-04-01
Inform the user if no files are uploaded to bintray

### LINKED ISSUE #686: bintrayUpload should inform if there are no files uploaded
If possible, let's make "bintrayUpload" inform the user if no files are uploaded. Otherwise the behavior is confusing: the task succeeds but nothing is uploaded:

```
$ ./gradlew bintrayUpload

> Configure project : 
  Building version '0.8.2' (value loaded from 'version.properties' file).

> Task :bintrayUpload 
:bintrayUpload - publishing to Bintray
  - dry run: false, version: 0.8.2, Maven Central sync: true
  - user/org: franklinyinanding/linkedin, repository/package: test-repo/play-parseq


BUILD SUCCESSFUL in 2s
1 actionable task: 1 executed
```

epeee:
fixes #686 

if no artifacts are uploaded the log now looks like:
```
  Building version '1.5.23'.
:bintrayUpload - publishing to Bintray
  - dry run: false, version: 1.0.0, Maven Central sync: false
  - user/org: shipkit-bot/shipkit.org, repository/package: shipkit/shipkit-example
No artifacts have been published to bintray!
```

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
@@ -56,7 +56,7 @@ public class ShipkitBintrayPlugin implements Plugin<Project> {
         });
 
         bintrayUpload.doLast(task -> {
-                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask)task;
+                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask) task;
                 if ((bintrayUploadTask.getFileUploads() == null || bintrayUploadTask.getFileUploads().length == 0) &&
                     (bintrayUploadTask.getConfigurationUploads() == null || bintrayUploadTask.getConfigurationUploads().length == 0) &&
                     (bintrayUploadTask.getPublicationUploads() == null || bintrayUploadTask.getPublicationUploads().length == 0)) {
@@ -64,7 +64,6 @@ public class ShipkitBintrayPlugin implements Plugin<Project> {
                 }
             }
         );
-
 
         final BintrayExtension.PackageConfig pkg = bintray.getPkg();
         pkg.setPublicDownloadNumbers(true);
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
@@ -55,6 +55,17 @@ public class ShipkitBintrayPlugin implements Plugin<Project> {
             LOG.lifecycle(welcomeMessage);
         });
 
+        bintrayUpload.doLast(task -> {
+                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask)task;
+                if ((bintrayUploadTask.getFileUploads() == null || bintrayUploadTask.getFileUploads().length == 0) &&
+                    (bintrayUploadTask.getConfigurationUploads() == null || bintrayUploadTask.getConfigurationUploads().length == 0) &&
+                    (bintrayUploadTask.getPublicationUploads() == null || bintrayUploadTask.getPublicationUploads().length == 0)) {
+                    LOG.lifecycle("No artifacts have been published to bintray!");
+                }
+            }
+        );
+
+
         final BintrayExtension.PackageConfig pkg = bintray.getPkg();
         pkg.setPublicDownloadNumbers(true);
         pkg.getVersion().getGpg().setSign(true);


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -55,6 +55,16 @@ public void apply(final Project project) {
             LOG.lifecycle(welcomeMessage);
         });
 
+        bintrayUpload.doLast(task -> {
+                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask) task;
+                if ((bintrayUploadTask.getFileUploads() == null || bintrayUploadTask.getFileUploads().length == 0) &&
```
mockitoguy: How about we use "doFrist" and fail if no artifacts to upload are configured? The exception message should tell the user to either:
 - disable bintrayUpload task / avoid running it
 - or configure the task so that it has artifacts to publish

It would be great to drop a ticket to Bintray plugin on GitHub, report this problem and offer PR to their code!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -55,6 +55,16 @@ public void apply(final Project project) {
...
+                    (bintrayUploadTask.getConfigurationUploads() == null || bintrayUploadTask.getConfigurationUploads().length == 0) &&
```
mockitoguy: All those getters return object of the same type so you can reduce the duplication by adding a method like "isEmpty(getter1, getter2)".

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
epeee: It looks like we can't use `doFirst` because the files for upload are resolved during task execution.
Ad open a ticket to Bintray: Yes, I will do so, thx!

epeee:
@mockitoguy Thank you! I have reworked this pr now.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -55,6 +57,17 @@ public void apply(final Project project) {
             LOG.lifecycle(welcomeMessage);
         });
 
+        bintrayUpload.doLast(task -> {
+                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask) task;
+                if (bintrayUploadTask.isEnabled() && hasNoArtifacts(bintrayUploadTask)) {
+                    throw new GradleException("'" + bintrayUploadTask.getName() + "' currently has no artifacts for publication.\n" +
+                        " - if the current project '" + project.getPath() + "' is not supposed to publish to bintray -> let's disable the bintray task (bintrayUpload.enabled = false)\n" +
+                        " - if it is supposed to publish to bintray -> configure the task so that it has artifacts to publish\n"
```
mockitoguy: Great! Can you point to official Bintray documentation!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -55,6 +57,17 @@ public void apply(final Project project) {
             LOG.lifecycle(welcomeMessage);
         });
 
+        bintrayUpload.doLast(task -> {
+                BintrayUploadTask bintrayUploadTask = (BintrayUploadTask) task;
+                if (bintrayUploadTask.isEnabled() && hasNoArtifacts(bintrayUploadTask)) {
+                    throw new GradleException("'" + bintrayUploadTask.getName() + "' currently has no artifacts for publication.\n" +
+                        " - if the current project '" + project.getPath() + "' is not supposed to publish to bintray -> let's disable the bintray task (bintrayUpload.enabled = false)\n" +
```
mockitoguy: "let's disable the bintray task" reads somewhat odd. I suggest "please disable the bintray task"

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -55,6 +57,17 @@ public void apply(final Project project) {
             LOG.lifecycle(welcomeMessage);
         });
 
+        bintrayUpload.doLast(task -> {
```
mockitoguy: I suggest to use "doFirst" - it is more standard approach for pre-condition validation.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/bintray/ShipkitBintrayPlugin.java
```diff
@@ -105,6 +118,16 @@ public void apply(final Project project) {
         });
     }
 
+    private boolean hasNoArtifacts(BintrayUploadTask bintrayUploadTask) {
```
mockitoguy: Nice!

epeee:
@mockitoguy the reason why I have chosen `doLast` is that we do not have to resolve possible artifacts at our own in this case (since resolving artifacts will be done in the task action of bintrayupload task).
If we'd like to do it in `doFirst` we'd have to somehow duplicate this logic, isn't it?

mockitoguy:
>@mockitoguy the reason why I have chosen doLast is that we do not have to resolve possible artifacts at our own in this case (since resolving artifacts will be done in the task action of bintrayupload task).

Oh I see now, thank you for making it clear. I'm approving it. Can your run './gradlew testRelease' on a sample project to make sure the new version of Bintray plugin is good for us? Thanks!

PR #691, opened 2018-04-02
Fail fast when core plugins are not applied to root project

### LINKED ISSUE #671: Fail fast when plugin not applied to root project
Our core plugins
 - org.shipkit.java
 - org.shipkit.gradle-plugin

need to be applied to the root project.

This is not intuitive as evidenced by a problem reported at #655

Let's implement a short term solution and fail early and gracefully when the plugin is applied to a subproject rather than to a root project.

Example implementation we already use in a different plugin: 

https://github.com/mockito/shipkit/blob/59aad490aa3fcd0955bd972ce939f44b8d2e72e4/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitOriginPlugin.java#L32-L37

Long term plan is to make the plugins work in a submodule, too.

epeee:
fixes #671

as described in the ticket, let's check if our core plugins are applied to the root project.

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/ProjectUtil.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/ProjectUtil.java
@@ -3,18 +3,16 @@ package org.shipkit.internal.gradle.util;
 import org.gradle.api.GradleException;
 import org.gradle.api.Project;
 
-import java.util.Optional;
-
 public class ProjectUtil {
 
     public static void requireRootProject(final Project project, final Class<?> clazz) {
-        requireRootProject(project, clazz, Optional.empty());
+        requireRootProject(project, clazz, null);
     }
 
-    public static void requireRootProject(final Project project, final Class<?> clazz, final Optional<String> explanation) {
+    public static void requireRootProject(final Project project, final Class<?> clazz, final String explanation) {
         if (project.getParent() != null) {
             throw new GradleException("Plugin '" + clazz.getSimpleName() + "' is intended to be applied only root project.\n" +
-                (explanation.isPresent() ? explanation.get() + "\n" : "" ) +
+                (explanation != null ? explanation + "\n" : "" ) +
                 "Please apply this plugin to the root project instead of '" + project.getPath() + "'.");
         }
     }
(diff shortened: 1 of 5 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/ProjectUtil.java
```diff
@@ -0,0 +1,21 @@
...
+
+import java.util.Optional;
+
+public class ProjectUtil {
+
+    public static void requireRootProject(final Project project, final Class<?> clazz) {
+        requireRootProject(project, clazz, Optional.empty());
+    }
+
+    public static void requireRootProject(final Project project, final Class<?> clazz, final Optional<String> explanation) {
```
mockitoguy: Minor: Optional is an overkill in this instance and allowing null is simpler and cleaner.

Optionals are very useful for output parameters (e.g. return types) because they force the clients to code against NPE. For input parameters, they are contrived.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/ProjectUtilTest.groovy
```diff
@@ -0,0 +1,31 @@
...
+        when:
+        ProjectUtil.requireRootProject(project, ProjectUtilTest)
+
+        then:
+        noExceptionThrown()
+    }
+
+    def "requireRootProject fails for non-rootProject"() {
+        def project = Mock(Project)
+        project.path >> ':subproject'
```
mockitoguy: Why is path relevant? If not, I suggest to remove it to keep the test clean!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/ProjectUtilTest.groovy
```diff
@@ -0,0 +1,31 @@
...
+
+        when:
+        ProjectUtil.requireRootProject(project, ProjectUtilTest)
+
+        then:
+        noExceptionThrown()
+    }
+
+    def "requireRootProject fails for non-rootProject"() {
+        def project = Mock(Project)
```
mockitoguy: Minor: I suggest to make it an initialized field to make the test cleaner and eliminate small duplication.

epeee:
thx for your feedback!
merging...

PR #696, opened 2018-04-08
Added design spec for building snapshot versions

mockitoguy:
Let's try out using design specs / markdown for design. This way, we can review them as pull requests and get shared understanding in the team.

When reviewing, I suggest to click "view" button in GitHub UI. This way you can review the html of markdown rather than the source code (much easier to read!)


mockitoguy:
@epeee, thank you for review! @mockito/shipkit-developers, any feedback about the design spec?

mockitoguy:
If there is no further feedback, I'm merging!

PR #709, opened 2018-04-15
[SHIPKIT-513] Automated commit message should have CI build ID

ArturSkowronski:


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitPlugin.java
```diff
@@ -55,13 +57,15 @@
 
     public void apply(final Project project) {
         final ShipkitConfiguration conf = project.getPlugins().apply(ShipkitConfigurationPlugin.class).getConfiguration();
+        String commitMessage = generateCommitMessagePostfix(conf, getProperty("TRAVIS_BUILD_NUMBER"));
```
mockitoguy: GitPlugin should be decoupled from Travis, so that we can use it for Circle CI, etc. Take a look at #514 - we're adding new code to TravisPlugin.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/gradle/util/CommitMessageUtilsTest.groovy
```diff
@@ -0,0 +1,40 @@
+package org.shipkit.internal.gradle.util
+
+import spock.lang.Specification
+
+import org.shipkit.gradle.configuration.ShipkitConfiguration
+
+class CommitMessageUtilsTest extends Specification {
```
mockitoguy: Nice test coverage! Thanks!

ArturSkowronski:
@mockito/shipkit-developers Thanks for the feedback :) I'll try to implement suggestion quickly

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/git/GitPlugin.java
ArturSkowronski: That's why I exposed it to CommitMessageUtil but using Travis Plugin for that seems to be really better way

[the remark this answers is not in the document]
ArturSkowronski: Good catch! Thanks

ArturSkowronski:
@mockitoguy I did a few changes (it took some time but that was really busy week...), I still have a question what do you @mockito/shipkit-developers suggest is the best way of testing changes like that. I see there are not many automatic tests for interoperability of plugins and I'm looking for a better way than doing that manually - what itself is hard when it's touching Travis builds.

Do you have any good idea?

PR #701, opened 2018-04-12
Waiting time logging during github status checks

micd:
Added log info how long we have been waiting so far during github status checks 
#657

PR #710, opened 2018-04-17
Extracting retry logic to separate class

micd:
As proposed by @mockitoguy in #701 I have extracted retry logic to separate class to make this code more testable and reusable.


subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
```diff
@@ -103,8 +93,16 @@ private JsonObject findFirstError(Collection<JsonObject> statuses) {
...
         return statusDetails != null &&
-               statusDetails.getString("state") != null &&
-               (statusDetails.getString("state").equals("error") ||
-               statusDetails.getString("state").equals("failure"));
+            statusDetails.getString("state") != null &&
+            (statusDetails.getString("state").equals("error") ||
+                statusDetails.getString("state").equals("failure"));
+    }
+
+    private void waitingMethod(Long t) {
```
mockitoguy: Why not pushing this to retryManager, making it testable? The more complexity we push out the better!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/RetryManagerTest.groovy
```diff
@@ -0,0 +1,53 @@
...
+
+    def "providing correct description"() {
+        given:
+        RetryManager retryCounter = RetryManager.defaultRetryValues()
+        for (int i = 0; i < 7; i++) {
+            retryCounter.waitNow({ })
+        }
+
+        expect:
+        retryCounter.describe() == "Waiting time so far: 210 seconds. Waiting 70 seconds..."
```
mockitoguy: Does it mean that this test executes in minutes? Let's make sure the unit tests are supersonic fast so that we get the best dev productivity. I suggest to create manager with super short timeouts for the purposes of unit testing.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
micd: I wanted to extract waiting method to be passed from external as a lambda expression for each usage of RetryManager as there might be different waiting approaches, but only one way of counting it which is done by RetryManager logic.

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/RetryManagerTest.groovy
micd: No it does not. 

Waiting method is delivered by lambda expression. For tests it's an empty lambda so every wait method just does nothing and is fast.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/util/GitHubStatusCheck.java
mockitoguy: I think so. This way we push out more complexity, simplifying code, and potentially making things easier to test. It is an internal API so we can always change it. Good question!

subprojects/shipkit/src/test/groovy/org/shipkit/internal/util/RetryManagerTest.groovy
mockitoguy: Great!

mockitoguy:
Thanks!

PR #713, opened 2018-04-24
Removed TODO we are not going to fix

mockitoguy:
We are fully reading the process output so there is no good way to set timeout other than spinning extra threads. That's too much complexity.


epeee:
Thank you, merging...

PR #733, opened 2018-07-02
Improved log message

mockitoguy:
I found it useful when onboarding "mockito-scala" to Shipkit


epeee:
Yes, indeed. 
Thank you!

Issue #740, opened 2018-07-25
Question: Can shipkit be used in parts?


jjstreet:
I have a company project that is not released as open source. It's also a project where I'm not trying to keep track of contributors or release notes (at least, not yet). We also do not use bintray and store artifacts in a enterprise owned nexus repository.

I would like to find a way to use shipkit in an incremental fashion. That is, I want to begin by letting shipkit handle version bumps, git tagging for releases, and potentially the publish of the jars (or just have the `maven-publish` plugin handle that particular part instead). Is it possible to apply just the plugins I need to perform those tasks? I've poured through the limited FAQs and HOW TO pages but was unable to find that information.

mockitoguy:
It should be possible. I'll look into this use case this evening.

mockitoguy:
I put together an example project: https://github.com/mockito/shipkit-examples/tree/master/versioning-and-tagging

Is this what you're after? Thanks for using Shipkit!

jjstreet:
Awesome I believe it is. I hope to check it out tomorrow in more detail with a colleague.

jjstreet:
I've been playing around with a sample project derived from the cookbook and its working great mostly. There is an issue with being able to release specific versions (manually changing the version.properties file as described by the docs). It seems `bumpVersionFile` is called even if not explicitly entered as a task, so I am not able to release without bumping. Any ideas?

mockitoguy:
> It seems bumpVersionFile is called even if not explicitly entered as a task, so I am not able to release without bumping. Any ideas?

Can you describe the use case a bit more? Regardless if we are pushing a specific version or not, we always want the 'bumpVersionFile' task to run. Examples:

- I'm changing "version.properties" to 2.0.0, run "./gradlew performGitPush", 2.0.0 is released (tag pushed) and 2.0.1 is stored in the "version.properties" (via bumpVersionFile). The file always contains "next" version that I will release.
- I'm not changing "version.properties", just running "./gradlew performGitPush". 2.0.1 is released (tag pushed) and 2.0.2 version is written to version.properties (via bumpVersionFile).

jjstreet:
I am closing this issue as I ended up getting my use case to function for the most part.

I wasn't able to use the gitPush task, but believe it is due to how our project is set up and not about shipkit.

I also looked at how bumpVersionFile works and understand why it is run the way it is.

PR #743, opened 2018-07-26
Exposed Git plugin

mockitoguy:
Needed for #740


mockitoguy:
I'll merge it without review if that's ok. A user is waiting for it and the code does not change any functionality, just exposes an existing plugin (plus a change in a javadoc).

PR #749, opened 2018-08-11
Ms4

mstachniuk:
Please review. I got a feedback from the community that GitHub token section is not clear.
You can check the result here: https://github.com/mockito/shipkit/blob/ms4/docs/getting-started.md#production-configuration


mockitoguy:
Absolutely! Thanks a lot!

Issue #452, opened 2017-10-01
Wrong badge type in release notes generated by shipkit gradle plugin


StefMa:
While working on #446 I found out that the release notes have the wrong badge "name".
shipkit is published to the gradle portal. But it shows "bintray".

See https://github.com/mockito/shipkit/blob/master/docs/release-notes.md

<img width="868" alt="bildschirmfoto 2017-10-01 um 10 35 06" src="https://user-images.githubusercontent.com/10229883/31053034-3d433be8-a694-11e7-8388-8a4d8f7b2177.png">

We should either found out where the plugin will be published and add the correct badge.
Or remove the badge completely and replace it with "repolink" or something ...

szczepiq:
Very good point. It looks like the publication repository is configurable but not quite the badge. Do you want to take a stab at this problem? We offer guidance where to find it in the codebase!

StefMa:
Yeah. Why not 🙃 
Would be great if you can say me some class names where I can find that snipped.

Additional to that.
How do we want to behave with that badge? 🤔 
I mean the repo can be "anything". 
It can be published to bintray, sonartype, gradle plugin portal, own maven repo...
I think we should agree on some for a badge (or get rid of the badge? 🤔 )
Should we also try to parse the URL to find out the repo?

I would suggest (because of "easieness"):
* We have a default string like `published repository ($VERSION)`. While "repository" is a link to the publicationRepository config (or automatic via bintray plugin)
* We try to parse the URL for bintray and the gradle plugin portal:
** If bintray add the shields.io badge. If gradle plugin portal use `published at gradle plugin portal ($VERSION)`(while gradle plugin portal is the link)
* We provide a `shipkitconfig.releaseNotes.publicationRepositoryBadge` (but the name is wrong 🤔 ) which can be a customize string... Like `my awesome repo`.. leads to `my awesome repo ($VERSION)`

Wow... Sounds not so easy 😄

epeee:
@StefMa currently, the badge is generated in
https://github.com/mockito/shipkit/blob/514ac7a43af0e3513a83f7518221086b203d4b45/subprojects/shipkit/src/main/groovy/org/shipkit/internal/notes/format/DetailedFormatter.java#L78-L84

wwilk:
Thanks for submitting this issue @StefMa !

[...]

PR #750, opened 2018-08-27
#452 Wrong badge type in release notes generated by shipkit gradle plugin

mstachniuk:


mstachniuk:
Good feedback. I will work on that...

Issue #755, opened 2018-09-18
License is always MIT in the generated POM


jpkrohling:
I just realized that no matter what I add as `bintray/pkg/licenses`, I always end up with MIT as license in the generated POM:

https://github.com/opentracing-contrib/java-interceptors/blob/v0.0.3/gradle/shipkit.gradle#L22

https://dl.bintray.com/opentracing/maven/io/opentracing/contrib/opentracing-interceptors/0.0.3/opentracing-interceptors-0.0.3.pom

This seems to be the relevant code:
https://github.com/mockito/shipkit/blob/98b9ff160ff73bf681b72baa6cad6ad99c8822ed/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java#L79

epeee:
Thank you for reporting this one!
I will work on a fix later on today..

jpkrohling:
I'm currently working on it and will send a PR soon.

jpkrohling:
Unfortunately, it seems that it's more complicated than I initially thought, as the POM customizer belongs to the Shipkit plugin but the licenses information is stored in the Bintray plugin. I'll let the experts fix then :)

jpkrohling:
@epeee, would you be able to provide some insight on the best way to solve this? As this is probably a blocker for us to use it, I'd be willing to give it a shot at fixing it, in case you don't have the time to fix it.

epeee:
@jpkrohling I already thought about how to fix it and it looks like I will be able to spend some time on this one tomorrow. Sounds good?

jpkrohling:
Sounds good to me :)

jpkrohling:
Looks like it does work, thanks!

PR #760, opened 2018-09-26
Make license used for pom configurable

epeee:
Very first draft (#755).

This pr extends shipkit configuration in a way that we have to configure the license used (incl url now).

```
shipkit {
    ...

    licenseInfo.license = "Eclipse Public License v2.0"
    licenseInfo.url = "http://www.eclipse.org/legal/epl-v20.html"
}
```

This will end up in a pom like 

```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
  ...
  <licenses>
    <license>
      <name>Eclipse Public License v2.0</name>
      <url>http://www.eclipse.org/legal/epl-v20.html</url>
      <distribution>repo</distribution>
    </license>
  </licenses>
  ...
```

The downside of this approach is that we still have to configure the license used by gradle bintray plugin, e.g.:

```
        bintray {
            pkg {
                ...
                licenses = ['EPL-2.0']
        ...
```

We could add this one also to shipkit config and do the configuration in our gradle plugin. As a result we'd have the license config in one place.

Note: this is a breaking change and will cause builds without a shipkit.licenseInfo configuration to fail:

````
Execution failed for task ':api:generatePomFileForJavaLibraryPublication'.
> Could not apply withXml() to generated POM
   > Please configure 'shipkit.licenseInfo.license' value (String).

```

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -26,6 +26,7 @@ public class ShipkitConfiguration {
 
     private final GitHub gitHub = new GitHub();
     private final ReleaseNotes releaseNotes = new ReleaseNotes();
+    private final LicenseInfo licenseInfo = new LicenseInfo();
     private final Git git = new Git();
     private final Team team = new Team();
 
@@ -87,6 +88,10 @@ public class ShipkitConfiguration {
 
     public ReleaseNotes getReleaseNotes() {
         return releaseNotes;
+    }
+
+    public LicenseInfo getLicenseInfo() {
+        return licenseInfo;
     }
 
     public Git getGit() {
@@ -219,6 +224,24 @@ public class ShipkitConfiguration {
 
         public void setWriteAuthToken(String writeAuthToken) {
             store.put("gitHub.writeAuthToken", writeAuthToken);
+        }
+    }
+
+    public class LicenseInfo {
+        public String getLicense() {
+            return store.getString("licenseInfo.license");
+        }
+
+        public void setLicense(String license) {
+            store.put("licenseInfo.license", license);
+        }
+
+        public String getUrl() {
+            return store.getString("licenseInfo.url");
+        }
+
+        public void setUrl(String url) {
+            store.put("licenseInfo.url", url);
         }
     }
 
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java
@@ -76,8 +76,8 @@ public class PomCustomizer {
         }
 
         Node license = root.appendNode("licenses").appendNode("license");
-        license.appendNode("name", "The MIT License");
-        license.appendNode("url", repoLink + "/blob/master/LICENSE");
+        license.appendNode("name", conf.getLicenseInfo().getLicense());
+        license.appendNode("url", conf.getLicenseInfo().getUrl());
         license.appendNode("distribution", "repo");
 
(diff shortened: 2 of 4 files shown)


mstachniuk:
@epeee Maybe you can read somehow Bintray plugin property to avoid duplication? 
@mockitoguy any hint?

jpkrohling:
Any updates on this one?

mockitoguy:
Looking at it. I met @epeee at CodeOne conference and he mentioned that his PR needs bump. @jpkrohling, sorry for slow turnaround! I'm on it.

mockitoguy:
Hey @epeee, I will review and provide my feedback shortly. In the meantime, I commented on the ticket so that @jpkrohling is unblocked and can customize the pom.

mstachniuk:
I found a hack how to get license info from bintray configuration. @epeee @mockitoguy WDYT?

```
        Object bintray = project.getExtensions().findByName("bintray");

        try {
            Method getPkg = bintray.getClass().getDeclaredMethod("getPkg");
            getPkg.setAccessible(true);
            Object pkg = getPkg.invoke(bintray);
            Field licensesField = pkg.getClass().getDeclaredField("licenses");
            licensesField.setAccessible(true);
            String[] licenses = (String[]) licensesField.get(pkg);
            for (String license : licenses) {
                LOG.lifecycle(" lic " + license);
            }
        } catch (Exception e) {
            e.printStackTrace();
        }
```

mockitoguy:
> I found a hack how to get license info from bintray configuration

Please no :) That's a lot of complexity for little benefit.

mstachniuk:
It's a simple reflection ;-) But as you wish!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java
```diff
@@ -76,8 +76,8 @@ static void customizePom(Node root, ShipkitConfiguration conf,
         }
 
         Node license = root.appendNode("licenses").appendNode("license");
-        license.appendNode("name", "The MIT License");
-        license.appendNode("url", repoLink + "/blob/master/LICENSE");
+        license.appendNode("name", conf.getLicenseInfo().getLicense());
```
mockitoguy: Can we add defaults, Apache 2.0 and blob/master/LICENSE?

mockitoguy:
Once we merge it, can we get the license value populated to the Bintray extension, too? Thanks!!! Nice change!

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java
epeee: Sure, good feedback, thx!

mockitoguy: Cool. Let me know when you're ready.

MIT is 30y old and lawyers no longer like it. Let's default to Apache 2.0.

mockitoguy:
Before merging, we need to add defaults (as my comment) or make it "lenient". I suggest to add defaults because it is simpler and more useful. Then we can merge :) Let me know when you're ready!

subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -95,6 +96,10 @@ public ReleaseNotes getReleaseNotes() {
         return releaseNotes;
     }
 
+    public LicenseInfo getLicenseInfo() {
```
mstachniuk: How to now configure a few licenses? E.g. one open source and one commercial?

Issue #75, opened 2017-04-17
Automatically update releases on GitHub


TimvdLippe:
GitHub has nice releases (https://github.com/mockito/mockito/releases), but we are not using them right now. It would be great if the changelog can be published there. The corresponding endpoint is https://developer.github.com/v3/repos/releases/#create-a-release

It would also be nice if "notable versions" would be released as a full release, while all other versions are released as pre-release (see the boolean parameter to the endpoint).

szpak:
I've been using [Chandler](https://github.com/mattbrictson/chandler) for a CHANGELOG.md file synchronization and it works fine (maybe except for https://github.com/mattbrictson/chandler/issues/33). Alternatively the mentioned API can be used directly to do not use a Ruby application.

szczepiq:
Great idea! Thank you for submitting!!!

thekingnothing:
I'm voting for the feature.

jpkrohling:
We could make use of this for the Jaeger project as well. This is probably the only manual step remaining for our release process.

TimvdLippe:
This was implemented in #766

Issue #233, opened 2017-06-13
Start shipping Javadoc


mockitoguy:
### Problem

Currently, we rely on external https://javadoc.io service to host viewable Javadoc. When we publish the jars with javadocs to Maven Central, the javadoc is automatically downloaded, unzipped and hosted at javadoc.io. This works well for projects that ship to Central. Projects that don't ship to Central (some project ship only to https://bintray.com) don't get this benefit. Best example is Shipkit project itself - currently we don't host unzipped Javadocs that are easy to view online.

### Design spec

See PR #697

szczepiq:
@mstachniuk, do you want to scope this ticket a bit more? I can help if needed :)

mstachniuk:
I will do after a long weekend.

mockitoguy:
@mstachniuk, how's life? Will you be able to take a stab at this feature? Cheers!

mockitoguy:
I worked on designing the plugin today and added a pseudo code design to the description. Hope it helps!

mstachniuk:
@mockitoguy Sure, it will help me. Thank you!

mockitoguy:
@mstachniuk, the design spec is ready #697 :)

mstachniuk:
It's in progress...

PR #766, opened 2018-12-28
#75 Automatically update releases on GitHub

mstachniuk:


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/ReleaseNotesPlugin.java
```diff
@@ -51,46 +53,76 @@ public void apply(final Project project) {
...
+            task.setPreviewMode(previewMode);
+
+            LazyConfiguration.lazyConfiguration(task, () -> {
+                task.setGitHubWriteToken(conf.getGitHub().getWriteAuthToken());
+            });
+        });
+
+        updateReleaseNotesOnGitHubCleanupTask.setGitHubApiUrl(conf.getGitHub().getApiUrl());
+        updateReleaseNotesOnGitHubCleanupTask.setUpstreamRepositoryName(conf.getGitHub().getRepository());
+        updateReleaseNotesOnGitHubCleanupTask.setPublicationRepository("aa");
```
epeee: hmm, this one looks suspicious.

epeee:
What about making new features like this one optional in the first place?
Additionally, what do you think about extracting the "release notes update to github" to a separate plugin?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/ReleaseNotesPlugin.java
mstachniuk: Good catch! This line will be deleted by me in the next PR. The proper value is set in BintrayReleasePlugin.java:65

mstachniuk:
In my opinion, there is no point to make it optional - a user doesn't need to configure anything, it's only added value. Also incubating fetures usually stay incubating very long (sometimes forever) what is not good for those features.

I thougt about separate plugin, it will be very simmilar to `ReleaseNotesPlugin` and needed to duplicate a lot of code. For compleatly new things, like https://github.com/mockito/shipkit/issues/233 I will create separate plugin. 

Do you think it should be separated? We can discuss on our Slack.
Thanks a lot for the review, better later then never :)

PR #771, opened 2019-02-13
Update dependencies

epeee:


epeee:
also merging since it is just about updating some non prod dependencies.

Issue #775, opened 2019-02-20
Fix updateReleaseNotesOnGitHub task


mstachniuk:
After release Shipkit 2.1.0 and use it in https://github.com/mockito/shipkit-example the task updateReleaseNotesOnGitHub doesn't work:

```
[performRelease] :updateReleaseNotesOnGitHub
[performRelease]   Building new release notes based on /home/travis/build/mockito/shipkit-example/docs/release-notes.md
[performRelease] GET /repos/mockito/shipkit-example/releases/tags/v0.16.18
[performRelease] :updateReleaseNotesOnGitHub FAILED
```

GET https://api.github.com/repos/mockito/shipkit-example/releases/tags/v0.16.18 returns

```
{
"message": "Not Found",
"documentation_url": "https://developer.github.com/v3/repos/releases/#get-a-release-by-tag-name"
}
```

It means that releases are not automatically created when tag is created. Release needs to be created by API.

mstachniuk:
Fixed in 2.1.3

PR #776, opened 2019-02-20
Fix updateReleaseNotesOnGitHub task #775

mstachniuk:
Tag in GitHub doesn't automatically create GitHub Release
GitHub Release needs to be created by API

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotesOnGitHub.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotesOnGitHub.java
@@ -29,13 +29,8 @@ public class UpdateReleaseNotesOnGitHub {
             LOG.lifecycle("  Preview of release notes update:\n" +
                 "  ----------------\n" + releaseNotesText + "\n----------------");
         } else {
-            updateOnGitHub(task, releaseNotesText);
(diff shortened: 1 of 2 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/notes/tasks/UpdateReleaseNotesOnGitHub.java
```diff
@@ -58,24 +53,37 @@ private String findReleaseByTagName(UpdateReleaseNotesOnGitHubTask task) throws
...
-        JsonObject responseJson = (JsonObject) Jsoner.deserialize(response);
+        JsonObject body = new JsonObject();
+        body.put("tag_name", tagName);
+        body.put("name", tagName);
+        body.put("body", text);
 
-        String htmlUrl = responseJson.getString("html_url");
-        LOG.lifecycle("  Successfully updated release notes on GitHub: {}", htmlUrl);
+        LOG.debug("GitHub create release by tag name POST {}", url);
+        LOG.lifecycle("POST {}", url);
```
epeee: a minor one: we can get rid of one of this log lines since they basically provide the same information.

mstachniuk: Thanks a lot for the review, I removed duplicated logs in this class (and checked other classes for 2.1.0). Next PRs coming soon...

PR #780, opened 2019-02-24
Update objenesis (v3.0.1) and get rid of not needed dependency

epeee:


epeee:
... merging since it is just about test dependencies.

PR #784, opened 2019-03-02
fix gradle badge generation

epeee:
fixes #774 

now generates badges like Gradle

I've already adjusted the badges of the last two releases in docs/release-notes.md

PR #785, opened 2019-03-02
deprecate publicationPluginName (ShipkitConfiguration)

epeee:
we can get rid of it's usage once #784 is merged.


epeee:
merging since we agreed on deprecating this one.

mstachniuk:
It would be nice to add a comment, why and from with version a property is depricated.
In my opinion `ShipkitConfiguration` is public API so minor version should be bump, see my comment: https://github.com/mockito/shipkit/pull/784#discussion_r261886839
and https://semver.org/#spec-item-7

PR #792, opened 2019-03-07
Describe why we are deprecating publicationPluginName

epeee:
see #785

### DIFF
--- a/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
+++ b/subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
@@ -303,6 +303,8 @@ public class ShipkitConfiguration {
          * Get the Publication Plugin Name
          *
          * @see @setPublicationPluginName(String)
+         *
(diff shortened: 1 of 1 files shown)


subprojects/shipkit/src/main/groovy/org/shipkit/gradle/configuration/ShipkitConfiguration.java
```diff
@@ -325,6 +327,8 @@ public String getPublicationPluginName() {
          *     https://img.shields.io/maven-metadata/v/https/plugins.gradle.org/m2/org/shipkit/java/org.shipkit.java.gradle.plugin/maven-metadata.xml.svg?colorB=007ec6&label=Gradle
          * </pre>
          * This will show nice badge with actual plugin version in Gradle Plugin Portal.
+         *
+         * @deprecated since 2.1.6 because we no longer are using this one. It is scheduled to be removed in 2.2.0.
```
mstachniuk: It's a backward incompatible change, so it should be removed in next major release. https://semver.org/#spec-item-8
Please change the description and add TODO / GitHub issue to not forget about that.

Thanks for fixing that.

PR #804, opened 2019-03-22
Re-enable temporary disabled features

mstachniuk:


mockitoguy:
CIs are failing, do you want to check before re-enabling?

PR #812, opened 2019-03-31
Add jacoco

mstachniuk:


epeee:
What do you think about setting jacoco toolversion?
we are using gradle 4.0.2 which by default uses jacoco '0.7.8'.
let's use a more recent version.

mockitoguy:
Nice!

PR #815, opened 2019-04-03
Set Jacoco path to fix upload report problem

mstachniuk:


mstachniuk:
Now uploading of the report should works

PR #818, opened 2019-04-04
use sourceSets.main for JacocoReport

epeee:
it looks like we do get coverage for the test source set now, let's change it to the main source set.


epeee:
see coverage for [this](https://codecov.io/gh/mockito/shipkit/tree/ep2/subprojects/shipkit/src/main/groovy/org/shipkit/gradle) branch vs coverage on [master](https://codecov.io/gh/mockito/shipkit/tree/master/subprojects/shipkit/src/test/groovy/org/shipkit/gradle)

PR #821, opened 2019-04-05
Fixed logging issue

epeee:
Fixes problem: 'Number of placeholders does not match number of arguments in logging call'


mockitoguy:
Thank you!

PR #829, opened 2019-05-31
Add basic Android library support [obsolete]

koral--:
AndroidPublishPlugin unit test cannot pass due to Gradle 5.3. requirement.

`testDownstream` task fails for me even on master:
```
Starting a Gradle Daemon, 1 incompatible Daemon could not be reused, use --status for details
Parallel execution is an incubating feature.
  Building version '2.2.6' (value loaded from 'version.properties' file).
  [INCUBATING] upgrade-dependency plugin is incubating and may change in any version.
  [INCUBATING] downstream-testing plugin is incubating and may change in any version.
Build cache is an incubating feature.
Using local directory build cache for build ':shipkit-root' (location = /home/koral/.gradle/caches/build-cache-1).
:shipkit-root:shipkit:compileJava NO-SOURCE
:shipkit-root:shipkit:compileGroovy FROM-CACHE
:shipkit-root:shipkit:processResources
:shipkit-root:shipkit:classes
:shipkit-root:shipkit:jar
  Building version '0.16.23' (value loaded from 'version.properties' file).
  [INCUBATING] upgrade-dependency plugin is incubating and may change in any version.

FAILURE: Build failed with an exception.

* What went wrong:
Please export 'GH_WRITE_TOKEN' variable first!
It is highly recommended to keep write token secure and store env variable 'GH_WRITE_TOKEN' with your CI configuration. Alternatively, you can configure GitHub write auth token explicitly (don't check this in to Git!):
  shipkit.gitHub.writeAuthToken = 'secret'
```


docs/getting-started.md
```diff
@@ -111,6 +111,34 @@ allprojects {
...
+
+Example:
+
+"gradle/shipkit.gradle" file:
+
+```Gradle
+shipkit {
+    //other properties
+    
+    android.artifactId = 'shipkit-android'
```
mstachniuk: Is it possible to publish a few Android libs for one project?

mstachniuk:
When the PR is ready and merged please provide an example project with the usage of new plugin: https://github.com/mockito/shipkit-android-example

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/android/AndroidPublishPlugin.java
```diff
@@ -0,0 +1,81 @@
...
+                });
+                LOG.info("{} - configured '{}' publication", project.getPath(), p.getArtifactId());
+            });
+
+            //so that we flesh out problems with maven publication during the build process
+            project.getTasks().getByName("build").dependsOn(MAVEN_LOCAL_TASK);
+        });
+    }
+
+    private static void ensureGradleVersion() {
```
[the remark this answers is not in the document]
koral--: aar-publish plugin under the hood uses [SoftwareComponentFactory](https://docs.gradle.org/current/javadoc/org/gradle/api/component/SoftwareComponentFactory.html) which was added in 5.3

docs/getting-started.md
koral--: With the current code it is impossible. I'll fix this.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/android/AndroidPublishPlugin.java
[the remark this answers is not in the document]
koral--: Is it acceptable to use it or should I rewrite that logic using APIs available in Gradle 4.0.2?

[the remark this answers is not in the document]
mstachniuk: Good question. @mockitoguy is always against other dependencies because it introduces a cost of support and there is no separation in Gradle between different plugins (all of them use the same classloader) so it can produce potential conflicts in future. If the logic is not so big it's better to rewrite/copy it to Shipkit.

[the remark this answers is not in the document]
koral--: OK. I'll try to get rid of that dependency.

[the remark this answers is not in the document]
koral--: Done, since AGP 3.6.0-beta05 it is no longer needed.

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/android/AndroidPublishPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+        Task snapshotTask = project.getTasks().getByName(LocalSnapshotPlugin.SNAPSHOT_TASK);
+        snapshotTask.dependsOn(MAVEN_LOCAL_TASK);
+
+        project.getPlugins().apply("maven-publish");
+
+        BintrayExtension bintray = project.getExtensions().getByType(BintrayExtension.class);
+        bintray.setPublications(PUBLICATION_NAME);
+
+        project.getPlugins().withId("com.android.library", plugin -> {
+            project.afterEvaluate(evaluatedProject -> {
```
mstachniuk: Please use DeferredConfiguration#deferredConfiguration instead

subprojects/shipkit/shipkit.gradle
```diff
@@ -22,8 +22,9 @@ dependencies {
 
     compile "com.github.cliftonlabs:json-simple:2.1.2"
     compile 'com.jfrog.bintray.gradle:gradle-bintray-plugin:1.8.4'
-    compile "com.gradle.publish:plugin-publish-plugin:0.9.10"
+    compile 'com.gradle.publish:plugin-publish-plugin:0.9.10'
```
mstachniuk: Is it still needed?

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/android/AndroidPublishPlugin.java
```diff
@@ -0,0 +1,82 @@
...
+
+                GradleDSLHelper.publications(project, publications -> {
+                    MavenPublication p = publications.create(PUBLICATION_NAME, MavenPublication.class, publication -> {
+                        publication.setArtifactId(androidPublishConfiguration.getArtifactId());
+
+                        SoftwareComponent releaseComponent = project.getComponents().findByName("release");
+                        if (releaseComponent == null) {
+                            throw new GradleException("'release' component not found in project. " +
+                                "Make sure you are using Android Gradle Plugin 3.6.0-beta05 or newer.");
```
mstachniuk: This message may be confusing. This PR include com.android.tools.build:gradle:3.6.0-beta05 so it should be applied by Shipkit autmatically. Or documentation should describe that the user needs to apply it manually.

koral--: User has to apply `com.android.library` plugin to every Android Library plugin.

However indeed, Gradle dependency on `com.android.tools.build:gradle` is not needed since we are not using any classes from it. We only refer to plugin and component by names given in plain strings. I'll remove the dependency.

I'll also mention min AGP version in documentation.

subprojects/shipkit/shipkit.gradle
koral--: Gradle plugin itself seems to be needed since shipkit is also published to gradle plugin portal. However, change in this line is not needed.

koral--:
Superseded by #853

subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/util/PomCustomizer.java
```diff
@@ -65,10 +66,13 @@ public void execute(XmlProvider xml) {
      */
     static void customizePom(Node root, ShipkitConfiguration conf,
                              String projectName, String projectDescription,
-                             ProjectContributorsSet contributorsFromGitHub) {
-        //Assumes project has java plugin applied. Pretty safe assumption
+                             ProjectContributorsSet contributorsFromGitHub,
+                             boolean isAndroidLibrary) {
         root.appendNode("name", projectName);
-        root.appendNode("packaging", "jar");
+        if (!isAndroidLibrary) {
```
mockitoguy: Oups, I changed this part of the code in my recent PR. Can you rebase? Thanks!

PR #861, opened 2020-02-09
Spec for discussing future of Shipkit

mockitoguy:
@mockito/shipkit-developers, any thoughts?

### DIFF
--- a/docs/design-specs/future-shipkit.md
+++ b/docs/design-specs/future-shipkit.md
@@ -1,0 +1,26 @@
+## Future Shipkit
+
+Discussion of future development of Shipkit.
+
+### What worked well
+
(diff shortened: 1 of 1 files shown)


epeee:
:+1:, thx!

mockitoguy:
I'll keep it open for 1-2 weeks to collect more feedback.

Thanks for review!

nallar:
> adoption is small (too opinionated? / too complicated?).
a little of that, ~~but also you're missing out on expensive consultants pushing people to move to it as a way to generate work~~

Not that it means much, but shipkit's saved a lot of time here and very happy that it exists.

mockitoguy:
@nallar, thank you for feedback!

Issue #842, opened 2019-08-20
Instructions/compatibility with GitHub Actions


magneticflux-:
Now that GitHub Actions is in beta and more people are getting access to it, it makes sense to broaden the compatible CI environments for ShipKit.

aepfli:
Just an fyi, we (https://github.com/junit-pioneer/junit-pioneer) migrated now from travis to github actions. it is working with some "hacks".

1. shipkit relies on the environment variable `TRAVIS_BRANCH` to work successfully, but this can be easily injected
2. `github.ref` which contains the branch contains it with `refs/head/` in the beginning, which cause troubles and needs to be split

our deploy action look like

``` yaml
deploy:
    needs: build
    runs-on: ubuntu-latest
    name: Deploy with Shipkit
    if: github.ref == 'refs/heads/master'
    steps:
      - name: Sanitize Branch
        uses: frabert/replace-string-action@master
        id: gitbranch
        with:
          pattern: 'refs/heads/(.*)'
          string: "${{ github.ref }}"
          replace-with: '$1'
      - uses: actions/checkout@v2
      - name: Setup java
        uses: actions/setup-java@v1
        with:
          java-version: 8
      - name: Perform Release
        uses: eskatos/gradle-command-action@v1
        env:
          TRAVIS_BRANCH: ${{ steps.gitbranch.outputs.replaced }}
          GH_WRITE_TOKEN: ${{ secrets.GH_WRITE_TOKEN }}
          BINTRAY_USER_NAME: ${{ secrets.BINTRAY_USER_NAME }}
          BINTRAY_API_KEY: ${{ secrets.BINTRAY_API_KEY }}
          NEXUS_TOKEN_USERNAME: ${{ secrets.NEXUS_TOKEN_USERNAME }}
          NEXUS_TOKEN_PASSWORD: ${{ secrets.NEXUS_TOKEN_PASSWORD }}
        with:
          arguments: ciPerformRelease
```
and in our shipkit.gradle we added some parameters so that CI is correctly linked.

mstachniuk:
I created an example project for that issue: https://github.com/mstachniuk/shipkit-gh-actions-example

It's a little bit complicated but works better in case of setting `TRAVIS_BRANCH` for Pull Request build. 
Using just `TRAVIS_BRANCH: ${{ steps.gitbranch.outputs.replaced }}` sets it to `refs/pull/7/merge` but Travis sets it to the real PR source branch.

[...]

PR #868, opened 2020-05-02
Additional CI Provider

aepfli:
This is a draft pullrequest, it is just a small idea, i wanted to challenge first, before investing more time into this. #842 

## Refactoring TravisPlugin to generic Ci plugin

Moving all generics part into new CiPlugin,
and implementing an interface called CiProvider,
which allows for easier generation of CI specific
data.

## Adding GitHubActionsProvider

Allows usage of GitHub Actions, and utilizes the provided
environment variables to generate necessary data.

## Adding Generic Ci Provider

Which uses default fields, so other systems can be easier integrated.


aepfli:
sidenote, this is untested, and far from final, i just wanted to show what we could theoretically do, and if this would be a suitable approach.

Issue #864, opened 2020-02-10
github access_token query parameter deprecated, breaks after 2020/07/01


nallar:
deprecation notice should be here but it's 404ing: https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters

> Please use the Authorization HTTP header instead, as using the `access_token` query parameter is deprecated and will be removed July 1st, 2020.

mockitoguy:
Thanks for reporting!

magneticflux-:
Here's a newer post on the issue: https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/

mockitoguy:
@shestee, can you take a look at this issue?

shestee:
@mockitoguy Sure, I'll check this out.

magneticflux-:
@nallar Closed by #871 now?

nallar:
Thanks! ^_^

PR #871, opened 2020-06-26
Fixed GitHub authorization

shestee:
GitHub will no longer support authentication through query parameters (reported in issue #864; more info: https://developer.github.com/changes/2019-11-05-deprecated-passwords-and-authorizations-api/#authenticating-using-query-parameters, https://developer.github.com/changes/2020-02-10-deprecating-auth-through-query-param/).
To deal with this issue there was a need to move authentication to the connection header.
It was done in GitHubListFetcher and GitHubObjectFetcher classes. Followed by this refactor there was also a necessity to refactor GitHubContributorsFetcher, RecentContributorsFetcher and
GitHubTicketFetcher classes.


mockitoguy:
I'm going to merge it because I cannot reproduce the failure locally so I think it's a red herring. The change is useful and avoid deprecation of GitHub API.

Issue #876, opened 2020-07-20
GradlePortalReleasePlugin: Cannot add task 'validatePlugins' as a task with that name already exists


vlsi:
Shipkit produces an error when modern Gradle is used since Gradle's `validatePlugins` task conflicts with shipkit's one.

https://github.com/mockito/shipkit/blob/d0dcd77ace3f0f59a8a2a9b30eeb7def8a87acfa/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/release/GradlePortalReleasePlugin.java#L54

https://github.com/mockito/shipkit/blob/98b9ff160ff73bf681b72baa6cad6ad99c8822ed/subprojects/shipkit/src/main/groovy/org/shipkit/internal/gradle/plugin/PluginValidationPlugin.java#L27

Does that mean the task should be renamed?

Test case:

1. https://github.com/myniva/gradle-s3-build-cache
2. Update Gradle to 6.5.1
3. Run `gradlew help`

```
org.gradle.api.GradleScriptException: A problem occurred evaluating root project 's3-build-cache'.
	at org.gradle.groovy.scripts.internal.DefaultScriptRunnerFactory$ScriptRunnerImpl.run(DefaultScriptRunnerFactory.java:93)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.lambda$apply$0(DefaultScriptPluginFactory.java:133)
	at org.gradle.configuration.ProjectScriptTarget.addConfiguration(ProjectScriptTarget.java:77)
	at org.gradle.configuration.DefaultScriptPluginFactory$ScriptPluginImpl.apply(DefaultScriptPluginFactory.java:136)
	at org.gradle.configuration.BuildOperationScriptPlugin$1$1.run(BuildOperationScriptPlugin.java:69)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:395)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:387)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor$1.execute(DefaultBuildOperationExecutor.java:157)
	at org.gradle.internal.operations.DefaultBuildOperationExecutor.execute(DefaultBuildOperationExecutor.java:242)
[...]

PR #877, opened 2020-07-24
Gradle 6.x compatibility: cannot add task 'validatePlugins' as a task with that name already exists

vlsi:
This fixes https://github.com/mockito/shipkit/issues/876 (GradlePortalReleasePlugin: Cannot add task 'validatePlugins' as a task with that name already exists)


vlsi:
ShipkitJavaIntegTest fails with Gradle 6.5, and the cause seems to be tasks returned in the different order.

I don't know what does the test verify, so I don't know how to heal it.

@mockitoguy , can you please clarify?

Expected:

```
...
:api:generatePomFileForJavaLibraryPublication
:api:compileJava
:api:processResources
:api:classes
:api:jar
:api:javadoc
:api:javadocJar
...
```

Actual:

```
...
:api:compileJava
:api:processResources
:api:classes
:api:jar
:api:generateMetadataFileForJavaLibraryPublication
...
```

vlsi:
One of the workarounds could be "ignore :generatePom and :generateMetadata" tasks in `skippedTaskPathsGradleBugWorkaround`. The tasks do not seem to be important much, and they seem to be the only ones that produce unordered results.

WDYT?

mockitoguy:
@vlsi, we are going to put the project in maintenance mode soon. The new plugins are:

 - https://github.com/shipkit/shipkit-changelog
 - https://github.com/shipkit/shipkit-auto-version

Check out the new plugins. Would they work for you?

PR #881, opened 2020-08-23
add Gradle Wrapper validation

sullis:
https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action
﻿

### DIFF
--- a/.github/workflows/gradle-wrapper-validation.yml
+++ b/.github/workflows/gradle-wrapper-validation.yml
@@ -1,0 +1,10 @@
+name: "Validate Gradle Wrapper"
+on: [push, pull_request]
+
+jobs:
+  validation:
+    name: "Gradle wrapper validation"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: gradle/wrapper-validation-action@v1


sullis:
Screen Shot 2020-08-23 at 12 34 58 PM

koral-- (changes requested):
We have sha cheksum declared in Gradle wrapper properties: https://github.com/mockito/shipkit/blob/master/gradle/wrapper/gradle-wrapper.properties#L7

So every invocation when wrapper is not yet downloaded will cause validation to be performed automatically. No matter if running on CI or not.
As you can see on Travis CI log it is also downloaded there.

Apart from that we're using Travis, not Github actions, what is the purpose of performing additional validation?

szpak:
> Apart from that we're using Travis, not Github actions, what is the purpose of performing additional validation?

@koral-- That validation can be useful to confirm that the wrapper upgrade proposed in a PR is not the rogue one. For example, that check could [detect](https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action) malicious `gradle-wrapper.jar` which always downloads Gradle from arbitrary location, ignoring defined SHA hash.

However, as you noted, using Travis instead of GitHub Actions is a very good point which complicates that PR in ShipKit ;).

sullis:
> > Apart from that we're using Travis, not Github actions, what is the purpose of performing additional validation?
> 
> @koral-- That validation can be useful to confirm that the wrapper upgrade proposed in a PR is not the rogue one. For example, that check could [detect](https://blog.gradle.org/gradle-wrapper-checksum-verification-github-action) malicious `gradle-wrapper.jar` which always downloads Gradle from arbitrary location, ignoring defined SHA hash.
> 

Gradle's GitHub Action validates the Gradle wrapper jar against a known list of valid checksums.

The `gradle-wrapper.properties` file in the ShipKit repo could contain a rogue checksum.

koral--:
OK now I get it.
So it seems that we can have this GH action validation in parallel with Travis. GitHub allows multiple PR checks.
What do you think guys?

mstachniuk:
For me it's ok to validate gradle-wrapper using GH Actions.

sullis:
ping @mstachniuk @koral--

koral-- (approved):

